PR #7, opened 2018-07-05
main: wait for console and rootfs, try to exec init or shell

kaman:


src/main.c
```diff
@@ -115,11 +115,22 @@ void main_initthr(void *unused)
 	proc_fileSet(2, 3, &oid, 0, 0);
 
 	if (last != NULL)
```
agkaminski: I suggest using brackets for if (last != NULL)

kaman: done

PR #60, opened 2020-01-13
New version of the build-toolchain.sh

kemonats:
The libraries gmp, mpfr, mpc are built together with GCC [link](https://gcc.gnu.org/install/prerequisites.html) :

Alternatives [link](https://gcc.gnu.org/wiki/InstallingGCC) :
1. installation and use of system libraries.
2. after extracting the GCC source archive, simply run the ./contrib/download_prerequisites script in the GCC source directory.

Both do not provide version control.


nalajcie:
Tested for target arm-phoenix. The "--disable-shared" functionality stays intact. LGTM.

PR #63, opened 2020-01-15
STM32L4 port

agkaminski:
Added new target, STM32L4x6

### DIFF
--- a/src/hal/armv7/stm32l4.c
+++ b/src/hal/armv7/stm32l4.c
@@ -21,21 +21,18 @@
 
 struct {
 	volatile u32 *rcc;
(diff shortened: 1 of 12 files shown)


agkaminski:
Please don't merge without squashing

src/hal/armv7/stm32l4.c
```diff
@@ -0,0 +1,923 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * STM32 basic peripherals control functions
```
anglov: STM32L4 maybe?
L1 in stm32l1.c respectively

PR #69, opened 2020-01-22
stm32: Do not clean console and change newline mode on start

anglov:
"\033c" – reset console 
"\033[2J" – clear screen
"\033[20h" – change \r to \r\n
"\033[f" – Set cursor to 0,0

I remove console resetting and changing newline mode which should not be needed.

Now it's aligned to imx6ull console.
@Maxez Please confirm if I wrote correctly :)

### DIFF
--- a/src/hal/armv7/console-stm32l1.c
+++ b/src/hal/armv7/console-stm32l1.c
@@ -114,9 +114,8 @@ void _hal_consoleInit(void)
 	*(console_common.base + brr) = console_common.cpufreq / 9600; /* 9600 baud rate */
 	*(console_common.base + cr1) |= 0x2000;
 
-	_hal_consolePrint("\033c");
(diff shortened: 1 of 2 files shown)


anglov:
Are you sure?
In picocom terminal on linux it behaves following way:
Return key write \r\n  instead of only \r
Should depend on tty settings (which should be set by a serial console terminal), not the OS.
Also – as I pointed before – now it's aligned to imx6ull console which should works correctly even on 🍎 systems.
Please take a look at
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/2e3e35d48b55fd46926df8d1a33344e13ec21745/src/hal/arm/console-imx6ull.c#L84

agkaminski:
@js6i can you check on your :apple: ?

PR #86, opened 2020-05-01
ia32/string.c: other use of registers

kemonats:
`hal_memset`:
value `n` is already in `edx` `"+d" (n)`

`hal_memsetw`:
`ecx` instead `edx`


src/hal/ia32/string.h
```diff
@@ -48,7 +48,6 @@ static inline void hal_memset(void *where, u8 v, unsigned int n)
 	(" \
 		cld; \
 		movl %0, %%ecx; \
-		movl %%ecx, %%edx; \
```
agkaminski, line 51: Isn't it an accident that _n_ is already in edx? I don't see edx in clobber, so it might be.

PR #88, opened 2020-05-02
arm/string.h: code optimization

kemonats:
My suggestion for changes to the function code written in assembler

Major changes:
1. other use of registers,
2. use of other instructions

`mem*` - I changed the code size without changing the loop algorithm.
`str*` - practically same as generated by a compiler with the `-Os` option for functions in `c`.

Result:
* Difference in `*.img` file size

`armv7a7-imx6ull` - about `1.6kb (-Og)`
`armv7m7-stm32l152xd` - about `400b (-Os)`

`memcpy and memset` - for better performance, we can use a different algorithm in loops

### DIFF
--- a/src/hal/arm/string.h
+++ b/src/hal/arm/string.h
@@ -23,29 +23,26 @@ static inline void hal_memcpy(void *dst, const void *src, unsigned int l)
 {
 	__asm__ volatile
 	(" \
-		mov r1, %2; \
-		mov r3, %1; \
-		mov r4, %0; \
-		orr r2, r3, r4; \
-		ands r2, #3; \
+		orr r3, %0, %1; \
+		lsls r3, r3, #30; \
 		bne 2f; \
 	1: \
-		cmp r1, #4; \
+		cmp %2, #4; \
 		ittt hs; \
-		ldrhs r2, [r3], #4; \
-		strhs r2, [r4], #4; \
-		subshs r1, #4; \
+		ldrhs r3, [%1], #4; \
+		strhs r3, [%0], #4; \
+		subshs %2, #4; \
 		bhs 1b; \
 	2: \
-		cmp r1, #0; \
+		cmp %2, #0; \
 		ittt ne; \
-		ldrbne r2, [r3], #1; \
-		strbne r2, [r4], #1; \
-		subsne r1, #1; \
+		ldrbne r3, [%1], #1; \
+		strbne r3, [%0], #1; \
+		subsne %2, #1; \
 		bne 2b"
+	: "+r" (dst), "+r" (src), "+r" (l)
 	:
-	: "r" (dst), "r" (src), "r" (l)
-	: "r1", "r2", "r3", "r4", "memory", "cc");
+	: "r3", "memory", "cc");
 }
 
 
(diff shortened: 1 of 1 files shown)


src/hal/arm/string.h
```diff
@@ -55,26 +52,22 @@ static inline int hal_memcmp(const void *ptr1, const void *ptr2, size_t num)
...
-		beq 3f; \
-		sub r3, #1; \
-		ldrb r4, [r1], #1; \
-		ldrb r5, [r2], #1; \
-		cmp r4, r5; \
-		beq 1b; \
-		blo 2f; \
-		mov %0, #1; \
-		b 3f; \
+		cbz  %3,3f; \
```
agkaminski, line 56: Missing space after ,

src/hal/arm/string.h
```diff
@@ -55,26 +52,22 @@ static inline int hal_memcmp(const void *ptr1, const void *ptr2, size_t num)
...
+		subs %3, #1; \
+		ldrb r3, [r1], #1; \
+		ldrb r4, [r2], #1; \
+		cmp  r3, r4; \
```
agkaminski, line 60: One space between opcode and arg is enough

src/hal/arm/string.h
```diff
@@ -117,10 +107,10 @@ static inline unsigned int hal_strlen(const char *s)
 	(" \
 	1: \
 		ldrb r1, [%1, %0]; \
-		add %0, #1; \
-		cmp r1, #0; \
-		bne 1b; \
-		sub %0, #1;"
+		cbz r1, 2f; \
+		adds %0, #1; \
```
agkaminski, line 111: This changes code behaviour. New code outputs one less.

kemonats: The value returned by the function is valid (the code is almost identical to the one generated by the compiler).
disassembly strlen in c:
```
00000000 <strlen>:
   0:   4603            mov     r3, r0
   2:   2000            movs    r0, #0
   4:   5c1a            ldrb    r2, [r3, r0]
   6:   b902            cbnz    r2, a <strlen+0xa>
   8:   4770            bx      lr
   a:   3001            adds    r0, #1
   c:   e7fa            b.n     4 <strlen+0x4>
```

agkaminski: I tested it, result is the same indeed. I missed sub on the previous version.
I think that there is no need for _adds_ instruction, _add_ if enough. We don't need add result's flags, are we?

agkaminski:
I tested it. Sometimes unexpected register allocation causes problems:
/tmp/cctn74uL.s: Assembler messages:
/tmp/cctn74uL.s:2953: Error: lo register required -- `cbz ip,4f'
(in hal_strncmp, it seems).

kemonats:
> I tested it. Sometimes unexpected register allocation causes problems:
> /tmp/cctn74uL.s: Assembler messages:
> /tmp/cctn74uL.s:2953: Error: lo register required -- `cbz ip,4f'
> (in hal_strncmp, it seems).

I tested it in version 7.1 and 9.3 and didn't have any error.
I suggested the -mtumb flag, and therefore cbz, movs, adds ands.
I will change this to work in 32-bit version.
Maybe it would be better to use the functions written in c for hal_str *.

agkaminski:
movs vs mov, adds vs add have nothing to do with thumb. In ARM asm "s" postfix causes ALU flags to be stored.

I tested it on armv7m4 target. It has -mthumb compilation flag as well.

kemonats:
> movs vs mov, adds vs add have nothing to do with thumb. In ARM asm "s" postfix causes ALU flags to be stored.
> 
> I tested it on armv7m4 target. It has -mthumb compilation flag as well.

Should the assembler code work in 32-bit and 16-bit versions?

agkaminski:
We use thumb2 for both cortex-a and cortex-m cores. It doesn't matter much, thanks to unified asm syntax

kemonats:
I will review the code and fix it.

kemonats:
I changed the use of `cbz`
I tested with: `arm7v7m3 arm7v7m4 arm7v7m7 arm7v7a7`

agkaminski:
I had to solve conflicts. Merged in 7f7f3e493a7750d30de8493b88c9ee04cd7baa72
Thank you

PR #91, opened 2020-06-05
hal/arm/exceptions.c: simpler concatenating of strings

kemonats:
f you accept these changes, I will send changes to other targets.

### DIFF
--- a/src/hal/arm/exceptions.c
+++ b/src/hal/arm/exceptions.c
@@ -57,8 +57,11 @@ static int exceptions_i2s(char *prefix, char *s, unsigned int i, unsigned char b
 	char c;
 	unsigned int l, k, m;
 
-	m = hal_strlen(prefix);
-	hal_memcpy(s, prefix, m);
+	m = 0;
+	while (*prefix) {
+		s[m] = *prefix++;
+		m++;
+	}
 
 	for (k = m, l = (unsigned int)-1; l; i /= b, l /= b) {
 		if (!zero && !i)
@@ -78,6 +81,16 @@ static int exceptions_i2s(char *prefix, char *s, unsigned int i, unsigned char b
 }
 
 
+static char *exceptions_cpy(char *s1, const char *s2)
+{
+	while ((*s1++ = *s2++))
+		;
+	s1--;
+
+	return s1;
+}
+
+
 void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
 {
 	static const char *const mnemonics[] = {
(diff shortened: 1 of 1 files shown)


src/hal/arm/exceptions.c
```diff
@@ -57,8 +57,11 @@ static int exceptions_i2s(char *prefix, char *s, unsigned int i, unsigned char b
 	char c;
 	unsigned int l, k, m;
 
-	m = hal_strlen(prefix);
-	hal_memcpy(s, prefix, m);
+	m = 0;
+	while (*prefix) {
+		s[m] = *prefix++;
+		m++;
+	}
```
agkaminski, line 64: Maybe _for_ loop fits here better? 
`for (m = 0; *prefix != '\0'; ++m)`

src/hal/arm/exceptions.c
```diff
@@ -78,6 +81,16 @@ static int exceptions_i2s(char *prefix, char *s, unsigned int i, unsigned char b
 }
 
 
+static char *exceptions_cpy(char *s1, const char *s2)
+{
+	while ((*s1++ = *s2++))
+		;
+	s1--;
+
+	return s1;
```
agkaminski, line 90: Maybe 
`return s1 - 1;` 
for compactness?

agkaminski: Anyway, should we really subtract one?
buff = exceptions_cpy(buff, "\nException: ");

*buff now contains _space_ (char after : and before '\0') in this version, which we will overwrite in next invocation of exceptions_cpy.

src/hal/arm/exceptions.c
kemonats: This code has been accepted in commit f46531644d938fb6fbe0fe730ee29203340e2573
but I can change it,
the code above and e.g.
```
for (m = 0; prefix[m] != 0; ++m)
    s[m] = prefix[m];
```
are the same ( generated by compiler )

src/hal/arm/exceptions.c
kemonats: `exceptions_cpy` is implementation of `stpcpy`

after the loop 
` while ((*s1++ = *s2++))`
the pointer s1 is set to the character after '\0', so we need to subtract one to point to the character '\0'

`return s1-1`
this version is probably bad because return value and s1 are different.
In my code I my use this functon as
`exceptiions_cpy(s1, s2)` and next use s1
or
`buf = exceptions(s1, s2)` and use buf

What do you think about adding the `hal_stpcpy` function to string.h.

agkaminski: > exceptions_cpy is implementation of stpcpy

after the loop
while ((*s1++ = *s2++))
the pointer s1 is set to the character after '\0', so we need to subtract one to point to the character '\0'

Yes, you're right, my mistake.

I think that

```
s1--;
return s1;
```

is equivalent to
`return s1 - 1;`
or
`return --s1;`

or am I missing something? We don't really care about s1 value on the end, as it is an local variable.

src/hal/arm/exceptions.c
agkaminski: I myself prefer _for_ version better ;)

src/hal/arm/exceptions.c
kemonats: Yes, you're right, my mistake, s1 is local variable
Yes, these three versions are the equivalent.

PR #92, opened 2020-06-08
hal/ia32/cpu.c: hal_cpuFeatures() - a small change of code

kemonats:
Instead of three calls `strlen ()`, one and variable.


src/hal/ia32/cpu.c
```diff
@@ -448,6 +448,7 @@ char *hal_cpuFeatures(char *features, unsigned int len)
 	u32 nb, nx, v[4], a;
 	unsigned int i = 0, overflow = 0;
 	const struct cpu_feature_t *p;
+        unsigned int ln;
```
agkaminski, line 451: Spaces used for indentation instead of tab.

PR #93, opened 2020-06-08
hal/ia32/string.h: hal_i2s() -  instead of using an array, simple code

kemonats:
The current version of the `hal_i2s()` and its use with the `-O2` flag is very inefficient.
A lot of code, little content.

My proposition: instead `static inline unsigned int hal_i2s()` use of `unsigned int hal_i2s()`
- create `string.c` and insert `hal_i2s() in it
- remove in `string.h` `hal_i2s()` and add `extern unsigned int hal_i2s()`
- adding in Makefile `string.c` to `SRCS`

### DIFF
--- a/src/hal/ia32/string.h
+++ b/src/hal/ia32/string.h
@@ -159,7 +159,6 @@ static inline char *hal_strncpy(char *dest, const char *src, size_t n)
 
 static inline unsigned int hal_i2s(char *prefix, char *s, unsigned int i, unsigned char b, char zero)
 {
-	char digits[] = "0123456789abcdef";
 	char c;
 	unsigned int l, k, m;
 
@@ -172,7 +171,8 @@ static inline unsigned int hal_i2s(char *prefix, char *s, unsigned int i, unsign
 	for (k = m, l = (unsigned int)-1; l; i /= b, l /= b) {
 		if (!zero && !i)
 			break;
-		s[k++] = digits[i % b];
+		c = i % b;
+		s[k++] = c < 10 ? c + '0' : c + ('a' - 10);
 	}
 
 	l = k--;


pawelpisarczyk:
I would like to understand the current inefficiency. In which domain the current code is inefficient?

kemonats:
I mean using `hal_i2s()` as `static inline` when compiling with the `-O2` option.

`hal_i2s()` is used in 2 functions: `hal_exceptionsDumpContext()` and `hal_cpuInfo()`.
In the `hal_exceptionsDumpContext()` function, `hal_i2s()` is called 23 times.
The compiler during optimization `-O2` inserts the code `hal_i2s()` 23 times.
e.g
array `char digits []` is copied to the stack 23 times
disassemble the code
```
c7 44 24 17 30 31 32 33  movl  $0x33323130.0x17(% esp)
c7 44 24 1b 34 35 36 37  movl  $0x37363534.0x1b(% esp)
c7 44 24 1f 38 39 61 62  movl  $0x62613938,0x1f(% esp)
c7 44 24 23 63 64 65 66  movl  $0x66656463,0x23(% esp)
```
`8 bytes x 4 x (23 - 1) = 704` bytes of redundant code
For example, this was what I meant by inefficiency.

pawelpisarczyk:
Your modifications don't solve this problem.  You propose to change the simple indexing into conditional instruction and some arithmetics. I doubt if this code will be more efficient in terms of size (maybe for 16 bytes), but I have no doubts that it will be more inefficient because of branch and it is not  so clear for potential reader. If we would like to save the space on IA32 the best idea is to change the inline function into the regular function to prevent the code copying.

pawelpisarczyk:
You propose this solution in the comment to your PR - I noticed this, bu I would like to clarify.

kemonats:
Of course, my modification will be less efficient when it comes to speed - more operations in the loop.
But when it comes to the size of the code, it looks like this:
```
current version:
hal_exceptionsDumpContext 4598
hal_cpuInfo 1268
phoenix-ia32.img 150600

my modification:
hal_exceptionsDumpContext 3560
hal_cpuInfo 1108
phoenix-ia32.img 149416

difference over 1kb
```

What do you think about use
```
static character digits[]
```

agkaminski:
I think `static const char digits[]` would be best.
Difference in code side won't matter if we remove `inline`.
Also, original should be smaller, much less operation to index an array. Perhaps change in asm size comes from need to initialize `digits` in code? If we use static const, it should be smaller than arithmetics version.

pawelpisarczyk:
I agree with you. The best solution is to use static const char for digits[].

kemonats:
For information, code size using `static const char digits[]`
```
hal_exceptionsDumpContext 3558
hal_cpuInfo  1108
phoenix-ia32.img 149480
```

agkaminski:
So it seems that `static const char digits` and ditching `inline` is the best approach overall

pawelpisarczyk:
Could you propose a solution with static cons as PR?

pawelpisarczyk:
Idea of changes are accepted, but new PR should be provided.

PR #95, opened 2020-06-17
hal/ia32/string.h: hal_i2s() - use static const char[] instead char[]

kemonats:
I don't know if it's about such a modification or using `hal_i2s ()` as a normal function.


agkaminski:
It's about both modifications at the same time (both `static const` and removing `inline`)

PR #96, opened 2020-06-18
hal/ia32/string: hal_i2s() as a normal function

kemonats:


kemonats:
I am not a co-author of the code, I just copied the function code from one file to another and changed the type of the variable.

PR #99, opened 2020-06-27
hal/ia32/_init.S: removed assembler warnings

kemonats:
Removed assembler warnings:
```
_init.S:312: Warning: indirect jmp without `*'
_init.S:340: Warning: indirect call without `*'
```


Maxez:
Both warnings occur because of invalid [indirect branch](https://en.wikipedia.org/wiki/Indirect_branch) in AT&T syntax, so I agree with the change (the code is correct in Intel syntax).
The compiler treats this code as if * was there anyway, producing the same opcodes with or without it:
image

PR #102, opened 2020-07-03
ia32/page.c fixed _page_showPages() adding empty lines

Maxez:
After writing a character at column 80 the console automatically moves cursor to the next line, causing pages not to display properly (empty line is added):
fix-showPages
This PR fixes the problem.

### DIFF
--- a/src/vm/page.c
+++ b/src/vm/page.c
@@ -270,9 +270,11 @@ void _page_showPages(void)
 		/* Print markers in case of memory gap */
 		if (p->addr > a) {
 			if ((rep = (p->addr - a) / SIZE_PAGE) >= 4) {
-				k = page_digits(rep, 10);
-				if ((w += k + 3) > 80) {
-					lib_printf("\nvm: ");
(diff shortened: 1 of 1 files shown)


agkaminski:
It is fixing behaviour on pc-tty, but has it been tested on pc-uart too?

Maxez:
> It is fixing behaviour on pc-tty, but has it been tested on pc-uart too?
We're printing pages using kernel console (we don't need pc-tty or pc-uart). On IA32 VGA kernel console moves cursor to the next line once 80 character line limit has been reached. There is also serial console, in which case the receiver terminal determines when (and if at all) a line gets wrapped around.
I see 2 simple options to deal with the problem:
1. We assume that the terminal we use works in 80x25 mode with autowrap enabled (the way VGA kernel console works on IA32).
2. We write 79 characters to the console so that autowrap doesn't trigger. This way we control when cursor moves to the next line. The drawback is that we lose one character space on the screen.

PR #103, opened 2020-07-12
main.c: change condition

kemonats:
The second argument 
```
*cmdline == 'X' && ++xcount
```
is always true.
Such code will probably look better.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -74,7 +74,8 @@ void main_initthr(void *unused)
 		while (*end && *(++end) != ' ');
 		while (*end && *end == ' ')
 			*(end++) = 0;
-		if (*cmdline == 'X' && ++xcount) {
+		if (*cmdline == 'X') {
+			++xcount;
 			argend = cmdline;
 
 			for (i = 0; i < sizeof(argv) / sizeof(*argv) - 1; ++i) {


pawelpisarczyk:
This code is correct. It is specific notation which allow you to increment xcount before entering into the conditional statement when first comparison is true.

kemonats:
You closed this PR but compiler does this like that:
```
		if (*cmdline == 'X' && ++xcount) {
 10d:	80 fa 58             	cmp    $0x58,%dl
 110:	74 18                	je     12a <main_initthr+0x12a>
						lib_printf("main: failed to spawn %s (%d)\n", argv[0], res);
					}
				}
			}
		}
		else if (*cmdline == 'M') {
 112:	80 fa 4d             	cmp    $0x4d,%dl
 115:	0f 84 5c 01 00 00    	je     277 <main_initthr+0x277>
	while (cmdline != NULL && *cmdline != '\0') {
 11b:	0f b6 13             	movzbl (%ebx),%edx
 11e:	84 d2                	test   %dl,%dl
 120:	0f 84 3e 01 00 00    	je     264 <main_initthr+0x264>
 126:	89 dd                	mov    %ebx,%ebp
 128:	eb b7                	jmp    e1 <main_initthr+0xe1>
		if (*cmdline == 'X' && ++xcount) {
 12a:	83 44 24 14 01       	addl   $0x1,0x14(%esp)
 12f:	89 e8                	mov    %ebp,%eax
```
so check first and then increase.

PR #104, opened 2020-07-12
programs.c: added conversion error check

kemonats:
Comparing unsigned int with negative value looks not very good but `programs_a2i()` returns this value.
I don't know if it will be better in this case `continue` or `return -EINVAL`
In addition, the value returned by `programs_decode()` is not checked in `main_initthr()`.


pawelpisarczyk:
Better idea is to return the -EINVAL instead of continuing parsing. fs is used to jump to the next program in the archive.

PR #106, opened 2020-07-14
syspage: fix parsing syspage progs argsi

Cinek28:
The last argument in argv table is always being
set to NULL. Fixed for NULL to be set after the
arguments.


Cinek28:
The previous code always overwrites the last argument (for example a program called with "prog_name;-u;-a" will be launched as "prog_name;-u"). This error appeared now because the 'while' loop was changed to 'for' (previously 'i' was incremented inside 'while' before the break). My fix is insufficient too, because when it happens to be 32 args, NULL would be assigned to argv[32] (which will segfault). I brought back previously working version (with while loop).

PR #116, opened 2020-08-13
Makefile: adding directory creation

kemonats:


agkaminski:
This code won't work on OSX. It's mkdir don't support -p option, sadly

PR #118, opened 2020-08-17
define architecture stack alignment and data type held in register

youki0:


pawelpisarczyk:
This proposal is overdesigned. It makes no sense to define new macro in every HAL. PUTONSTACK for specific architecture takes care on proper alignment. Alignment in copying could be preserved using sizeof(long) instead of sizeof(int).

PR #126, opened 2020-09-08
programs.c: other check of HAL name

kemonats:
The `HAL` value is known at compile time and instead of `hal_strcmp()` use `__builtin_strcmp()`.

### Other solutions
- **define variable in `Makefile`**

`Makefile`
```
ifeq ($(TARGET_SUFF), ia32)
CFLAGS += -DIA32_QEMU
endif
```
`programs.c`
```
#if defined(IA32_QEMU)
        cpio = (void *)(programs - 0x1000);
#endif
```
- **use predefined `__i386__`**

`programs.c`
```
#if defined(__ i386__)
        cpio = (void *)(programs - 0x1000);
#endif
```

### DIFF
--- a/programs.c
+++ b/programs.c
@@ -60,7 +60,7 @@ int programs_decode(vm_map_t *kmap, vm_object_t *kernel)
 	if (hal_strncmp(cpio->c_magic, "070701", 6)) {
 
 		/* Happens in QEMU when programs are not page aligned. What the hell? */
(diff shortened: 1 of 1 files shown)


pawelpisarczyk:
This is not good idea because it introduces new define. I try to minimize defines because they make source code more chaotic. Other defines will be consequently removed in the next months from hardware independent kernel part.

PR #140, opened 2020-09-24
armv7: rename to armv7m; provide separate directory for stm32 and imxrt.

HBuczynski:


pawelpisarczyk:
Hubert, I think that I should accept your PR, but some changes should be introduced. I think that to many #defines are introduced. Alek, what is your opinion?

hal/armv7m/cpu.c
```diff
@@ -19,14 +19,20 @@
...
-#include "imxrt.h"
+#if defined(CPU_IMXRT105X) || defined(CPU_IMXRT106X)
+#include "imxrt10xx.h"
 #endif
 
+#ifdef CPU_IMXRT117X
+#include "imxrt117x.h"
+#endif
+
+
```
agkaminski, line 35: This logic could be hidden inside some header file. mcu.h maybe? This would make this #ifs to be in one place only

hal/armv7m/cpu.c
```diff
@@ -228,7 +234,7 @@ void _hal_cpuInit(void)
 
 #ifdef CPU_STM32
 	_stm32_platformInit();
-#elif CPU_IMXRT
+#elif defined(CPU_IMXRT)
 	_imxrt_platformInit();
 #endif
 }
```
agkaminski, line 240: Maybe this functions should be implemented in platform specific file instead?

hal/armv7m/cpu.h
```diff
@@ -16,6 +16,16 @@
 #ifndef _HAL_CPU_H_
 #define _HAL_CPU_H_
 
+
+#if defined(CPU_STM32L152XD) || defined(CPU_STM32L152XE) || defined(CPU_STM32L4X6)
+#define CPU_STM32
```
agkaminski, line 21: We could do #include "stm32.h" here

hal/armv7m/imxrt/10xx/console.c
```diff
@@ -19,9 +19,9 @@
 
 #include "cpu.h"
 #include "console.h"
-#include "imxrt.h"
-#include "../../include/errno.h"
-#include "../../include/arch/imxrt.h"
+#include "imxrt10xx.h"
+#include "../../../../include/errno.h"
+#include "../../../../include/arch/imxrt.h"
```
agkaminski, line 24: Eh. Maybe we should add -I include directory to kernel CFLAGS. @pawelpisarczyk what do you think?

hal/armv7m/imxrt/interrupts.c
```diff
@@ -18,14 +18,17 @@
...
-#include "imxrt1170.h"
+#ifdef CPU_IMXRT117X
 #define SIZE_INTERRUPTS 217
-#else
-#include "imxrt.h"
+#include "imxrt117x.h"
+#endif
+
+#if defined(CPU_IMXRT105X) || defined(CPU_IMXRT106X)
 #define SIZE_INTERRUPTS 167
```
agkaminski, line 27: Maybe we should move SIZE_INTERRUPTS to the platform header

hal/armv7m/imxrt/syspage.c
```diff
@@ -0,0 +1,42 @@
...
+		progsz = MAX_PROGSZ;
+
+	hal_memcpy(_syspage_store, syspage, sizeof(syspage_t) + progsz * sizeof(syspage_program_t));
+	hal_strncpy(_syspage_arg, syspage->arg, sizeof(_syspage_arg));
+	_syspage_arg[sizeof(_syspage_arg) - 1] = '\0';
+
+	syspage = (void *)_syspage_store;
+	syspage->progssz = progsz;
+	syspage->arg = _syspage_arg;
+}
```
agkaminski: Is syspage copying needed for imxrt? It's already placed in RAM by plo, so I don't think so

hal/armv7m/stm32/interrupts.c
```diff
@@ -19,15 +19,20 @@
...
-#ifdef CPU_STM32L1
+
+#if defined(CPU_STM32L152XD) || defined(CPU_STM32L152XE)
 #define SIZE_INTERRUPTS     84
-#else
+#endif
+
+#ifdef CPU_STM32L4X6
 #define SIZE_INTERRUPTS     97
 #endif
```
agkaminski, line 33: As with imxrt, maybe we can hide this in platform specific include

hal/armv7m/cpu.c
pawelpisarczyk: I agree with you. This function should be implemented in platform specific file and it should be named using common name e.g. _platform_init().

hal/armv7m/cpu.c
pawelpisarczyk: This logic should be implemented only in hal.h. I agree with you that #ifs should be located in one place only.

hal/armv7m/cpu.h
pawelpisarczyk: I don't understand why we need this CPU_STM32 macro. I suggest to define TARGET macro in CFLAGS definition. This macro should be based on the same notation as TARGET definition i.e. TARGET_ARMV7M4_STM32L152XD for armv7m4-stm32l52xd.

hal/armv7m/imxrt/10xx/console.c
pawelpisarczyk: Let's leave it as is.

hal/armv7m/imxrt/interrupts.c
pawelpisarczyk: I agree with you. The right place for this is the platform header.

hal/armv7m/imxrt/syspage.c
pawelpisarczyk: I had a discussion with Hubert and the conclusion was that syspage copying should be removed.

hal/armv7m/stm32/interrupts.c
pawelpisarczyk: Definitely

pawelpisarczyk:
Alek, maybe I should accept current changes and provide modifications?

agkaminski:
@pawelpisarczyk Ok, if this code works, we can polish it later. But if we close this PR now, we will lose track of our comments about things to improve. So I think it would be best, to make this changes before merging

pawelpisarczyk:
Alek, but I think that Hubert is unable to make changes in one step.

agkaminski:
Ok, if we won't forget about things to do :)

pawelpisarczyk:
Hubert, your opinion?

hal/armv7m/cpu.c
HBuczynski: I agree with you. However I would like to do these changes in the next PR.

hal/armv7m/cpu.h
HBuczynski: `We could do #include "stm32.h" here`
I tried to include **stm32.h** in this header. However it has occurred that **stm32.h** includes **cpu.h** and there is a loop with headers. It is the main reason why I divided it.

`I don't understand why we need this CPU_STM32 macro`
CPU_STM32 macro is used in many places in kernel i.e.: programs.c, proc/threads.c, syscalls.c. We can get rid of it but the code of the kernel has to be changed. I didn't want to make such big changes in a single PR.

`I suggest to define TARGET macro in CFLAGS definition.`
Ok.

hal/armv7m/imxrt/syspage.c
HBuczynski: In the current version of code, the syspage copying is needed for imxrt. If I remove this function in this PR, the code will not work.
In the next PR, where I will introduce multimaps, this function will be removed for imxrt.

HBuczynski:
I think we can merge this PR in current state. I took note related to your comments and I will provide these changes in the next PRs.
Alek, there is a history of my previous PRs. So even if it is merged I can see the comments.

PR #144, opened 2020-10-02
userintr.c: removed compiler warning for TARGET != RISCV64

kemonats:


pawelpisarczyk:
I did the same in parallel, so I close this PR.

PR #149, opened 2020-10-07
armv7m: add hal_interruptsFeatures() for both imxrt and stm32.

HBuczynski:


hal/armv7m/imxrt/interrupts.c
```diff
@@ -147,6 +147,15 @@ int hal_interruptsDeleteHandler(intr_handler_t *h)
 }
 
 
+char *hal_interruptsFeatures(char *features, unsigned int len)
+{
+	hal_memcpy(features, "Using NVIC interrupt controller", min(32, len));
```
agkaminski, line 152: If we used hal_strncpy here we wouldn't need to count length of the string manually

PR #151, opened 2020-10-09
riscv64: fixed Qemu 5.1.50 DTB parsing, added helper function for pri…

Maxez:
…nting DTB nodes

### DIFF
--- a/hal/riscv64/dtb.c
+++ b/hal/riscv64/dtb.c
@@ -129,13 +129,34 @@ int dtb_parseMemory(void *dtb, u32 si, u32 l)
 }
 
 
+static void dtb_print(char *s) {
+	while (*s != 0) {
+		__asm__ (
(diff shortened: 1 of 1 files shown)


hal/riscv64/dtb.c
```diff
@@ -144,31 +165,35 @@ void dtb_parse(void *arg, void *dtb)
...
 		if (token == 1) {
+			// char buff[2] = " ";
+			// dtb_print(dtb);
+			// dtb_print(" ");
+			// buff[0] = '0' + d;
+			// dtb_print(buff);
+			// dtb_print(" ");
+			// buff[0] = '0' + state;
+			// dtb_print(buff);
+			// dtb_print("\n");
```
Maxez, line 184: Helper code for printing DTB nodes (in format: node depth state).

PR #156, opened 2020-11-24
multimaps: fix getting shared maps and bss max adress calculation for imxrt10xx

HBuczynski:


vm/map.c
```diff
@@ -1052,10 +1052,10 @@ void vm_mapGetStats(size_t *allocsz)
 vm_map_t *vm_getSharedMap(syspage_program_t *prog)
 {
 #ifdef NOMMU
-	if (prog->dmap <= 0 || prog->dmap > 16)
+	if (prog->dmap < 0 || prog->dmap > 16)
```
agkaminski, line 1055: prog->dmap >= 16)

PR #157, opened 2020-11-25
proc-nommu: get stack size from program's elf header

HBuczynski:


proc/process.c
```diff
@@ -559,6 +560,10 @@ int process_load(process_t *process, vm_object_t *o, offs_t base, size_t size, v
 	hal_memset(reloc, 0, sizeof(reloc));
 
 	for (i = 0, j = 0, phdr = (void *)ehdr + ehdr->e_phoff; i < ehdr->e_phnum; i++, phdr++) {
+
+		if (phdr->p_type == PT_GNU_STACK)
+			gnustacksz = phdr->p_memsz;
```
agkaminski, line 565: Align this size to page size, just in case.

proc/process.c
```diff
@@ -546,6 +546,7 @@ int process_load(process_t *process, vm_object_t *o, offs_t base, size_t size, v
 	char *snameTab;
 	ptr_t *got;
 	struct _reloc reloc[5];
+	size_t stacksz, gnustacksz = 0;
```
agkaminski, line 549: `size_t stacksz = SIZE_USTACK`, then we don't need `gnustacksz ` variable

proc/process.c
HBuczynski: Done

proc/process.c
HBuczynski: Done

PR #161, opened 2020-12-09
hal/ia32: fixes for PCI enumeration + code cleanup

Maxez:
Fixed PCI enumeration (resetting `d` and `f` indexes). Changed `b`, `d` and `f` indexes type to unsigned int due to `b` index overflow.

### DIFF
--- a/hal/ia32/pci.c
+++ b/hal/ia32/pci.c
@@ -136,7 +136,7 @@ int hal_pciGetDevice(pci_id_t *id, pci_dev_t *dev, void *caps)
 
 				val = _hal_pciGet(b, d, f, 0xb);
 
-				if ((id->subdevice != PCI_ANY) && (id->subdevice != ((val >> 16) & 0xffff)))
+				if ((id->subdevice != PCI_ANY) && (id->subdevice != (val >> 16)))
 					continue;
 
 				if ((id->subvendor != PCI_ANY) && (id->subvendor != (val & 0xffff)))
@@ -149,7 +149,7 @@ int hal_pciGetDevice(pci_id_t *id, pci_dev_t *dev, void *caps)
 				dev->device = dv >> 16;
 				dev->cl = cl;
 				dev->subvendor = val & 0xffff;
-				dev->subdevice = (val >> 16) & 0xffff;
+				dev->subdevice = val >> 16;
 
 				dv = _hal_pciGet(b, d, f, 1);
 				dev->status = dv >> 16;
--- a/hal/ia32/pci.c
+++ b/hal/ia32/pci.c
@@ -108,86 +108,91 @@ int hal_pciSetBusmaster(pci_dev_t *dev, u8 enable)
 int hal_pciGetDevice(pci_id_t *id, pci_dev_t *dev, void *caps)
 {
 	spinlock_ctx_t sc;
-	unsigned int b, d, f, i;
+	unsigned char b, d, f, i;
 	u32 dv, cl, mask, val;
 	int res = EOK;
 
 	if ((id == NULL) || (dev == NULL))
 		return -EINVAL;
 
-	hal_spinlockSet(&pci_common.spinlock, &sc);
-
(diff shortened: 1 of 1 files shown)


hal/ia32/pci.c
```diff
@@ -122,21 +108,19 @@ int hal_pciSetBusmaster(pci_dev_t *dev, u8 enable)
 int hal_pciGetDevice(pci_id_t *id, pci_dev_t *dev, void *caps)
 {
 	spinlock_ctx_t sc;
-	unsigned char b, d, f, i;
+	unsigned int b, d, f, i;
```
agkaminski, line 111: Why change to `unsigned int`? `_hal_pciGet` still gets arguments of `u8` type.

hal/ia32/pci.c
```diff
@@ -122,21 +108,19 @@ int hal_pciSetBusmaster(pci_dev_t *dev, u8 enable)
...
 	u32 dv, cl, mask, val;
 	int res = EOK;
 
 	if ((id == NULL) || (dev == NULL))
 		return -EINVAL;
 
+	hal_spinlockSet(&pci_common.spinlock, &sc);
```
agkaminski, line 118: for 0...255 (for 0 to 32 (for 0 to 8))) That's quite some time to keep spinlock on. Do we need to keep spinlock this long?

hal/ia32/pci.c
Maxez: The `b` index in `for (b = dev->bus; b < 256; b++)` being unsigned char caused infinite loop if no device was detected. Making it unsigned int solves the issue.

hal/ia32/pci.c
Maxez: I changed setting the spinlock for whole PCI bus enumeration for better readability (I figured it's not that much time). If you think it's too much I can restore setting the spinlock for single PCI configuration space access.

hal/ia32/pci.c
```diff
@@ -145,12 +129,12 @@ int hal_pciGetDevice(pci_id_t *id, pci_dev_t *dev, void *caps)
...
-				cl = hal_pciGet(b, d, f, 2) >> 16;
+				cl = _hal_pciGet(b, d, f, 2) >> 16;
 
 				if ((id->cl != PCI_ANY) && (id->cl != cl))
 					continue;
 
-				val = hal_pciGet(b, d, f, 0xb);
+				val = _hal_pciGet(b, d, f, 0xb);
 
 				if ((id->subdevice != PCI_ANY) && (id->subdevice != ((val >> 16) & 0xffff)))
```
kemonats, line 139: The mask is probably unnecessary.
val is `unsigned int`
after shift right by 16 it will be `0000xxxx`.

hal/ia32/pci.c
```diff
@@ -167,8 +151,6 @@ int hal_pciGetDevice(pci_id_t *id, pci_dev_t *dev, void *caps)
 				dev->subvendor = val & 0xffff;
 				dev->subdevice = (val >> 16) & 0xffff;
```
kemonats, line 152: Same as above.

hal/ia32/pci.c
Maxez: Fixed in next commit, thanks.

hal/ia32/pci.c
Maxez: Fixed.

hal/ia32/pci.c
```diff
@@ -200,9 +182,13 @@ int hal_pciGetDevice(pci_id_t *id, pci_dev_t *dev, void *caps)
```
kemonats, line 182: `hal_spinlockClear()` should probably be removed.

hal/ia32/pci.c
kemonats: Maybe hal_spinlockSet() set before:
```
for (d = dev->dev; d < 32; d++) 
```
or before
```
for (f = dev->func; f < 8; f++)
```

PR #163, opened 2020-12-11
armv7m: align user stack to 8 bits

HBuczynski:


hal/armv7m/cpu.c
```diff
@@ -57,6 +57,9 @@ int hal_cpuCreateContext(cpu_context_t **nctx, void *start, void *kstack, size_t
 	if (kstacksz < sizeof(cpu_context_t))
 		return -1;
 
+	/* Align user stack to 8 bits */
```
agkaminski, line 60: bytes*

agkaminski (changes requested):

PR #167, opened 2020-12-14
proc/nommu: add offset for allocated segment

HBuczynski:


proc/process.c
```diff
@@ -582,17 +583,20 @@ int process_load(process_t *process, vm_object_t *o, offs_t base, size_t size, v
 		if (phdr->p_flags & PF_W) {
 			prot |= PROT_WRITE;
 
-			if ((paddr = vm_mmap(process->mapp, NULL, NULL, round_page(phdr->p_memsz), prot, NULL, -1, flags)) == NULL)
+			reloffs = phdr->p_offset % SIZE_PAGE;
```
agkaminski, line 586: I think we should use here `p_vaddr` here instead of `p_offset`. IMHO `p_vaddr` gives us target place in memory, `p_offset` gives us offset in .elf file.

@nalajcie What do you think?

HBuczynski: I have made another test. I declared variables in **.bss** or **.data** sections and align them to 8, 16, 32, 64, 256, 512. Test proved that `p_vaddr` should be used. When `reloffs` was calculated using `p_offset`, some variables were not aligned correctly.

nalajcie: yes, for alignment <= `max-page-size` it would not matter (as `p_vaddr % max-page-size == p_offset`), but for bigger one there might be differences (should not matter if we would set `max-page-size` to `SIZE_PAGE`). The `p_vaddr` approach is definitely more correct.

PR #178, opened 2021-02-05
Enable jtag/debuger

gerard5:
This commit is analogous to https://github.com/phoenix-rtos/plo/pull/26 because of the same procedure.

* The debugger functionality has been checked against to use J-Link (jtag / swd) and [pyOCD](https://github.com/pyocd/pyOCD) (using CMSIS-DAP available on the mimixrt1064-evk board).
* Also a minor update in the comment about clock freq.


agkaminski (approved):
Waiting for confirmation from @Cinek28 that this PR solves the problem.

PR #180, opened 2021-02-18
test: add unity test framework

jsarzy:


test/bsearch/main.c
```diff
@@ -0,0 +1,80 @@
...
+{
+}
+
+TEST_TEAR_DOWN(test_bsearch)
+{
+}
+
+TEST(test_bsearch, basic)
+{
+	int arr_inc[] = {1, 2, 3, 4, 5, 6, 7, 8};
```
nalajcie, line 38: `const int` (the same apples everywhere else)

test/bsearch/main.c
```diff
@@ -0,0 +1,80 @@
...
+	int arr_dec[] = {7, 6, 5, 4, 3, 2, 1};
+	int arr_single[] = {1};
+
+	for (int key = 1; key < 8; key++) {
```
nalajcie, line 42: `<=`, You're never searching for `8`

test/bsearch/main.c
```diff
@@ -0,0 +1,80 @@
...
+		void *res = lib_bsearch(&key, arr_inc, 8, sizeof(int), cmp_gt);
```
nalajcie, line 43: `sizeof(arr_inc) / sizeof(arr_inc[0])`, better than magic number

test/bsearch/main.c
```diff
@@ -0,0 +1,80 @@
...
+		TEST_ASSERT_EQUAL_PTR(&arr_inc[key-1], res);
+	}
+
+	for (int key = 7; key > 1; key--) {
```
nalajcie, line 47: never searching for `1`

test/sched/main.c
```diff
@@ -0,0 +1,90 @@
...
+	actual_prio = -1;
+
+	for (unsigned int p = 0; p < THREADS_MAX_PRIO; p++) {
+		int res = proc_threadCreate(NULL, thread_prio, NULL, p, stacksz, NULL, 0, (void *)p);
+		TEST_ASSERT_EQUAL_INT(EOK, res);
+	}
+
+	for (unsigned int p = 0; p < THREADS_MAX_PRIO; p++) {
+		proc_reap();
+	}
```
nalajcie, line 42: maybe we could check `actual_prio` value at the test end also?

test/test_main.c
```diff
@@ -0,0 +1,24 @@
+#include "test.h"
+#include "proc/threads.h"
+
+DECLARE_TEST_GROUP(test_list);
```
nalajcie, line 4: we should probably comment that it's necessary to explicitly declare new test groups here

test/Makefile
```diff
@@ -5,5 +5,21 @@
...
+KERNEL_TEST_GROUPS := $(subst $(COMMA),$(SPACE),$(KERNEL_TEST_GROUPS))
+
+CFLAGS += -DKERNEL_TEST
+CFLAGS += -I$(PREFIX_H)unity
+
+UNITYLIB := $(PREFIX_A)unity.a
+
+OBJS += $(PREFIX_O)test/test_main.o
+
+include $(addprefix test/, $(addsuffix /Makefile, $(KERNEL_TEST_GROUPS)))
```
nalajcie, line 21: I see that using `,` as a separator is quite confusing in `Makeflie`. We can leave it as is, but I would welcome some way to compile and run all tests (would `KERNEL_TEST_GROUPS=*` work?)

nalajcie (changes requested):

test/bsearch/main.c
kemonats: better `static const int`

PR #183, opened 2021-02-22
main.c: removed redundant condition

kemonats:


agkaminski:
It's not redundant. It's guarding against going past \0.

kemonats:
Example
```
*end    *end != '\0'    *end == ' '     *end && *end == ' '
' '       1               1               1
'a'       1               0               0
'\0'      0               0               0

```
You can see that checking `*end != '\0'` is redundant.

agkaminski:
Right, sorry.

PR #186, opened 2021-03-09
initialization of Zynq-7000

HBuczynski:
The following components are under development:
- _hal.S, 
- _interrupts.S,
- gic,
- timer,
- pmap,
- syspage.

### DIFF
--- a/hal/armv7a/_hal.S
+++ b/hal/armv7a/_hal.S
@@ -1,0 +1,224 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * Low-level hal functions for Cortex-A9
+ *
+ * Copyright 2018, 2020, 2021 Phoenix Systems
+ * Author: Pawel Pisarczyk, Aleksander Kaminski, Maciej Purski, Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#define __ASSEMBLY__
+
+#include "cpu.h"
+
+.arm
+
+.globl hal_cpuGetCycles
+.type hal_cpuGetCycles, %function
+hal_cpuGetCycles:
+	mrc p15, 0, r1, c9, c13, 0
+	str r1, [r0]
+	bx lr
+.size hal_cpuGetCycles, .-hal_cpuGetCycles
+
+
+.globl hal_cpuInvalDataCache
+.type hal_cpuInvalDataCache, %function
+hal_cpuInvalDataCache:
+	ldr r1, =SIZE_CACHE_LINE - 1
+	bic r0, r0, r1
+	mcr p15, 0, r0, c7, c6, 1
+	bx lr
+.size hal_cpuInvalDataCache, .-hal_cpuInvalDataCache
+
+
+.globl hal_cpuFlushDataCache
+.type hal_cpuFlushDataCache, %function
+hal_cpuFlushDataCache:
+	ldr r1, =SIZE_CACHE_LINE - 1
+	bic r0, r0, r1
+	mcr p15, 0, r0, c7, c14, 1
+	bx lr
+.size hal_cpuFlushDataCache, .-hal_cpuFlushDataCache
+
+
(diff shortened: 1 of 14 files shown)


hal/armv7a/_hal.S
```diff
@@ -0,0 +1,220 @@
...
+ * Low-level hal functions for Cortex-A9
+ *
+ * Copyright 2018, 2020, 2021 Phoenix Systems
+ * Author: Pawel Pisarczyk, Aleksander Kaminski, Maciej Purski, Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
```
agkaminski, line 15: We need to make sure that this file is compiled to ARM asm. Rest of the kernel should be in thumb2. So we need .arm directive here

hal/armv7a/_interrupts.S
```diff
@@ -0,0 +1,281 @@
...
+ * Interrupts handlers for Cortex-A9
+ *
+ * Copyright 2018, 2020, 2021 Phoenix Systems
+ * Author: Pawel Pisarczyk, Aleksander Kaminski, Maciej Purski, Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
```
agkaminski, line 15: We might need need to make sure that this file is compiled to ARM asm.

hal/armv7a/zynq7000/_init.S
```diff
@@ -0,0 +1,322 @@
...
+	orr r0, r0, #((0x3 << 22) | (0x3 << 20))  /* Set CP11 and CP10: Privileged and User mode access */
+	mcr p15, 0, r0, c1, c0, 2                 /* Write R0 to CPACR                                  */
+	vmrs r0, fpexc
+	orr r0, r0, #(0x1 << 30)                  /* FPU enable bit                                     */
+	vmsr fpexc, r0
+
+	ldr pc, =main
+
+#include "../_interrupts.S"
+#include "../_hal.S"
```
agkaminski, line 322: We can include this files by compiling them and linking together, it's somewhat cleaner solution. Especially that _interrupts.S and _hal.S are common files

hal/armv7a/_hal.S
HBuczynski: Done

hal/armv7a/_interrupts.S
HBuczynski: Done

hal/armv7a/zynq7000/_init.S
HBuczynski: I agree with you.
However, interrupts need some change to be done. For example there is direct call to GIC controller in `_interrupts_dispatch` https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/eaffcef02fcd21966b5be870d7c4e371b9a7bfd4/hal/armv7a/imx6ull/_init.S#L859-L877
It should be moved to interrupt controller in imx6ull and Zynq-7000. 
When it will be done, I will add them to compilation process and remove `include`

hal/armv7a/zynq7000/pmap.c
```diff
@@ -0,0 +1,584 @@
...
+
+	pmap->start = (void *)VADDR_KERNEL;
+	pmap->end = (void *)VADDR_MAX;
+
+	/* Initialize kernel heap start address */
+	(*vstart) = (void *)(((u32)_end + SIZE_PAGE - 1) & ~(SIZE_PAGE - 1));
+
+	/* First 15 pages after bss are mapped for UART1, UART2, GIC, GPT1, CCM, IOMUX, WDOG and SRC */
+	(*vstart) += 16 * SIZE_PAGE;
+	(*vend) = (*vstart) + SIZE_PAGE;
```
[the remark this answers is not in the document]
HBuczynski: Currently pmap isn't implemented for Zynq so it was copied from imx6ull without changes to make compilation successful. Taking into account that it will be changed in the future I haven't put it into a common directory.

PR #206, opened 2021-06-29
posix: add gethostname/sethostname syscalls

ziemleszcz:
JIRA: DTR-12

## Description

Implement gethostname/sethostname syscalls.

## Motivation and Context

Required to get hostname command working.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

Requires https://github.com/phoenix-rtos/libphoenix/pull/102.

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -16,6 +16,7 @@
 #include HAL
 #include "../include/errno.h"
 #include "../include/ioctl.h"
+#include "../include/limits.h"
 #include "../proc/proc.h"
 
 #include "posix.h"
@@ -62,6 +63,7 @@ struct {
 	rbtree_t pid;
 	lock_t lock;
 	id_t fresh;
+	char hostname[HOST_NAME_MAX + 1];
 } posix_common;
 
 
@@ -1626,6 +1628,19 @@ int posix_connect(int socket, const struct sockaddr *address, socklen_t address_
 }
 
 
+int posix_gethostname(char *name, size_t namelen)
+{
+	TRACE("gethostname(%zu)", namelen);
+
+	if (namelen < 0)
+		return -EINVAL;
+
+	hal_strncpy(name, posix_common.hostname, namelen);
+
+	return 0;
+}
+
+
 int posix_getpeername(int socket, struct sockaddr *address, socklen_t *address_len)
 {
 	TRACE("getpeername(%d, %s)", socket, address == NULL ? NULL : address->sa_data);
@@ -1812,6 +1827,20 @@ int posix_shutdown(int socket, int how)
 	}
 
 	return err;
+}
+
+
+int posix_sethostname(const char *name, size_t namelen)
+{
+	TRACE("sethostname(%zu)", namelen);
+
+	if (namelen < 0 || namelen > HOST_NAME_MAX)
+		return -EINVAL;
+
+	hal_strncpy(posix_common.hostname, name, namelen);
+	posix_common.hostname[namelen] = '\0';
+
+	return 0;
 }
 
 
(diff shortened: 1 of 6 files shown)


posix/posix.c
```diff
@@ -1626,6 +1630,25 @@ int posix_connect(int socket, const struct sockaddr *address, socklen_t address_
...
+	if (namelen < 0)
+		return -EINVAL;
+
+	if (namelen > hostname_len) {
+		hal_memcpy(name, hostname, hostname_len);
+		name[hostname_len] = '\0';
+	}
+	else {
+		hal_memcpy(name, hostname, namelen);
+	}
```
agkaminski, line 1646: My suggestion (hostname_len not needed):

```suggestion
	hal_strncpy(name, hostname, namelen);
	name[namelen] = '\0';
```

posix/posix.c
```diff
@@ -1626,6 +1630,25 @@ int posix_connect(int socket, const struct sockaddr *address, socklen_t address_
...
+
+	if (namelen < 0)
+		return -EINVAL;
+
+	if (namelen > hostname_len) {
+		hal_memcpy(name, hostname, hostname_len);
+		name[hostname_len] = '\0';
+	}
+	else {
+		hal_memcpy(name, hostname, namelen);
```
agkaminski, line 1645: You don't set NULL terminator here (you copy only part of hostname, without '\0')

posix/posix.c
```diff
@@ -65,6 +65,10 @@ struct {
 } posix_common;
 
 
+static char hostname[HOST_NAME_MAX];
```
agkaminski, line 68: You need one more byte for NULL terminator

posix/posix.c
```diff
@@ -1815,6 +1838,20 @@ int posix_shutdown(int socket, int how)
...
 
+int posix_sethostname(const char *name, size_t namelen)
+{
+	TRACE("sethostname(%zu)", namelen);
+
+	if (namelen < 0 || namelen > HOST_NAME_MAX)
+		return -EINVAL;
+
+	hal_memcpy(hostname, name, namelen);
+	hostname_len = namelen;
```
agkaminski, line 1849: Right now NULL terminator is not copied to hostname, it's an unterminated string

```suggestion
	hal_strncpy(hostname, name, namelen);
	hostname[namelen] = '\0';
```

posix/posix.c
ziemleszcz: Your version is shorter but it always adds 0 at the end (POSIX is not clear on this, needs check for namelen == 0). It seems that ia32 hal_strncpy() doesn't handle 0 case properly, I'm not sure about other targets. Also sethostname doesn't guarantee that hostname is NULL terminated so hal_strncpy may copy all HOST_NAME_MAX bytes. I would prefer to keep this version.

posix/posix.c
ziemleszcz: It's intended, hostname_len keeps hostname length.

posix/posix.c
ziemleszcz: It's intended, hostname_len keeps hostname length.

posix/posix.c
ziemleszcz: It's intended, hostname_len keeps hostname length.

posix/posix.c
agkaminski: You can either keep one char extra for NULL terminator or 4 bytes (size_t) for string length. Code turns out better for 1st option imho, but it's a shame, that hal_strncpy() seems to be broken (it would be better to fix it rather than avoid using it).

agkaminski (approved):

PR #222, opened 2021-07-26
proc: allow sysexec to spawn programs with default map

gerard5:
## Description

A change to `proc_syspageSpawnName()` which allows to run syspage programs with the default memory map (dmap) set in plo using the `app` command.

This change additionally solves "a todo", and introduces that the write and read attributes of the selected map are checked before syspage program is launched.

JIRA: RTOS-74

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt1064).

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/80
https://github.com/phoenix-rtos/phoenix-rtos-doc/pull/12

### DIFF
--- a/main.c
+++ b/main.c
@@ -97,7 +97,7 @@ void main_initthr(void *unused)
 			for (prog = syspage->progs, i = 0; i < syspage->progssz; i++, prog++) {
 				if (!hal_strcmp(cmdline + 1, prog->cmdline)) {
 					argv[0] = prog->cmdline;
-					res = proc_syspageSpawn(prog, vm_getSharedMap(prog), prog->cmdline, argv);
+					res = proc_syspageSpawn(prog, vm_getSharedMap(prog, -1), prog->cmdline, argv);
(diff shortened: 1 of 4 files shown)


agkaminski:
You only check for map attributes when map is not given (initially map == NULL), so it solves TODO only partially

agkaminski (dismissed):
LGTM, but is it ok to modify syspage program? (not introduced by this PR, so no merge blocker)

gerard5:
> LGTM, but is it ok to modify syspage program? (not introduced by this PR, so no merge blocker)

I suggested the previous version, you are right, it should not modify syspage program, so it shall be handled differently - maybe the way like `vm_getSharedMap()`, so I will change it to draft for a moment, to fix that.

agkaminski (approved):

agkaminski (approved):

HBuczynski (approved):

PR #227, opened 2021-08-19
Fixed crash on imxrt117x

agkaminski:
## Description

Few patches that fixes various problems in kernel. It seems that undefined behaviour (taking address of packed struct member) was cause of crash when using sysexec command.

## Motivation and Context

We needed to fix the crash when using sysexec.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt117x).

## Checklist:

## Special treatment

### DIFF
--- a/main.c
+++ b/main.c
@@ -37,7 +37,7 @@ struct {
 void main_initthr(void *unused)
 {
 	size_t i;
-	syspage_program_t *prog;
+	syspage_program_t prog;
 	int xcount = 0, res;
 	char *cmdline = syspage->arg, *end;
 	char *argv[32], *arg, *argend;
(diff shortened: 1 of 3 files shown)


agkaminski:
@gerard5 I need your confirmation that this fixes the problem. I was able to reproduce the problem and fix it with above changes on my side.

gerard5:
> @gerard5 I need your confirmation that this fixes the problem. I was able to reproduce the problem and fix it with above changes on my side.

Yes, problem solved. It seems that the problem was related to *undefined behaviour when taking address of packed struct member with a copy from packed struct*.

gerard5 (approved):

Maxez (approved):

nalajcie:
During submodules update in `phoenix-rtos-project` `setjmp` test on `rt1064` starts to fail:

https://github.com/phoenix-rtos/phoenix-rtos-project/pull/174/checks?check_run_id=3384180109

I assume this PR might be the reason, @agkaminski can You check locally if it's ok?

Pasting exception in case it's not easy to repeat (not really much info as PC/LR is `null` :)):
```
	Unity test run
	Exception: 5 #BusFault
	 r0=200063c0  r1=00000000  r2=20006400  r3=00000000
	 r4=00000011  r5=20213c90  r6=2021fdf8  r7=20211e81
	 r8=202141c0  r9=2021f0e4 r10=202141b0 r11=bbbbbbbb
	r12=20006440  sp=2000f194  lr=00000000  pc=00000000
	psp=2021fc70 psr=00000000 exr=ffffffed bfa=00028000
	cfs=00008200
```

PR #231, opened 2021-08-31
hal_i2s(): change for argument with number zero

kemonats:
## Description

If number is `0` and argument `zero` is not set, only the prefix is added to the buffer.
This PR adds char `'0'` to the prefix in this case.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/exceptions.c
+++ b/hal/armv7a/exceptions.c
@@ -61,8 +61,11 @@ static int exceptions_i2s(char *prefix, char *s, unsigned int i, unsigned char b
 	hal_memcpy(s, prefix, m);
 
 	for (k = m, l = (unsigned int)-1; l; i /= b, l /= b) {
-		if (!zero && !i)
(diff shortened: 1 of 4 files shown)


agkaminski:
Isn't that the whole point of `zero` flag? This PR causes this flag to do nothing - if it's `1` we'll put `digits[i % b]` (which is `'0'`), if it's `0` we'll put also `'0'`, so it does nothing.

kemonats:
Example:
```
unsigned int hal_i2s(char *prefix, char *s, unsigned int i, unsigned char b, char zero)
```
Arguments:
```
prefix="r="
i = 0
b = 10
zero = 0
```
Current version:
```
s = "r="
```
My proposal:
```
s = "r=0"
```
In current version if `zero = 0` and `i = 0` the first time the for loop is run, it is breaked and the '0' is not added to the buffer.

PR #233, opened 2021-09-03
ia32: hal_cpuFeatures() fixed the 'nx' feature check

kemonats:
## Description

Now the '`nx'` feature is not shown for a processor that has it.
The variable `eax` of the `cpu_feature_t` structure is of type `u32`, and is never less  than `0`.
Changes:
1. Change variable `eax` to type `int`.
If the `eax` variable is to remain of type `u32`, then you need to cast the variable `eax` to `int` type .
2. Removed the `overflow` variable. If the buffer size is too small then exit the `for` loop.
3. The `'+'` char is added to the second and subsequent features.

## Types of changes

## How Has This Been Tested?

ia32-generic

## Checklist:

## Special treatment

### DIFF
--- a/hal/ia32/cpu.c
+++ b/hal/ia32/cpu.c
@@ -28,7 +28,7 @@
 
 struct cpu_feature_t {
 	const char *name;
-	u32 eax;
+	s32 eax;
 	u8 reg;
 	u8 offset; /* eax, ebx, ecx, edx */
 };
@@ -524,11 +524,10 @@ char *hal_cpuInfo(char *info)
 }
 
 
-
 char *hal_cpuFeatures(char *features, unsigned int len)
 {
 	u32 nb, nx, v[4], a;
-	unsigned int i = 0, overflow = 0;
+	unsigned int i = 0;
 	const struct cpu_feature_t *p;
 	unsigned int ln;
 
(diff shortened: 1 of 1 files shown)


agkaminski:
Sorry for taking interest this late, could you please rebase this PR so we can merge it?

gerard5 (commented):
I verified the correctness of operation on an AMD cpu on which the `nx` bit is set. 
@kemonats, please rebase the commits. I'll appreciate your pull-request be merged.

2023-03-25-221144_1013x111_scrot

Verified - ok, when `features` string overflows (manually resized [buffer)](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/1b41faf6f92bfa2a28d0fa3e04db5547348041ee/main.c#L96):
2023-03-25-223908_557x90_scrot

kemonats:
Done.
Should I delete the comment on line 542?

gerard5:
> Done. Should I delete the comment on line 542 ?

Yes, please.

gerard5 (commented):
Apart from casting to `int` instead of `s32` (see comment), the rest looks good to me. I see that this change removes the redundant (last) `+` in case the features list string does not fit in the buffer, this looks good to me.

gerard5 (approved):
Looks good to me.

agkaminski (approved):

PR #234, opened 2021-09-10
imx6ull: watchdog handling

jsarzy:
## Description

Add `wdgreload` syscall for imx6ull platform. Enable a watchdog timer in plugin.

PR requires changes in the psu scripts, so I will merge it by myself.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a-imx6ull

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/imx6ull/_init.S
+++ b/hal/armv7a/imx6ull/_init.S
@@ -88,6 +88,18 @@ plugin:
 	/* note: "ldr reg, =constant" pseudo instruction can't be used before
 	 * the serial download check below - the literal pool will not be included in the plugin */
 
+	/* init watchdog */
+	mov32 r4, 0x20bc000
+#if defined(WATCHDOG)
+	/* 128 sec timeout, debug, the clock is already enabled */
+	movw r5, #0xff36
+#else
+	/* we need to keep plugin section size the same regardless of
+	 * WATCHDOG define, so feed it with the default value */
+	movw r5, #0x0030
+#endif
+	strh r5, [r4]
+
 	/* check if reset was caused by watch dog */
 	mov32 r4, 0x20d8008
 	ldr r5, [r4]
@@ -186,9 +198,9 @@ dcd_clock:
 	.word SWAP(0x021b0000)
 	.word SWAP(0x80000000)
 
-	/* Enable DDR clock */
+	/* Enable DDR and watchdog clock */
 	.word SWAP(0x020c4074)
-	.word SWAP(0x0f300000)
+	.word SWAP(0x0f330000)
 
 	/* AIPSTZ1,2,3 clocks */
 	.word SWAP(0x020c4068)
--- a/hal/armv7a/cpu.h
+++ b/hal/armv7a/cpu.h
@@ -389,10 +389,7 @@ extern char *hal_cpuInfo(char *info);
 extern char *hal_cpuFeatures(char *features, unsigned int len);
 
 
-static inline void hal_wdgReload(void)
-{
-
-}
+extern void hal_wdgReload(void);
 
 
 extern void _hal_cpuInit(void);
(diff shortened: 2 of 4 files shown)


agkaminski (approved):
Have you tested that if WATCHDOG is not defined watchdog remains not enabled?

jsarzy:
> Have you tested that if WATCHDOG is not defined watchdog remains not enabled?

Yep, it remains not enabled.

nalajcie (approved):

hal/armv7a/imx6ull/_init.S
```diff
@@ -88,6 +88,17 @@ plugin:
 	/* note: "ldr reg, =constant" pseudo instruction can't be used before
 	 * the serial download check below - the literal pool will not be included in the plugin */
 
+	/* init watchdog */
+	mov32 r4, 0x20bc000
+#if defined(WATCHDOG)
+	/* 128 sec timeout, the clock is already enabled */
+	movw r5, #0xff34
```
nalajcie, line 95: can we change it to `#0xff36` to enable WDBG bit by default (_Watchdog DEBUG Enable. Determines the operation of the WDOG during DEBUG mode_ - 1 means Suspend WDG during DEBUG mode).

Makes debugging using JLink much easier :)

hal/armv7a/imx6ull/_init.S
```diff
@@ -88,6 +88,17 @@ plugin:
...
+#else
+	/* if watchdog is disabled feed him default value */
```
nalajcie, line 97: Maybe add comment that it's needed to keep `plugin` section size the same regardless of `WATCHDOG` define?

nalajcie (changes requested):

hal/armv7a/imx6ull/_init.S
agkaminski: Does it have any side effects? If so, we might want to do this only if `NDEBUG` is not defined

hal/armv7a/imx6ull/_init.S
agkaminski: Good idea, it might get removed if no rationale is presented

hal/armv7a/imx6ull/_init.S
nalajcie: no side effects (affects only DEBUG mode). Still I would like to have it always enabled I might want to debug userspace app regardless of kernel build version.

Alternative approach is to disable wdog in gdb (only allowed in DEBUG mode):
```gdb
set *imx6ull_common.wdog &= ~0x4
```

nalajcie:
There is also a side issue - upon enabling watchdog on `imx6ull` our current `reboot` implementation is no longer valid.

I don't want to create separate PR as this is closely related, maybe You can add the below change:
```diff
--- a/hal/armv7a/imx6ull/imx6ull.c
+++ b/hal/armv7a/imx6ull/imx6ull.c
@@ -316,7 +316,8 @@ static int _imx6ull_getIOisel(int isel, char *daisy)

 static void _imx6ull_reboot(void)
 {
-       *(imx6ull_common.wdog + wdog_wcr) |= 4;
+       /* assert SRS signal by writing 0 to bit 4 and 1 to bit 2 (WDOG enable) */
+       *(imx6ull_common.wdog + wdog_wcr) = (1 << 2);
        for (;;) ;
 }
```

jsarzy:
> There is also a side issue - upon enabling watchdog on `imx6ull` our current `reboot` implementation is no longer valid.
> 
> I don't want to create separate PR as this is closely related, maybe You can add the below change:
> 
> ```diff
> --- a/hal/armv7a/imx6ull/imx6ull.c
> +++ b/hal/armv7a/imx6ull/imx6ull.c
> @@ -316,7 +316,8 @@ static int _imx6ull_getIOisel(int isel, char *daisy)
> 
>  static void _imx6ull_reboot(void)
>  {
> -       *(imx6ull_common.wdog + wdog_wcr) |= 4;
> +       /* assert SRS signal by writing 0 to bit 4 and 1 to bit 2 (WDOG enable) */
> +       *(imx6ull_common.wdog + wdog_wcr) = (1 << 2);
>         for (;;) ;
>  }
> ```

Sure, I'll add.

nalajcie (approved):

PR #239, opened 2021-09-29
fixes for psh kill working

damianloew:
## Description

-syscalls: fixes to prevent exceptions in syscalls_signalPost
-hal: ia32: apply clang-format to cpu.c file
-hal: ia32: fix calling hal_cpuReschedule with NULL arguments

## Motivation and Context

Killing processes in using psh command returned exceptions. These changes solve problems with kill.
## Types of changes

## How Has This Been Tested?

- [x] New test added: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/55, https://github.com/phoenix-rtos/phoenix-rtos-project/pull/196).
- [x] Tested by hand on: (ia32-generic - qemu).

## Checklist:

## Special treatment

### DIFF
--- a/hal/ia32/cpu.c
+++ b/hal/ia32/cpu.c
@@ -42,11 +42,10 @@ u32 cpu_getEFLAGS(void)
 {
 	u32 eflags;
 
-	__asm__ volatile
-	(" \
+	__asm__ volatile(" \
 		pushf; \
 		popl %0"
-	: "=a" (eflags));
+					 : "=a"(eflags));
 
 	return eflags;
 }
@@ -68,7 +67,9 @@ int hal_cpuDebugGuard(u32 enable, u32 slot)
 	else
 		cpu.dr5 &= ~mask;
 
-	__asm__ volatile ("movl %0, %%dr5" : : "r" (cpu.dr5));
+	__asm__ volatile("movl %0, %%dr5"
+					 :
+					 : "r"(cpu.dr5));
 
 	return EOK;
 }
@@ -146,7 +147,7 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
 			spinlock->dmin = spinlock->e - spinlock->b;
 	}
 
-	__asm__ volatile (
+	__asm__ volatile(
 		"movl %1, %%eax;"
 		"cmp $0, %%eax;"
 		"je 1f;"
@@ -182,15 +183,15 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
 		"popl %%esp;"
 
 		"call _interrupts_multilockClear;"
- 		"pushl %%esp;"
+		"pushl %%esp;"
 
 		"jmp interrupts_popContext;"
 
 		"3:;"
 		"movl %%eax, %0"
-	: "=g" (err)
-	: "m" (spinlock), "m" (spinlock->lock), "r" (*scp), "g" (threads_schedule)
-	: "eax", "edx", "esp", "cc", "memory");
+		: "=g"(err)
+		: "m"(spinlock), "m"(spinlock->lock), "r"(*scp), "g"(threads_schedule)
+		: "eax", "edx", "esp", "cc", "memory");
 
 	return err;
 }
@@ -222,7 +223,7 @@ void _cpu_gdtInsert(unsigned int idx, u32 base, u32 limit, u32 type)
 		limit = (limit >> 12);
 
 	descrh = (base & 0xff000000) | (type & 0x00c00000) | (limit & 0x000f0000) |
-	         (type & 0x0000ff00) | ((base >> 16) & 0x000000ff);
+		(type & 0x0000ff00) | ((base >> 16) & 0x000000ff);
 
 	descrl = (base << 16) | (limit & 0xffff);
 
@@ -250,7 +251,9 @@ void *_cpu_initCore(void)
 	cpu.tss[hal_cpuGetID()].esp0 = (u32)&cpu.stacks[hal_cpuGetID()][511];
 
 	/* Set task register */
-	__asm__ volatile ("ltr %%ax" : : "a" ((4 + cpu.ncpus) * 8));
+	__asm__ volatile("ltr %%ax"
+					 :
+					 : "a"((4 + cpu.ncpus) * 8));
 
 	return (void *)cpu.tss[hal_cpuGetID()].esp0;
 }
@@ -273,7 +276,8 @@ void _hal_cpuInitCores(void)
 	for (;;) {
 		k = cpu.ncpus;
 		i = 0;
-		while ((cpu.ncpus == k) && (++i < 50000000));
+		while ((cpu.ncpus == k) && (++i < 50000000))
+			;
 		if (i >= 50000000)
 			break;
 	}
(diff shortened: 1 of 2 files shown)


hal/ia32/cpu.c
```diff
@@ -135,36 +135,16 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
...
-		"1:;"
 		"pushf;"
-
-		"2:;"
 		"pushl %%cs;"
 		"cli;"
-		"leal 3f, %%eax;"
+		"leal 1f, %%eax;"
 		"pushl %%eax;"
 		"movl $0, %%eax;"
```
kemonats, line 149:
```suggestion
		"xorl %%eax, %%eax;"
```
Faster and less bytes.

hal/ia32/cpu.c
```diff
@@ -175,21 +155,21 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
 		"pushl $0;"
 		"pushl %%eax;"
 		"pushl $0;"
-		"movl %4, %%eax;"
+		"movl %1, %%eax;"
 		"call *%%eax;"
 		"cli;"
 		"addl $12, %%esp;"
 		"popl %%esp;"
```
kemonats, line 162:
```suggestion
		"popl %%esp;"
```
Adding `12` to the `esp` and then loading the value from the stack to `esp`?

hal/ia32/cpu.c
```diff
@@ -175,21 +155,21 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
...
 		"call _interrupts_multilockClear;"
- 		"pushl %%esp;"
+		"pushl %%esp;"
 
 		"jmp interrupts_popContext;"
 
-		"3:;"
+		"1:;"
 		"movl %%eax, %0"
 	: "=g" (err)
```
kemonats, line 171:
```suggestion
	: "=a" (err)
```
You avoid contruction type:
```
mov %eax, %ecx
mov %ecx, %eax
```
If you accept this suggestion remove `eax` from `clobbers`.

hal/ia32/cpu.c
```diff
@@ -175,21 +155,21 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
...
-	: "m" (spinlock), "m" (spinlock->lock), "r" (*scp), "g" (threads_schedule)
+	: "g" (threads_schedule)
 	: "eax", "edx", "esp", "cc", "memory");
```
kemonats, line 173:
```suggestion
	: "eax", "esp", "cc", "memory");
```
`edx` is not used in the `asm` block.

kemonats (commented):
My suggestions are beyond this PR, but since you are changing the code in this function I give them.

hal/ia32/cpu.c
Maxez: I agree, added in next commit.

hal/ia32/cpu.c
Maxez: We're adding 12 to `popl` 3 arguments `pushl`ed for `threads_schedule()` call. Doing `popl %%esp` and `pushl %%esp` (after `call _interrupts_multilockClear`) seems unnecessary so I've removed it (btw. it didn't hurt as `esp` is properly saved earlier with `call interrupts_pushContext`).

hal/ia32/cpu.c
Maxez: I agree, I also removed last line `movl %%eax, %0` as it's no longer needed.

hal/ia32/cpu.c
Maxez: It's there due to cdecl calling convention used for `threads_schedule()` - `eax`, `ecx` and `edx` aren't saved by the callee-side and could be modified by `threads_schedule()`.
Added missing `ecx` in clobber list.

hal/ia32/cpu.c
```diff
@@ -68,7 +67,9 @@ int hal_cpuDebugGuard(u32 enable, u32 slot)
 	else
 		cpu.dr5 &= ~mask;
 
-	__asm__ volatile ("movl %0, %%dr5" : : "r" (cpu.dr5));
+	__asm__ volatile(
+		"movl %0, %%dr5; "
+	:: "r" (cpu.dr5));
```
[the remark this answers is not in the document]
Maxez: Using special formatting for asm code.

hal/ia32/cpu.c
```diff
@@ -135,62 +135,36 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
...
+		"cli; "
+		"addl $12, %%esp; "
+
+		"call _interrupts_multilockClear; "
+
+		"jmp interrupts_popContext; "
+		"1: "
+	: "=a" (err)
+	: "g" (threads_schedule)
+	: "ecx", "edx", "esp", "cc", "memory");
```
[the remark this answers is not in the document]
Maxez: Using special formatting for asm code.

hal/ia32/cpu.c
```diff
@@ -250,7 +224,9 @@ void *_cpu_initCore(void)
 	cpu.tss[hal_cpuGetID()].esp0 = (u32)&cpu.stacks[hal_cpuGetID()][511];
 
 	/* Set task register */
-	__asm__ volatile ("ltr %%ax" : : "a" ((4 + cpu.ncpus) * 8));
+	__asm__ volatile(
+		"ltr %%ax; "
+	:: "a" ((4 + cpu.ncpus) * 8));
```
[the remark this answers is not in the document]
Maxez: Using special formatting for asm code.

Maxez:
I've added new commit with small changes (mostly formatting and changes proposed by @kemonats). I think the code looks good now.

Maxez (dismissed):

hal/ia32/cpu.c
```diff
@@ -135,62 +135,36 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
...
-
-		"jmp interrupts_popContext;"
-
-		"3:;"
-		"movl %%eax, %0"
-	: "=g" (err)
-	: "m" (spinlock), "m" (spinlock->lock), "r" (*scp), "g" (threads_schedule)
-	: "eax", "edx", "esp", "cc", "memory");
+	if (spinlock != NULL)
+		hal_spinlockClear(spinlock, scp);
```
agkaminski, line 139: spinlock should be cleared only when reschedule is ready - see example here https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/6a2f9bc0bf3eba9829f0a4f271003fa141c0bad3/hal/armv7m/stm32/_init.S#L226 and here https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/6a2f9bc0bf3eba9829f0a4f271003fa141c0bad3/hal/armv7a/_interrupts.S#L129

So at least interrupts should be disabled before clearing spinlock.

But from what I see this implementation wasn't doing that anyway - spinlock clear on beginning, just like on this PR, so no regress here

syscalls.c
```diff
@@ -802,8 +801,9 @@ int syscalls_signalPost(void *ustack)
 	err = threads_sigpost(proc, t, signal);
 
 	proc_put(proc);
-	threads_put(t);
-	hal_cpuReschedule(NULL, NULL);
+	if (t != NULL)
+		threads_put(t);
+
```
agkaminski, line 806: `if` I get, but why remove reschedule?

Maxez: Reschedule is already called inside `threads_sigpost()` - I think there is no need to call it again.

agkaminski: Ok, makes sense

hal/ia32/cpu.c
```diff
@@ -135,62 +135,37 @@ int hal_cpuReschedule(spinlock_t *spinlock, spinlock_ctx_t *scp)
...
- 		"pushl %%esp;"
-
-		"jmp interrupts_popContext;"
-
-		"3:;"
-		"movl %%eax, %0"
-	: "=g" (err)
-	: "m" (spinlock), "m" (spinlock->lock), "r" (*scp), "g" (threads_schedule)
-	: "eax", "edx", "esp", "cc", "memory");
+	hal_cpuDisableInterrupts();
```
agkaminski, line 138: This won't work - spinlockClear will enable interrupts by restoring interrupt mask. That's why we need to do spinlockClear manually in reschedule - so we'll only clear lock without restoring interrupt mask

Maxez: @agkaminski I see your point - `popf` in `hal_spinlockClear()` enables interrupts back, so I've restored manual spinlock clearing. Please take a look at the code again.

Sidenote: the original problem with `hal_cpuReschedule()` was passing "r" (*scp) in the input list. It caused page fault (NULL dereference) when calling `hal_cpuReschedule(NULL, NULL)`. The `*scp` dereference (red rectangle) got executed before `spinlock != NULL` check (`cmp $0x0, %eax` below the rectangle).
reschedule

hal/ia32/cpu.c
```diff
@@ -250,7 +245,9 @@ void *_cpu_initCore(void)
 	cpu.tss[hal_cpuGetID()].esp0 = (u32)&cpu.stacks[hal_cpuGetID()][511];
 
 	/* Set task register */
-	__asm__ volatile ("ltr %%ax" : : "a" ((4 + cpu.ncpus) * 8));
+	__asm__ volatile(
+		"ltr %%ax; "
+	:: "a" ((4 + cpu.ncpus) * 8));
```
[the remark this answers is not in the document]
Maxez: I agree. I've added the change.

agkaminski (approved):
LGTM

PR #253, opened 2021-11-17
syspage: new implementation

HBuczynski:
## Description

This PR is a part of changes made in [HBuczynski/syspage](https://github.com/phoenix-rtos/phoenix-rtos-kernel/tree/HBuczynski/syspage) branch. In order to provide better readability of review, the changes were separated into smaller pieces.

The following changes consist of:
 - making generic syspage for all platforms,
 - removing cpio archive,
 - sharing syspage headers between plo,
 - syspage structure on MMU platforms is copied to memory above `.bss`, so that `hal_syspageRelocate` is responsible for address relocation - it is temporary solution,

Syspage's code refers to hal functions which will be introduced in the next PR. 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt106x, imx6ull, zynq7000, qemu-system-aarch64).

## Checklist:

## Special treatment


Makefile
```diff
@@ -27,7 +27,7 @@ include ../phoenix-rtos-build/Makefile.common
 include ../phoenix-rtos-build/Makefile.$(TARGET_SUFF)
 
 CFLAGS += $(BOARD_CONFIG)
-CFLAGS += -I. -DHAL=\"hal/$(TARGET_SUFF)/hal.h\" -DVERSION=\"$(VERSION)\"
+CFLAGS += -I. -DHAL=\"hal/hal.h\" -DVERSION=\"$(VERSION)\"
```
kemonats, line 30: But there is an error:
```
<command-line>: fatal error: hal/hal.h: No such file or directory
```

Makefile
```diff
@@ -54,20 +52,10 @@ include lib/Makefile
...
-	@printf "EMBED programs.cpio\n"
-	$(SIL)$(OBJCOPY) --update-section .data=$(BUILD_DIR)/programs.cpio $(PREFIX_O)programs.o --add-symbol programs=.data:0 $(PREFIX_O)programs.o.cpio
-
-
-$(PREFIX_PROG)phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf: $(OBJS) $(PREFIX_O)/programs.o.cpio
+$(PREFIX_PROG)phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf: $(OBJS)
 	@mkdir -p $(@D)
 	@(printf "LD  %-24s\n" "$(@F)");
-	$(SIL)$(LD) $(LDFLAGS) -e _start --section-start .init=$(VADDR_KERNEL_INIT) -o $(PREFIX_PROG)phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf $(OBJS) $(PREFIX_O)/programs.o.cpio $(GCCLIB)
+	$(SIL)$(LD) $(LDFLAGS) -e _start --section-start .init=$(VADDR_KERNEL_INIT) -o $(PREFIX_PROG)phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf $(OBJS) $(GCCLIB)
```
kemonats, line 58: Shorter:
```suggestion
	$(SIL)$(LD) $(LDFLAGS) -e _start --section-start .init=$(VADDR_KERNEL_INIT) -o $@ $(OBJS) $(GCCLIB)
```
A litle of this PR maybe on occasion change `{}` to `()`.

Makefile
```diff
@@ -54,20 +52,10 @@ include lib/Makefile
...
-	$(SIL)touch $(BUILD_DIR)/programs.cpio
-
-
-$(PREFIX_O)/programs.o.cpio: $(PREFIX_O)programs.o $(BUILD_DIR)/programs.cpio
-	@printf "EMBED programs.cpio\n"
-	$(SIL)$(OBJCOPY) --update-section .data=$(BUILD_DIR)/programs.cpio $(PREFIX_O)programs.o --add-symbol programs=.data:0 $(PREFIX_O)programs.o.cpio
-
-
-$(PREFIX_PROG)phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf: $(OBJS) $(PREFIX_O)/programs.o.cpio
+$(PREFIX_PROG)phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.elf: $(OBJS)
```
kemonats, line 55:
```suggestion
$(PREFIX_PROG)phoenix-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf: $(OBJS)
```
replace `{}` with `()` also on line:
```
all: $(PREFIX_PROG_STRIPPED)phoenix-${TARGET_FAMILY}-${TARGET_SUBFAMILY}.el
```

Makefile
kemonats, line 58:
```suggestion
	$(SIL)$(LD) $(LDFLAGS) -e _start --section-start .init=$(VADDR_KERNEL_INIT) -o $@ $(OBJS) $(GCCLIB)
```

include/arch/syspage-imx6ull.h
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * %LICENSE%
+ */
+
+#ifndef _PHOENIX_SYSPAGE_IMX6ULL_H_
+#define _PHOENIX_SYSPAGE_IMX6ULL_H_
+
+
+typedef struct {
+
```
agkaminski, line 21: We need a dummy field here, empty struct is not allowed by standard C (it's allowed as GCC extension). So maybe `int dummy;`?

include/arch/syspage-imxrt.h
```diff
@@ -0,0 +1,33 @@
...
+typedef struct {
+	struct {
+		u32 type;
+		u32 allocCnt;
+		struct {
+			u32 rbar;
+			u32 rasr;
+		} table[16] __attribute__((aligned(8)));
+		u32 map[16]; /* ((u32)-1) = map is not assigned */
+	} mpu;
```
agkaminski, line 30: inner structs need to be packed as well

main.c
```diff
@@ -36,97 +36,77 @@ struct {
...
 
 	/* Set stdin, stdout, stderr ports */
-//	proc_fileAdd(&h, &oid, 0);
-//	proc_fileAdd(&h, &oid, 0);
-//	proc_fileAdd(&h, &oid, 0);
+	/*
+	proc_fileAdd(&h, &oid, 0);
+	proc_fileAdd(&h, &oid, 0);
+	proc_fileAdd(&h, &oid, 0);
+	*/
```
agkaminski, line 64: `//` was used here to indicate that the code has been disabled temporally.
It's disabled for quite some time, let's remove it for now (we have version control to store old code).

main.c
```diff
@@ -36,97 +36,77 @@ struct {
...
+				*cmdline = '\0';
+				continue;
+			}
 
-				argv[i++] = arg;
+			/* 'X' is no longer useful */
+			++prog->argv;
+			cmdline = prog->argv;
+			while (argc < (sizeof(argv) / sizeof(*argv) - 1)) {
+				argv[argc++] = cmdline;
```
agkaminski, line 86: `for` loop might fit better here

main.c
```diff
@@ -36,97 +36,77 @@ struct {
...
+	/*
+	proc_lookup("/dev/console", &oid);
 
-//	proc_fileSet(0, 3, &oid, 0, 0);
-//	proc_fileSet(1, 3, &oid, 0, 0);
-//	proc_fileSet(2, 3, &oid, 0, 0);
+	proc_fileSet(0, 3, &oid, 0, 0);
+	proc_fileSet(1, 3, &oid, 0, 0);
+	proc_fileSet(2, 3, &oid, 0, 0);
+	*/
```
agkaminski, line 109: As above, better remove it

syspage.c
```diff
@@ -0,0 +1,187 @@
...
+					return map;
+				break;
+
+			case syspage_nameResolve:
+				if (hal_strcmp((const char *)data, map->name) == 0)
+					return map;
+				break;
+
+			default:
+				return NULL;
```
agkaminski, line 71: I think that API would be easier to use and more efficient if there were 3 funtions for 3 types of resolve:
```
const syspage_map_t *syspage_mapIdResolve(unsigned int id);
const syspage_map_t *syspage_mapAddrResolve(addr_t addr);
const syspage_map_t *syspage_mapNameResolve(const char *name);
```

Right now this API is non-intuitive and increases number of MISRA violations.

syspage.c
```diff
@@ -0,0 +1,187 @@
...
+					return prog;
+				break;
+
+			case syspage_nameResolve:
+				if (hal_strcmp((const char *)data, prog->argv) == 0)
+					return prog;
+				break;
+
+			default:
+				return NULL;
```
agkaminski, line 122: As above, I suggest making two funtions:
```
const syspage_prog_t *syspage_progIdResolve(unsigned int id);
const syspage_prog_t *syspage_progNameResolve(const char *name);
```

Makefile
HBuczynski: As I wrote in a description, this PR is a part of the bigger changes in the kernel. To make review easier, the changes have been dividing into several parts, after reviewing all of them, the code will compile.
This PR will be merge to temporary branch `feature/syspage` so that it will not corrupt current master.

syspage.c
HBuczynski: Initially I wrote the same API as you proposed. 
However, taking into account main assumptions from writing API in `plo/hal.h`, that we should limit the count of functions I decided to concatenate 3 functions into one.
If from your perspective it is non-intuitive, I can change it.

main.c
HBuczynski: I have changed  `//` to `/* */` to be consistent with our coding standard. 
I was willing to remove it before, but I wasn't sure whether I would be still used somewhere.
 
What do you think about removing this part of code as well?
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/8538c0c28fd74847fa383e5b6f665660b0bcb882/main.c#L134-L141

agkaminski: `/* */` should be used for comments, that's true, but this is not a comment - it's a disabled code.
`#if 0` would be most right way to do it. But I think this code is dead for now, no need to keep it.
Don't remove tests for now, they are being used sometimes.

syspage.c
agkaminski: I think it is, I cannot use these function by just reading prototypes - I need to read it's body as well. Also, this approach introduces many casts (enum -> int -> enum, const void * -> unsigned int */ addr_t / const char *) that violate MISRA rules.

main.c
HBuczynski: done

main.c
HBuczynski: Done

include/arch/syspage-imxrt.h
HBuczynski: Done

include/arch/syspage-imx6ull.h
HBuczynski: Done for `imx6ull` and `zynq7000`

Makefile
HBuczynski: Done

Makefile
HBuczynski: Done

syspage.c
HBuczynski: Done

syspage.c
HBuczynski: Done

main.c
```diff
@@ -36,98 +36,54 @@ struct {
...
 
-	if (!xcount && syspage->progssz != 0) {
-		argv[1] = NULL;
-		/* Start all syspage programs */
-		for (i = 0; i < syspage->progssz; i++) {
-			argv[0] = syspage->progs[i].cmdline;
-			hal_memcpy(&prog, &syspage->progs[i], sizeof(prog));
-			res = proc_syspageSpawn(&prog, vm_getSharedMap(&prog, -1), prog.cmdline, argv);
-			if (res < 0) {
+			if ((res = proc_syspageSpawn((syspage_prog_t *)prog, vm_getSharedMap(prog, -1), argv[0], argv)) < 0)
```
kemonats, line 82:
```suggestion
			if ((res = proc_syspageSpawn(prog, vm_getSharedMap(prog, -1), argv[0], argv)) < 0)
```
`prog` is probably of the `syspage_prog_t*` type.

syscalls.c
```diff
@@ -316,23 +317,31 @@ void syscalls_meminfo(void *ustack)
...
-	prog->size = syspage->progs[i].end - syspage->progs[i].start;
-	sz = min(sizeof(prog->name), sizeof(syspage->progs[i].cmdline));
-	hal_memcpy(prog->name, syspage->progs[i].cmdline, sz);
-	prog->name[sz - 1] = '\0';
+	prog->addr = progSys->start;
+	prog->size = progSys->end - progSys->start;
+
+	/* TODO: change syspageprog_t to allocate data for name dynamically */
+	namesz = hal_strlen(progSys->argv);
+	sz = min(sizeof(prog->name), namesz);
```
kemonats, line 342:
```suggestion
	sz = min(sizeof(prog->name) - 1, namesz);
```
if `sz == sizeof(prog->name)` then `prog->name[sz] ` is out off buffer.

HBuczynski: Yeah, thanks for catching this one.

main.c
HBuczynski: Additional casting is not necessary, removed

main.c
```diff
@@ -36,98 +36,54 @@ struct {
...
+			}
 
-				argv[i++] = arg;
+			/* 'X' is no longer useful */
+			++prog->argv;
+			cmdline = prog->argv;
+			for (argc = 0; argc < (sizeof(argv) / sizeof(*argv) - 1);) {
+				argv[argc++] = cmdline;
+				while (*cmdline != ';' && *cmdline != '\0')
+					++cmdline;
```
agkaminski, line 73:
```suggestion
			for (argc = 0; (argc < (sizeof(argv) / sizeof(*argv) - 1)) && (*cmdline != '\0'); ++argc) {
				argv[argc] = cmdline;
				while (*cmdline != ';' && *cmdline != '\0')
					++cmdline;
```
Remove `if (...) break;` below and `argc++` in line below for, should work fine, I think.

I mention this because of MISRA's `for loop should be well formed` rule.

HBuczynski: https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/24c96fa6795abe2aa12f7ecb43d4a2b89bc9b242/main.c#L72-L73
Above condition will not be true if there is a `;` or `\0`.

If `cmdline` points to `;`, it will change value to `\0` and starts pointing to the next byte. It is ok.
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/24c96fa6795abe2aa12f7ecb43d4a2b89bc9b242/main.c#L75-L78

However, if `cmdline` points to `\0`, it should exit and not perform `line 78`. 
In Your proposition, `cmdline` will point to data outside a string what will cause that `for` loop will execute again, what is wrong.

agkaminski: Ah, yes, you're right, not that simple. Ok, so let's drop this for now

agkaminski (approved):

PR #258, opened 2021-11-24
hal: common interface for all platforms

HBuczynski:
## Description

This PR is a part of changes made in [HBuczynski/syspage](https://github.com/phoenix-rtos/phoenix-rtos-kernel/commits/HBuczynski/syspage) branch. In order to provide better readability of review, the changes were separated into smaller pieces.

The following changes consist of:
 - adding common interface in hal directory,
 - adopting new interface for armv7a & armv7m,
 - changing timer interface

## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: (imx6ull, zynq7000, imxrt106x, qemu-system-aarch64).

## Checklist:

## Special treatment


hal/armv7a/zynq7000/timer.c
```diff
@@ -72,13 +60,38 @@ static int _timer_irqHandler(unsigned int n, cpu_context_t *ctx, void *arg)
...
+}
+
+
+time_t hal_timerCyc2Us(time_t cyc)
+{
+	return (cyc * 1000LL) / 55555LL;
+}
+
+
+time_t hal_timerGetUs(void)
```
agkaminski, line 81: Why is this exact copy of `hal_timerGetCyc`? Looking at functions above `cyc != us` on this arch.

hal/armv7m/cpu.c
```diff
@@ -13,39 +13,140 @@
...
+
+unsigned int hal_cpuGetFirstBit(const u32 v)
+{
+	unsigned pos;
+
+	__asm__ volatile ("\
+		rbit %0, %1; \
+		clz  %0, %0;" : "=r" (pos) : "r" (v));
+
+	return (pos);
```
agkaminski, line 123:
```suggestion
	return pos;
```

hal/cpu.h
```diff
@@ -0,0 +1,130 @@
...
+
+extern void hal_cpuEnableInterrupts(void);
+
+
+/* performance */
+
+
+extern void hal_cpuLowPower(time_t us);
+
+/* TODO: it is only related to stm32 - IFDEF in syscall!! To remove! */
```
agkaminski, line 37: No - for now it is, but we might want to add low power modes on others archs

hal/armv7m/cpu.c
HBuczynski: Done

hal/cpu.h
HBuczynski: In such case, I can assume that this function will be available in all platforms and I can remove `ifdef` in syscall:
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/5e3e16d27303d392620b1203e26166d7718b596d/syscalls.c#L653-L658

hal/armv7a/zynq7000/timer.c
HBuczynski: Functions are not the same.

agkaminski: Right, not exact. But still, this function can mostly replaced by call to `hal_timerGetCyc`

kemonats:
General code question.
What caused functions to change from `inline` type to regular type, even those containinhg only one instruction?

hal/armv7a/cpu.c
```diff
@@ -87,21 +152,105 @@ int hal_cpuCreateContext(cpu_context_t **nctx, void *start, void *kstack, size_t
...
-	n += 10;
-#elif CPU_IMX6UL
-	hal_strcpy(&info[n], "i.MX 6UL ");
-	n += 9;
-#else
-	hal_strcpy(&info[n], "unknown ");
-	n += 8;
-#endif
+	hal_strcpy(&info[n], HAL_NAME_PLATFORM);
+	n += hal_strlen(HAL_NAME_PLATFORM);
```
kemonats, line 253:
```suggestion
	hal_strcpy(info, HAL_NAME_PLATFORM);
	n = sizeof(HAL_NAME_PLATFORM) - 1;
```
The length of HAL_NAME_PLATFORM is known at compile time.
Maybe a better solution would be to use a pointer instead of an array in this function.

HBuczynski:
> General code question. What caused functions to change from `inline` type to regular type, even those containinhg only one instruction?

The main cause of those changes is to provide consistent interface used by all targets. Moving function declarations into common headers eliminated options to use `inline` functions with assembly code. However, in many situations `inline` functions were abused. After presented changes, the kernel size decreased by about 6 KB.

hal/armv7a/cpu.c
HBuczynski: Good suggestion, done.

hal/armv7a/zynq7000/timer.c
HBuczynski: Done

agkaminski (approved):

kemonats:
How does this relate to @agkaminski statement [look](https://github.com/phoenix-rtos/libphoenix/issues/48
)
@kemonats:
What is the purpose of using str * and mem * functions as inline in kernel?
@agkaminski:
**We prefer performance over binary size in the kernel.**

Now what will be preferred performance or size?
Will functions written using inline assembler remain ( e.g. str* and mem*)?

agkaminski:
@kemonats This is always a compromise. We decided in the end that str_* and mem_* function being inline is too much. But we moved most content of cpu.c back to inline, because as you pointed out, these function mostly have single instruction - not inlining them would hurt both size and performance.

PR #260, opened 2021-11-26
hal: update cpu info & features

HBuczynski:
## Description

 - fix console
 - update cpu info & features for Cortex-M
 
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/cpu.c
+++ b/hal/armv7m/cpu.c
@@ -165,29 +165,37 @@ char *hal_cpuInfo(char *info)
 
 #ifdef CPU_STM32
 	cpuinfo = _stm32_cpuid();
-	hal_strcpy(info, "STM32 ");
-	i = 6;
 #elif defined(CPU_IMXRT)
 	cpuinfo = _imxrt_cpuid();
-	hal_strcpy(info, "i.MX RT ");
-	i = 8;
 #else
 	hal_strcpy(info, "unknown");
 	return info;
 #endif
-	if (((cpuinfo >> 24) & 0xff) == 0x41) {
-		hal_strcpy(info + i, "ARM ");
-		i += 4;
-	}
 
-	*(info + i++) = 'r';
-	*(info + i++) = '0' + ((cpuinfo >> 20) & 0xf);
-	*(info + i++) = ' ';
+	hal_strcpy(info, HAL_NAME_PLATFORM);
+	i = sizeof(HAL_NAME_PLATFORM) - 1;
+
+	if (((cpuinfo >> 24) & 0xff) == 0x41) {
+		hal_strcpy(info + i, "ARMv7 ");
+		i += 6;
+	}
 
 	if (((cpuinfo >> 4) & 0xfff) == 0xc23) {
 		hal_strcpy(info + i, "Cortex-M3 ");
 		i += 10;
 	}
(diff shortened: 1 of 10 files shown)


hal/armv7m/cpu.c
```diff
@@ -165,29 +165,33 @@ char *hal_cpuInfo(char *info)
...
 
-	*(info + i++) = 'r';
-	*(info + i++) = '0' + ((cpuinfo >> 20) & 0xf);
-	*(info + i++) = ' ';
-
 	if (((cpuinfo >> 4) & 0xfff) == 0xc23) {
 		hal_strcpy(info + i, "Cortex-M3 ");
 		i += 10;
 	}
+	else if (((cpuinfo >> 4) & 0xfff) == 0xc27) {
```
agkaminski, line 187:
```suggestion
	else if (((cpuinfo >> 4) & 0xfff) == 0xc24) {
		hal_strcpy(info + i, "Cortex-M4 ");
		i += 10;
	}
	else if (((cpuinfo >> 4) & 0xfff) == 0xc27) {
```

hal/armv7m/cpu.c
```diff
@@ -199,7 +203,39 @@ char *hal_cpuInfo(char *info)
...
+	if ((len - n) > 5) {
+		hal_strcpy(features + n, "FPU, ");
+		n += 5;
+	}
+
+	if ((len - n) > 9) {
+		hal_strcpy(features + n, "MPU 16R, ");
+		n += 9;
+	}
+#elif defined(CPU_IMXRT)
```
agkaminski, line 217:
```suggestion
#elif defined(CPU_STM32)
```

hal/armv7m/cpu.c
```diff
@@ -199,7 +203,39 @@ char *hal_cpuInfo(char *info)
...
-	features[0] = '\0';
+	unsigned int n = 0;
+#ifdef CPU_IMXRT
+	if ((len - n) > 5) {
+		hal_strcpy(features + n, "FPU, ");
+		n += 5;
+	}
+
+	if ((len - n) > 9) {
+		hal_strcpy(features + n, "MPU 16R, ");
```
agkaminski, line 214: Number of MPU regions should be extracted from MPU_RNR register (https://developer.arm.com/documentation/ddi0439/b/Memory-Protection-Unit/MPU-programmers-model)

gerard5: FYI, In the current plo (not feature/syspage), at this stage the plo has already detected the `mpu` regions, information about which is passed through syspage to the kernel, so it can be reused, also additional information about pre-allocated regions can be printed there (allocated/max).

HBuczynski: Currently, the best approach is to omit the the number of `MPU` regions and add it in the future when `MPU` interface will be implemented in the kernel. I don't want to add one function only to get MPU region.

hal/armv7m/cpu.c
HBuczynski: Done

hal/armv7m/cpu.c
HBuczynski: Done

agkaminski (approved):

PR #261, opened 2021-11-29
all files: replace '#include HAL' with path to a file

HBuczynski:
## Description

`define HAL` is removed.

## Motivation and Context

The headers need some cleaning to not include the whole `hal` but only specific unit's header. 
However, @Maxez is currently working on changes in including headers, so I don't want to interfere in his work and I've done only minimal changes.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/lib/bsearch.c
+++ b/lib/bsearch.c
@@ -14,7 +14,7 @@
  *
  */
 
-#include HAL
+#include "../hal/hal.h"
 #include "bsearch.h"
 
 
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ include ../phoenix-rtos-build/Makefile.common
 include ../phoenix-rtos-build/Makefile.$(TARGET_SUFF)
 
 CFLAGS += $(BOARD_CONFIG)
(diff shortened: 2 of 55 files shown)


lib/bsearch.c
```diff
@@ -14,7 +14,7 @@
  *
  */
 
-#include HAL
+#include "../hal/hal.h"
```
nalajcie, line 17: please note that You have `-I.` in `CFLAGS` to avoid the need to use `..` across includes (which for some reason is ignored in many places :)). 

Side note: I think that `-I.` is also unneeded as CWD is always added to local include path - otherwise `include "file.h" would not work (alas might have change how `include <file.h>` works).

nalajcie:
HA, finally! :clap:

lib/bsearch.c
HBuczynski: I agree with You that this part also should be cleaned up, but like I wrote in a description that there are coming changes in headers in `phoenix-rtos-kernel` so I have decided to wait on it before making a bigger changes.

agkaminski (approved):

PR #266, opened 2021-12-06
armv7a/imx6ull: remove persistent bits clearing in init

nalajcie:
## Description
Remove persistent bit clearing at bootup to be read out later in userspace. Using psh `reboot` or disconnecting power will reset this bit anyway.

## Motivation and Context
JIRA: DTR-181

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `imx6ull`.

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/imx6ull/_init.S
+++ b/hal/armv7a/imx6ull/_init.S
@@ -348,12 +348,6 @@ _start:
 
 	/* WARNING: R7 - can't touch this until MMU enable */
 
-	/* Clear PERSIST_SECONDARY_BOOT bit */
-	ldr r0, =0x020d8044
-	ldr r1, [r0]
-	bic r1, r1, #(1 << 30)
-	str r1, [r0]
-
 	/* Enable PMU cycle counter */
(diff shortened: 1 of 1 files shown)


agkaminski:
Reboot other then POR should leave this bit as is (so if it's set psh reboot will cause secondary image boot)

nalajcie:
> Reboot other than POR should leave this bit as is (so if it's set psh reboot will cause secondary image boot)

Nope, see:
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/0b08b023132ecded98df9cc999b72a800138003c/hal/armv7a/imx6ull/imx6ull.c#L380-L393

If there would be watchdog reboot, I would like to resume secondary reboot if possible (primary partitions might be in unusable state (eg. new rootfs + old kernel)  if FW upgrade is designed the wrong way), so IMHO this was always invalid.

agkaminski (approved):
Ok, a point about reset caused by the watchdog is a good one.

PR #268, opened 2021-12-13
klog: Add kernel log buffer

MaciejPurski:
JIRA: DTR-3

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/klog/klog.c
+++ b/klog/klog.c
@@ -104,7 +104,8 @@ static klog_reader_t *klog_readerFind(unsigned pid)
 {
 	klog_reader_t *r, *ret = NULL;
 
-	if ((r = klog_common.readers) != NULL) {
+	r = klog_common.readers;
+	if (r != NULL) {
 		do {
 			if (r->pid == pid) {
 				ret = r;
@@ -125,7 +126,8 @@ static int klog_readerAdd(unsigned pid, unsigned nonblocking)
 	if (klog_readerFind(pid) != NULL)
 		return -EINVAL;
 
-	if ((r = vm_kmalloc(sizeof(klog_reader_t))) == NULL)
+	r = vm_kmalloc(sizeof(klog_reader_t));
+	if (r == NULL)
 		return -ENOMEM;
 
 	hal_memset(r, 0, sizeof(klog_reader_t));
@@ -234,7 +236,8 @@ static void klog_close(unsigned pid)
 	klog_reader_t *r;
 
 	proc_lockSet(&klog_common.lock);
-	if ((r = klog_readerFind(pid)) != NULL) {
+	r = klog_readerFind(pid);
+	if (r != NULL) {
 		while (r->msgs != NULL)
 			_klog_msgRespond(r, -EIO);
 		LIST_REMOVE(&klog_common.readers, r);
@@ -301,7 +304,8 @@ static int klog_readerBlock(klog_reader_t *r, msg_t *msg, unsigned long rid)
 {
 	klog_readMsg_t *rmsg;
 
-	if ((rmsg = vm_kmalloc(sizeof(*rmsg))) == NULL)
+	rmsg = vm_kmalloc(sizeof(*rmsg));
+	if (rmsg == NULL)
 		return -ENOMEM;
 
 	rmsg->odata = msg->o.data;
@@ -336,13 +340,15 @@ static void msgthr(void *arg)
 					msg.o.io.err = klog_readerAdd(msg.pid, msg.i.openclose.flags & O_NONBLOCK);
 				break;
 			case mtRead:
-				if ((r = klog_readerFind(msg.pid)) == NULL) {
+				r = klog_readerFind(msg.pid);
+				if (r == NULL) {
 					msg.o.io.err = -EINVAL;
 				}
 				else {
 					msg.o.io.err = klog_read(r, msg.o.data, msg.o.size);
 					if (msg.o.io.err == 0 && !r->nonblocking) {
-						if ((msg.o.io.err = klog_readerBlock(r, &msg, rid)) == EOK)
+						msg.o.io.err = klog_readerBlock(r, &msg, rid);
+						if (msg.o.io.err == EOK)
 							respond = 0;
 					}
 					else if (msg.o.io.err == 0 && r->nonblocking) {
--- a/klog/klog.c
+++ b/klog/klog.c
@@ -135,9 +135,9 @@ static int klog_readerAdd(unsigned pid, unsigned nonblocking)
 }
 
 
-static int _klog_readln(klog_reader_t *r, char *buf, size_t sz)
+static ssize_t _klog_readln(klog_reader_t *r, char *buf, size_t sz)
 {
-	size_t n = 0;
+	ssize_t n = 0;
 
 	while (r->ridx < klog_common.tail && n < sz) {
 		buf[n++] = _fifo_get(r->ridx++);
@@ -147,19 +147,25 @@ static int _klog_readln(klog_reader_t *r, char *buf, size_t sz)
 
 	/* Always end with newline */
 	if (n > 0 && buf[n - 1] != '\n') {
-		if (n < sz)
-			buf[n++] = '\n';
-		else
+		if (buf[n - 1] == '\0') {
 			buf[n - 1] = '\n';
+		}
+		else if (n < sz) {
+			buf[n++] = '\n';
+		}
+		else {
+			buf[n - 1] = '\n';
+			r->ridx--;
+		}
 	}
 
 	return n;
 }
 
 
-static int klog_read(klog_reader_t *r, char *buf, size_t sz)
+static ssize_t klog_read(klog_reader_t *r, char *buf, size_t sz)
 {
-	int ret;
+	ssize_t ret;
 
 	proc_lockSet(&klog_common.lock);
 	/* We need to catch up the ring buffer's head */
--- a/klog/klog.c
+++ b/klog/klog.c
@@ -1,0 +1,327 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * Kernel log buffer
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Maciej Purski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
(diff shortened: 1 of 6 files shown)


klog/klog.c
```diff
@@ -0,0 +1,327 @@
...
+
+	if ((r = vm_kmalloc(sizeof(klog_reader_t))) == NULL)
+		return -ENOMEM;
+
+	r->next = NULL;
+	r->prev = NULL;
+	r->nonblocking = !!nonblocking;
+	r->pid = pid;
+	r->waiting = 0;
+	hal_memset(&r->msg, 0, sizeof(msg_t));
```
agkaminski, line 128: We don't really need to keep the whole msg reader sends us - we only need few params from it (buffer and buffsz I think)

klog/klog.c
```diff
@@ -0,0 +1,327 @@
...
+void _klog_init(void)
+{
+	hal_memset(&klog_common, 0, sizeof(klog_common));
+	proc_lockInit(&klog_common.lock);
+}
+
+
+static klog_reader_t *klog_readerFind(unsigned pid)
+{
+	klog_reader_t *r = klog_common.readers;
```
agkaminski, line 99: How about - use tree instead of list. This way we can use `pid` as our index.

klog/klog.c
```diff
@@ -0,0 +1,327 @@
...
+static int _klog_readln(klog_reader_t *r, char *buf, size_t sz)
+{
+	size_t n = 0;
+	char c;
+
+	while (r->ridx < klog_common.tail && n < sz) {
+		c = _fifo_get(r->ridx++);
+		buf[n++] = c;
+		if (c == '\n' || c == '\0')
+			break;
```
agkaminski, line 148: We don't really need c
```suggestion
		buf[n++] = _fifo_get(r->ridx++);
		if (buf[n] == '\n' || buf[n] == '\0')
			break;
```

klog/klog.c
```diff
@@ -0,0 +1,327 @@
...
+	}
+
```
agkaminski, line 150: Maybe we should end each line with '\n' and/or '\0'? Now it's pretty random

klog/klog.c
```diff
@@ -0,0 +1,327 @@
...
+
+static void klog_close(unsigned pid)
+{
+	klog_reader_t *r;
+
+	proc_lockSet(&klog_common.lock);
+	if ((r = klog_readerFind(pid)) != NULL) {
+		if (r->waiting) {
+			r->msg.o.io.err = -EINVAL;
```
agkaminski, line 196: EINVAL doesn't really fit this case. EIO maybe?

klog/klog.c
```diff
@@ -0,0 +1,327 @@
...
+	 * only checks if a device responds to TCGETS
+	 */
+	if (in->request == TCGETS) {
+		out->err = EOK;
+	}
+	else {
+		out->err = -EINVAL;
+	}
+
+	out->err = 0;
```
agkaminski, line 224: Ooops?

klog/klog.c
```diff
@@ -0,0 +1,327 @@
...
+}
+
+
+void _klog_initSrv(void)
+{
+	/* Create port 0 for /dev/kmsg */
+	if (proc_portCreate(&klog_common.port) != 0)
+		return;
+
+	proc_threadCreate(NULL, msgthr, NULL, 4, 2048, NULL, 0, NULL);
```
agkaminski, line 326: 1024 bytes of stack should be enough.

klog/klog.h
```diff
@@ -0,0 +1,28 @@
...
+#define _DMESG_H_
+
+
+extern void _klog_init(void);
+
+
+extern void _klog_initSrv(void);
+
+
+extern int klog_write(const char *data, size_t len);
```
agkaminski, line 25: Move this on top, inits should last

klog/klog.c
MaciejPurski: Yes, I was considering using a tree, but I decided that it would be a bit of an overkill, as for most of the time we would have a very limited number of readers. I can change it to an rbtree if you prefer.

klog/klog.c
MaciejPurski: I agree, but the code you have suggested is not an equivalent of what I have written. There should be `buf[n - 1]`.

agkaminski: Right, sorry

klog/klog.c
MaciejPurski: Ok, we can always finish each read line with `\n`.

klog/klog.c
agkaminski: I suggest a tree mostly because I think it might simplify logic here, your call

klog/klog.c
MaciejPurski: Indeed.

klog/klog.c
MaciejPurski: I agree.

klog/klog.h
MaciejPurski: Ok.

klog/klog.c
MaciejPurski: buffer, bufsz, pid, and rid. Type is always mtRead. I agree we can get rid of it.

klog/klog.c
```diff
@@ -0,0 +1,348 @@
...
+	r->ridx = klog_common.head;
+	LIST_ADD(&klog_common.readers, r);
+	proc_lockClear(&klog_common.lock);
+
+	return 0;
+}
+
+
+static int _klog_readln(klog_reader_t *r, char *buf, size_t sz)
+{
```
kemonats, line 139: Why `int` and not `size_t`?

klog/klog.c
```diff
@@ -0,0 +1,348 @@
...
+			break;
+	}
+
+	/* Always end with newline */
+	if (n > 0 && buf[n - 1] != '\n') {
+		if (n < sz)
+			buf[n++] = '\n';
+		else
+			buf[n - 1] = '\n';
+	}
```
kemonats, line 154: 1. Input: `line = "abcd\n" sz = 2` 
`n == 2` and `buf[1] == 'b'`
You overwrite char in `buf[n -1]` with a newline char.
Output:
`buf == "a\n" n == 2`

2. Input: `line="abc\0" sz = 10`
`n == 4 buf[3] == '\0'`
You append '\n' to the oryginal line
Output:
`buf="abc\0\n" n == 5`

MaciejPurski: The first case is desired and that is what I meant. The idea is to *always* end one read with a newline character. So we read as many characters as we can, but always finish with `\n`. However now this overwritten character would be completely lost, so it would be better to update a reader's read index in order to read this character with a next read.

The second case is arguable, but I think it would be better to overwrite this `\0` character.

klog/klog.c
MaciejPurski: The return value is assigned to `msg.o.io.err` anyway, which is an `int` so it makes no difference here. The conversion would occur anyway.

klog/klog.c
```diff
@@ -0,0 +1,354 @@
...
+
+
+static ssize_t klog_read(klog_reader_t *r, char *buf, size_t sz)
+{
+	ssize_t ret;
+
+	proc_lockSet(&klog_common.lock);
+	/* We need to catch up the ring buffer's head */
+	if (r->ridx < klog_common.head)
+		r->ridx = klog_common.head;
```
agkaminski, line 173: We need to generate -EPIPE error in that case

klog/klog.c
```diff
@@ -0,0 +1,354 @@
...
+	proc_lockClear(&klog_common.lock);
+#else
+	hal_consolePrint(ATTR_NORMAL, data);
+#endif /* KLOG_ENABLE */
+
+	return len;
+}
+
+
+static void klog_readerBlock(klog_reader_t *r, msg_t *msg, unsigned long rid)
```
agkaminski, line 283: Edge case is not handled - what if we do two blocking reads on one fd?
We need a list of read msg within a list of readers, I'm afraid

klog/klog.c
```diff
@@ -169,55 +174,65 @@ static ssize_t klog_read(klog_reader_t *r, char *buf, size_t sz)
...
+	klog_readMsg_t *rmsg;
 	msg_t msg;
 
+	rmsg = r->msgs;
+	LIST_REMOVE(&r->msgs, rmsg);
+
 	msg.type = mtRead;
+	msg.pid = r->pid;
+	msg.o.data = rmsg->odata;
+	msg.o.size = rmsg->osize;
```
agkaminski, line 201: Missing msg.i initialization:
```
msg.i.data = NULL;
msg.i.size = 0;
```

klog/klog.c
```diff
@@ -169,55 +174,65 @@ static ssize_t klog_read(klog_reader_t *r, char *buf, size_t sz)
...
+	if (r == NULL)
+		return;
+
+	do {
+		if (r->msgs != NULL) {
+			ret = _klog_readln(r, r->msgs->odata, r->msgs->osize);
+			_klog_msgRespond(r, ret);
+		}
+		r = r->next;
+	} while (r != klog_common.readers);
```
agkaminski, line 224: MISRA:
```suggestion
	if (r != NULL) {
		do {
			if (r->msgs != NULL) {
				ret = _klog_readln(r, r->msgs->odata, r->msgs->osize);
				_klog_msgRespond(r, ret);
			}
			r = r->next;
		} while (r != klog_common.readers);
	}
```

klog/klog.c
```diff
@@ -0,0 +1,387 @@
...
+	int i = 0, overwrite = 0;
+	char c;
+
+	proc_lockSet(&klog_common.lock);
+
+	if (klog_common.tail + len >= klog_common.head + KLOG_BUFSZ)
+		overwrite = 1;
+
+	while (i < len)
+		_fifo_push(data[i++]);
```
agkaminski, line 284: Nitpick - for fits better
```suggestion
	for (i = 0; i < len; ++i)
		_fifo_push(data[i]);
```

klog/klog.c
agkaminski: So let's change it :D

agkaminski (approved):

PR #272, opened 2021-12-16
hal/ia32: apply new syspage

HBuczynski:
## Description

 - fix initialization of `_interrupts_multilock` variable
 - apply new syspage approach
 - update console

**NOTE:** console-vga doesn't work correctly. `lib_printf()` sets console color using ASCI what is incomprehensible to vga.
Changes will be introduced in a separate PR.
Furthermore, **console-serial** has hard-coded instance of base address. This code will be changed in the next PR as well.
 
## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: (lia32-generic).

## Checklist:

## Special treatment

- [ ] This PR needs additional PRs to work (https://github.com/phoenix-rtos/plo/pull/157).

### DIFF
--- a/hal/ia32/cpu.c
+++ b/hal/ia32/cpu.c
@@ -22,6 +22,7 @@
 #include "../pmap.h"
 #include "pci.h"
 #include "ia32.h"
+#include "halsyspage.h"
 
 
 struct cpu_feature_t {
@@ -86,6 +87,7 @@ u32 cpu_getEFLAGS(void)
 
 	return eflags;
 }
+
 
 #ifndef NDEBUG
 static int hal_cpuDebugGuard(u32 enable, u32 slot)
@@ -384,8 +386,6 @@ void cpu_sendIPI(unsigned int cpu, unsigned int intr)
 
 static void _cpu_gdtInsert(unsigned int idx, u32 base, u32 limit, u32 type)
 {
-/* TODO: add new syspage */
-#if 0
 	u32 descrl, descrh;
 	u32 *gdt;
 
@@ -398,13 +398,12 @@ static void _cpu_gdtInsert(unsigned int idx, u32 base, u32 limit, u32 type)
 
 	descrl = (base << 16) | (limit & 0xffff);
 
-	gdt = (void *)*(u32 *)&syspage->gdtr[2];
+	gdt = (void *)syspage->hs.gdtr.addr;
 
 	gdt[idx * 2] = descrl;
 	gdt[idx * 2 + 1] = descrh;
 
 	return;
-#endif
 }
 
 
@@ -433,8 +432,6 @@ void *_cpu_initCore(void)
 
 static void _hal_cpuInitCores(void)
 {
-/* TODO: add new syspage */
-#if 0
 	unsigned int i, k;
 
 	/* Prepare descriptors for user segments */
@@ -445,7 +442,7 @@ static void _hal_cpuInitCores(void)
 	cpu.ncpus = 0;
 	_cpu_initCore();
 
-	*(u32 *)((void *)syspage->stack + VADDR_KERNEL - 4) = 0;
+	*(u32 *)(syspage->hs.stack + VADDR_KERNEL - 4) = 0;
 
 	for (;;) {
 		k = cpu.ncpus;
@@ -455,7 +452,6 @@ static void _hal_cpuInitCores(void)
 		if (i >= 50000000)
 			break;
 	}
-#endif
 }
 
 
--- a/hal/ia32/halsyspage.h
+++ b/hal/ia32/halsyspage.h
@@ -3,7 +3,7 @@
  *
  * Operating system kernel
  *
- * Configuration file for ia2
+ * System information page (prepared by kernel loader)
  *
  * Copyright 2021 Phoenix Systems
  * Author: Hubert Buczynski
@@ -13,15 +13,12 @@
  * %LICENSE%
  */
 
-#ifndef _HAL_CONFIG_H_
-#define _HAL_CONFIG_H_
+#ifndef _HAL_SYSPAGE_H_
+#define _HAL_SYSPAGE_H_
 
-#ifndef __ASSEMBLY__
+#include <arch/types.h>
+#include "config.h"
 
-#include "../../include/arch/syspage-ia32.h"
-#include "../../include/syspage.h"
-
-#endif
-
+extern syspage_t *syspage;
 
 #endif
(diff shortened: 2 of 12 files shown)


hal/ia32/halsyspage.h
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * %LICENSE%
+ */
+
+#ifndef _HAL_SYSPAGE_H_
+#define _HAL_SYSPAGE_H_
+
+#include "config.h"
+
+extern hal_syspage_t *syspage;
```
agkaminski, line 21: Hmm, we already have this symbol, don't we? So this is it's redefinition?

hal/ia32/pmap.c
```diff
@@ -235,40 +226,48 @@ int pmap_getPage(page_t *page, addr_t *addr)
...
-	if (k == syspage->mmsize)
-		return -EINVAL;
-
-	/* Page is allocated by BIOS */
-	if (!(page->flags & PAGE_FREE)) {
-		page->flags |= PAGE_OWNER_BOOT;
-		return EOK;
-	}
-
+#if 0
```
agkaminski, line 270: You added #if 0, so maybe remove comment that comments-out the code inside?

include/arch/syspage-imxrt.h
```diff
@@ -20,13 +20,13 @@
...
+		unsigned int type;
+		unsigned int allocCnt;
 		struct {
-			u32 rbar;
-			u32 rasr;
+			unsigned int rbar;
+			unsigned int rasr;
 		} table[16] __attribute__((aligned(8)));
-		u32 map[16]; /* ((u32)-1) = map is not assigned */
+		unsigned int map[16]; /* ((unsigned int)-1) = map is not assigned */
```
agkaminski, line 29: Unrelated change and I'm not sure if we want to change it that way

hal/ia32/halsyspage.h
```diff
@@ -0,0 +1,23 @@
+/*
```
agkaminski, line 1: Do we really need this file?

hal/ia32/pmap.c
```diff
@@ -278,23 +277,23 @@ int pmap_getPage(page_t *page, addr_t *addr)
 	}*/
 
 	hal_spinlockClear(&pmap_common.lock, &sc);
+#endif
 
-	if ((page->addr >= (u32)syspage - VADDR_KERNEL) && (page->addr < (u32)syspage - VADDR_KERNEL + SIZE_PAGE)) {
+	/* Calculate syspage size */
+	sz = *(ptr_t *)((ptr_t)syspage + sizeof(hal_syspage_t));
```
agkaminski, line 283: Why not `syspage->size`?

hal/ia32/pmap.c
HBuczynski: What do you think about removing it permanently? 
This piece of code describes `ptable` which has a new approach in `pmap_getPage` function:

https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3b85faa3db0cb156f678c30dc1d430cc9f0a8f00/hal/ia32/pmap.c#L295-L298

include/arch/syspage-imxrt.h
HBuczynski: You have right that it is not related to the context of this PR but I have added appropriate comment in commit message. 
`u32` and `u8` are described in `hal`. This code is shared between plo and kernel so it will be better to not use types which are defined in both modules separately.

hal/ia32/pmap.c
HBuczynski: `syspage` is a type of `hal_syspage_t *`. `size` is a part of generic syspage structure `syspage_t.`

agkaminski: It would be better to handle main `syspage` symbol here. You include syspage.h anyway, so why not to access generic syspage by `syspage->field` and `hal_syspage_t` by `syspage->hs->field`?

hal/ia32/halsyspage.h
HBuczynski: The answer is above.

hal/ia32/pmap.c
HBuczynski: The main assumption was that hal has an access only to `hal_syspage_t` part of the syspage structure. 
`pmap_getPage` uses fields from generic part of syspage but it can be moved outside of hal, on each of the platform with MMU code looks similar.

Your proposition will solve the problem at the moment but I think that we should follow the rule that hal gets from the outside world only the minimum.

hal/ia32/cpu.c
```diff
@@ -445,7 +442,7 @@ static void _hal_cpuInitCores(void)
 	cpu.ncpus = 0;
 	_cpu_initCore();
 
-	*(u32 *)((void *)syspage->stack + VADDR_KERNEL - 4) = 0;
+	*(u32 *)((void *)syspage->hs.stack + VADDR_KERNEL - 4) = 0;
```
agkaminski, line 445: void * arithmetic

hal/ia32/pmap.c
```diff
@@ -417,38 +398,39 @@ int pmap_segment(unsigned int i, void **vaddr, size_t *size, int *prot, void **t
...
+		if (map->start < pmap_common.minAddr)
+			pmap_common.minAddr = map->start;
+
+		if (map->end > pmap_common.maxAddr)
+			pmap_common.maxAddr = map->end;
+	} while ((map = map->next) != syspage->maps);
 
 	/* Initialize kernel page table - remove first 4 MB mapping */
-	pmap->pdir = VADDR_KERNEL + (void *)syspage->pdir;
+	pmap->pdir = VADDR_KERNEL + (void *)syspage->hs.pdir;
```
agkaminski, line 423: void * arithmetic

include/arch/syspage-imxrt.h
agkaminski: Ok for now, but imho this types might be/should be shared between plo and kernel in the future

hal/ia32/pmap.c
HBuczynski: Fixed

agkaminski (approved):
LGTM, let's wait for @Maxez

Maxez (approved):
One thing to change - we've agreed that `syspage` variable points to the whole syspage (not just the hal part) so we should change its definition in hal.c (hal_syspage_t * needs to be changed to syspage_t *).

HBuczynski:
> One thing to change - we've agreed that `syspage` variable should point to the whole syspage (not just the hal part) so we should change its definition in hal.c (`hal_syspage_t *` needs to be changed to `syspage_t *`).

Thanks, done.

PR #277, opened 2021-12-17
hal/ia32/console-vga: change colors using ANSI escape sequences

HBuczynski:
## Description

RTOS-140

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/ia32/console.c
+++ b/hal/ia32/console.c
@@ -3,11 +3,11 @@
  *
  * Operating system kernel
  *
- * HAL console
+ * VGA console
  *
- * Copyright 2012, 2016, 2020 Phoenix Systems
+ * Copyright 2012, 2016, 2020, 2021 Phoenix Systems
  * Copyright 2001, 2005-2008 Pawel Pisarczyk
- * Author: Pawel Pisarczyk
+ * Author: Pawel Pisarczyk, Lukasz Kosinski
  *
  * This file is part of Phoenix-RTOS.
  *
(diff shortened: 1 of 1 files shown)


agkaminski (approved):
LGTM, let's wait for @Maxez approve

hal/ia32/console.c
```diff
@@ -21,133 +21,343 @@
...
+	/* Initialize VGA */
+	halconsole_common.vram = (u16 *)(VADDR_KERNEL + (color ? 0xb8000 : 0xb0000));
+	halconsole_common.crtc = (void *)(color ? 0x3d4 : 0x3b4);
 
-	hal_spinlockCreate(&console.spinlock, "console.spinlock");
-	return;
+	/* Default 80x25 text mode with magenta color attribute */
+	halconsole_common.rows = 25;
+	halconsole_common.cols = 80;
+	halconsole_common.attr = 0x05;
```
Maxez, line 361: In kernel we want to initialize default color attribute to cyan - I'd change the comment and set `halconsole_common.attr = 0x03`

HBuczynski: Done

Maxez (approved):
LGTM

hal/ia32/console.c
```diff
@@ -21,133 +21,343 @@
...
+
+
+static void console_memset(volatile u16 *vram, u16 val, unsigned int n)
+{
+	unsigned int i;
 
-	int attr;
-} console;
+	for (i = 0; i < n; i++)
+		*(vram + i) = val;
```
kemonats, line 65: Maybe inline assembler and using `movs`. It will be faster.

hal/ia32/console.c
```diff
@@ -21,133 +21,343 @@
...
-			break;
+							row = halconsole_common.parms[0] - 1;
+							col = halconsole_common.parms[1] - 1;
+							halconsole_common.esc = esc_init;
+							break;
+
+						case 'J':
+							switch (halconsole_common.parms[0]) {
+								case 0:
+									console_memset(halconsole_common.vram + row * halconsole_common.cols + col, (u16)halconsole_common.attr << 8 | ' ', halconsole_common.cols * (halconsole_common.rows - row) - col);
```
kemonats, line 195: A line that is 250 characters long?

hal/ia32/console.c
```diff
@@ -21,133 +21,343 @@
...
+									break;
+
+								case 1:
+									console_memset(halconsole_common.vram, (u16)halconsole_common.attr << 8 | ' ', row * halconsole_common.cols + col + 1);
+									break;
+
+								case 2:
+									console_memset(halconsole_common.vram, (u16)halconsole_common.attr << 8 | ' ', halconsole_common.rows * halconsole_common.cols);
+									break;
+							}
```
kemonats, line 205: A line that is 250 characters long?
Maybe someting like that.
```suggestion
							; /* This empty statement must be */
							volatile u16 *vram = halconsole_common.vram;
							const u16 val = (u16)halconsole_common.attr << 8 | ' ';
							unsigned int n = 0;

							switch (halconsole_common.parms[0]) {
								case 0:
									vram += row * halconsole_common.cols + col;
									n = halconsole_common.cols * (halconsole_common.rows - row) - col;
									break;

								case 1:
									// vram = halconsole_common.vram;
									n = row * halconsole_common.cols + col + 1;
									break;

								case 2:
									// vram = halconsole_common.vram;
									n = halconsole_common.rows * halconsole_common.cols;
									break;
							}
							if (n > 0)
								console_memset(vram, val, n);

```

hal/ia32/console.c
HBuczynski: This code is derived from [plo](https://github.com/phoenix-rtos/plo/blob/f5401788f72d58658a2aa1879ccfabc475cc900b/hal/ia32/console-vga.c#L79-L314) so in this PR I would not like to add significant modifications.

hal/ia32/console.c
HBuczynski: This code is derived from [plo](https://github.com/phoenix-rtos/plo/blob/f5401788f72d58658a2aa1879ccfabc475cc900b/hal/ia32/console-vga.c#L79-L314) so it slightly differs from master branch.

`volatile` is correct, the vga hardware uses simultaneously from `halconsole_common.vram`.

We prefer not to write variables declaration inside `switch ... case`. I leave it as it is.

kemonats: But it seems to me that the hardware only reads from vram, it doesn't change its contents. You're just changing the contents of that memory.

agkaminski: If hardware only reads this memory it must be volatile anyway - we need to make sure that we actually write to this memory

kemonats: If that's the case, what's changed that isn't in the master branch but is here.

PR #279, opened 2021-12-20
Agkamins/stm32 feature syspage

agkaminski:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/stm32/l4/timer.c
+++ b/hal/armv7m/stm32/l4/timer.c
@@ -83,6 +83,40 @@ static int timer_irqHandler(unsigned int n, cpu_context_t *ctx, void *arg)
 	return 1;
 }
 
+
+static time_t hal_timerCyc2Us(time_t ticks)
+{
+	return (ticks * 1000 * 1000) / (32768 / (1 << PRESCALER));
+}
+
+
+static time_t hal_timerUs2Cyc(time_t us)
+{
+	return ((32768 / (1 << PRESCALER)) * us + (500 * 1000)) / (1000 * 1000);
+}
+
+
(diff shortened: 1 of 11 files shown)


hal/armv7m/stm32/l4/timer.c
```diff
@@ -161,10 +161,18 @@ time_t hal_timerGetCyc(void)
 
 int hal_timerRegister(int (*f)(unsigned int, cpu_context_t *, void *), void *data, intr_handler_t *h)
 {
+	int err;
+
 	h->f = f;
 	h->n = lptim1_irq;
```
HBuczynski, line 167: `lptim1_irq` is registered in `hal_timerInit`, so I think that mistake is related only to wrong irq number.
What do you think, about changing only irq number to `SYSTICK_IRQ` instead of registering two handlers?

HBuczynski (commented):
After your changes in `proc/threads.c`, timer API can be simplified. The following functions are not longer used and can be removed:
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/5bac12a1dfc082f428117e5a27394f3ce064a529/hal/timer.h#L23-L29

It is not related to this PR, but I think it is worth to do it.

hal/armv7m/stm32/l4/timer.c
agkaminski: You're right, changed, thx

HBuczynski:
Thanks for the changing timer API.
However, making some functions static and unused in `timer.c` generates some additional warnings.

HBuczynski (approved):

PR #286, opened 2022-01-21
posix/unix: handle correctly removal of connected socket

ziemleszcz:
- set peer socket to disconnected state
- send on disconnected socket returns EPIPE (SOCK_STREAM, SOCK_SEQPACKET)
- send on disconnected socket return ECONNREFUSED (SOCK_DGRAM)
- send on disconnected socket sends SIGPIPE (SOCK_STREAM, SOCKSEQPACKET)
- recv on disconnected sockets returns 0 (SOCK_STREAM, SOCK_SEQPACKET)

JIRA: DTR-241

## Motivation and Context

Required for openiked. Currently send/recv on socket connected to closed socket will crash kernel.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/posix/unix.c
+++ b/posix/unix.c
@@ -26,11 +26,11 @@
 #define US_MIN_BUFFER_SIZE SIZE_PAGE
 #define US_MAX_BUFFER_SIZE 65536
 
-#define US_BOUND      (1 << 0)
-#define US_LISTENING  (1 << 1)
-#define US_ACCEPTING  (1 << 2)
-#define US_CONNECTING (1 << 3)
-
+#define US_BOUND       (1 << 0)
+#define US_LISTENING   (1 << 1)
+#define US_ACCEPTING   (1 << 2)
+#define US_CONNECTING  (1 << 3)
+#define US_PEER_CLOSED (1 << 4)
 
 typedef struct _unixsock_t {
 	rbnode_t linkage;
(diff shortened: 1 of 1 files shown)


posix/unix.c
```diff
@@ -218,20 +218,26 @@ static unixsock_t *unixsock_get(unsigned id)
...
 	proc_lockSet(&unix_common.lock);
-	if (--r->refs <= 0) {
-		lib_rbRemove(&unix_common.tree, &r->linkage);
+	if (--s->refs <= 0) {
+		lib_rbRemove(&unix_common.tree, &s->linkage);
+
+		/* TODO: handle connecting socket */
+
+		if (s->connect != NULL && !(s->state & US_DISCONNECTED))
+			s->connect->state |= US_DISCONNECTED;
```
nalajcie, line 230: maybe we should set s->connect->connect = NULL  to avoid dangling pointers?

posix/unix.c
```diff
@@ -698,6 +713,11 @@ static ssize_t send(unsigned socket, const void *buf, size_t len, int flags, con
 				err = -ENOTCONN;
 				break;
 			}
+			else if (s->state & US_DISCONNECTED) {
```
nalajcie, line 716: not sure, but it seems that we will be sending `SIGPIPE` on every `send()`, maybe we should instead send it once and set `s->connect = NULL` here and and clear `US_DISCONNECTED` flag?

posix/unix.c
```diff
@@ -26,11 +26,11 @@
...
-#define US_BOUND      (1 << 0)
-#define US_LISTENING  (1 << 1)
-#define US_ACCEPTING  (1 << 2)
-#define US_CONNECTING (1 << 3)
-
+#define US_BOUND        (1 << 0)
+#define US_LISTENING    (1 << 1)
+#define US_ACCEPTING    (1 << 2)
+#define US_CONNECTING   (1 << 3)
+#define US_DISCONNECTED (1 << 4) /* NOTE: invalidates connect pointer */
```
nalajcie, line 33: Hm... new socket is not in `disconnected` state, it's a special flag that indicates the socket was already connected and than disconnected from the other side (sooooo.... it's "disconnecting" now as "disconnected" sockets have state == 0?)

posix/unix.c
```diff
@@ -373,6 +379,7 @@ int unix_accept4(unsigned socket, struct sockaddr *address, socklen_t *address_l
 		LIST_REMOVE(&s->connect, conn);
 		s->state &= ~US_ACCEPTING;
 
+		conn->state &= ~US_DISCONNECTED;
```
nalajcie, line 382: maybe only clearing after sending SIGPIPE (+ add sigpipe sending to all socket operations?) to reset to "disconnected" state?

nalajcie (commented):
Not sure about full unix socket semantics, so leaving only comments ;)

posix/unix.c
ziemleszcz: Yeah, dangling pointer is not nice.

posix/unix.c
ziemleszcz: Linux sends SIGPIPE on every send.

posix/unix.c
ziemleszcz: It should really be US_PEER_DISCONNECTED but I've decided it's too long.

posix/unix.c
ziemleszcz: I don't think SIGPIPE should change socket state, it's related only to send() call (based on POSIX spec).

nalajcie: ok, it was just an idea to send `SIGPIPE` only once, if not needed - let's drop it.

ziemleszcz: Turns out Linux forbids reconnecting stream socket, connect() call on peer disconnected socket return EISCONN.

agkaminski (approved):
LGTM

ziemleszcz:
Hey, not so fast. I marked that I will merge it myself. It was draft commit.

PR #287, opened 2022-01-24
posix/unix: add unix sockets polling

ziemleszcz:
JIRA: DTR-240

## Types of changes

## How Has This Been Tested?

- [x] New test added: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/85).

## Checklist:

## Special treatment

### DIFF
--- a/posix/unix.c
+++ b/posix/unix.c
@@ -218,13 +218,27 @@ static unixsock_t *unixsock_get(unsigned id)
 }
 
 
+static unixsock_t *unixsock_get_connected(unixsock_t *s)
+{
+	unixsock_t *r;
+
+	proc_lockSet(&unix_common.lock);
+	r = s->connect;
+	if (r != NULL)
+		r->refs++;
+	proc_lockClear(&unix_common.lock);
+
+	return r;
+}
+
+
 static void unixsock_put(unixsock_t *s)
 {
 	proc_lockSet(&unix_common.lock);
 	if (--s->refs <= 0) {
 		lib_rbRemove(&unix_common.tree, &s->linkage);
 
-		/* TODO: handle connecting socket */
+		/* FIXME: handle connecting socket */
 
 		if (s->connect != NULL) {
 			s->connect->state |= US_PEER_CLOSED;
@@ -380,6 +394,8 @@ int unix_accept4(unsigned socket, struct sockaddr *address, socklen_t *address_l
 
 		LIST_REMOVE(&s->connect, conn);
 		s->state &= ~US_ACCEPTING;
+
+		/* FIXME: handle connecting socket removal */
 
 		conn->state &= ~US_PEER_CLOSED;
 		conn->connect = new;
@@ -544,6 +560,8 @@ int unix_connect(unsigned socket, const struct sockaddr *address, socklen_t addr
 
 			_cbuffer_init(&s->buffer, v, s->buffsz);
 
+			/* FIXME: handle remote socket removal */
+
 			hal_spinlockSet(&remote->spinlock, &sc);
 			LIST_ADD(&remote->connect, s);
 			proc_threadWakeup(&remote->queue);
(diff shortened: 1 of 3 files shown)


posix/unix.c
```diff
@@ -947,6 +947,46 @@ int unix_close(unsigned socket)
...
+	else {
+		if (events & (POLLIN | POLLRDNORM | POLLRDBAND)) {
+			proc_lockSet(&s->lock);
+			if (_cbuffer_avail(&s->buffer) > 0)
+				err |= events & (POLLIN | POLLRDNORM | POLLRDBAND);
+			proc_lockClear(&s->lock);
+		}
+
+		if (events & (POLLOUT | POLLRDNORM | POLLRDBAND)) {
+			if ((conn = s->connect) != NULL) {
```
nalajcie, line 967: i'm wondering if we could have a race condition here (memory pointed by `conn` could be freed after this line by doing `unixsock_put` (eg. by `unixsock_close`) on the connected socket. I think that:
 - we should ensure every unixsock_put takes lock to closing socket or is called under lock
 - to modify `s->connect` in `unixsock_put` we should take `s->lock` (and every access to s->connect should be also done under the lock?)
 - (I don't know if we won't introduce some way to deadlock sockets closing, but it's done under the global lock so maybe we're safe - I don't know the usual code paths here)

Edit: maybe better would be to use `unixsock_get(s->connect_fd)` to ensure the connected socket won't be deleted while we're operating on it? (seems like a cleaner solution to me)

ziemleszcz: Yeah, good catch. Unfortunately this code is probably full of race conditions. I don't wont to dive into this right now. In this particular case your "unixsock_get" suggestion should fix it. Another option is to acquire "s->lock" in "unixsock_put" before destroying the socket - this should also fix the same problem in "send" function.

nalajcie: Yeah, IMHO we should always use `unixsock_get` when we're operating on a peer socket, but apart from the poll it's out of scope of this PR (maybe we should add a task not to forget it?). Socket deletion should also be reviewed. I was thinking about switching field `connected` to `connected_fd`, but this is not enough to fix deletion use case (and would result in strange behavior if somebody would close the connected socket and open new one reusing the FD :)).

I'll go for using `unixsock_get` here and at least adding FIXME in places we suspect race condition may occur.

nalajcie: IMHO there still could be a race condition (`s->connect` could be freed between  `s->connect != NULL` access and  && `conn = unixsock_get(s->connect->id)` - maybe add FIXME there? But now the window for race condition is much smaller :)

nalajcie (approved):

posix/unix.c
```diff
@@ -218,13 +218,30 @@ static unixsock_t *unixsock_get(unsigned id)
...
+{
+	unixsock_t *r;
+
+	proc_lockSet(&unix_common.lock);
+	if (s->connect) {
+		r = s->connect;
+		r->refs++;
+	}
+	else
+		r = NULL;
```
agkaminski, line 231: Coding style - missing bracket for `else`

posix/unix.c
```diff
@@ -218,13 +218,30 @@ static unixsock_t *unixsock_get(unsigned id)
 }
 
 
+static unixsock_t *unixsock_get_connected(unixsock_t *s)
+{
+	unixsock_t *r;
+
+	proc_lockSet(&unix_common.lock);
+	if (s->connect) {
```
agkaminski, line 226: Coding style
```suggestion
	if (s->connect != NULL) {
```

posix/unix.c
```diff
@@ -218,13 +218,30 @@ static unixsock_t *unixsock_get(unsigned id)
...
+		r = s->connect;
```
agkaminski, line 227: We assign `r = s->connect` in both cases, this can be simplified

agkaminski (approved):

PR #297, opened 2022-02-09
imx6ull: add short ddr test

jsarzy:
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imx6ull.

## Checklist:

## Special treatment


agkaminski:
My suggestion: move short test to the separate function, jump from asm to something like testStart and invoke either testShort or testAll depending on the define. Right now placement of the short test is bad - why we enter `while(1)` if we always `break` there?

agkaminski (approved):
I wonder if short test shouldn't be the default one, but that's not really important - LGTM

PR #302, opened 2022-02-11
posix/waitpid: Fixed waitpid with WNOHANG when there's a zombie,

agkaminski:
but of a process we don't wait for

JIRA: RTOS-155

## Description

Case when there is a zombie, but of a different process than the one we're waiting for wasn't handled at all - caused tight loop in waitpid.

## Motivation and Context

Fixes problem found in waitpid test

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic).

## Checklist:

## Special treatment

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -2375,7 +2375,7 @@ int posix_waitpid(pid_t child, int *status, int options)
 		}
 		while (!found && (c = c->next) != pinfo->zombies);
 	}
-	while (!found);
+	while (!found && !(options & 1));
 	proc_lockClear(&pinfo->lock);
 
 	if (found) {


posix/posix.c
```diff
@@ -2375,7 +2375,7 @@ int posix_waitpid(pid_t child, int *status, int options)
 		}
 		while (!found && (c = c->next) != pinfo->zombies);
 	}
-	while (!found);
+	while (!found && !(options & 1));
```
HBuczynski, line 2378: To not use magic numbers in the code, maybe it would be worth to define `WNOHANG` as `1 << 0`

agkaminski: @HBuczynski it is defined in libphoenix, I expect this to be solved by @Maxez during posix headers cleanup (it should be defined in kernel's include and used b6 libphoenix)

HBuczynski (approved):

PR #303, opened 2022-02-14
proc: add locks priority inheritance

Maxez:
## Description

Added locks priority inheritance. Fixes priority inversion issue.

## Motivation and Context

[JIRA: DTR-269](https://jira.phoenix-rtos.com/browse/DTR-269)

## Types of changes

## How Has This Been Tested?

- [x] New test added:
https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/90
- [x] Tested by hand on: ia32-generic.

## Checklist:

## Special treatment

### DIFF
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -462,54 +462,74 @@ int threads_timeintr(unsigned int n, cpu_context_t *context, void *arg)
  */
 
 
-static void thread_destroy(thread_t *t)
+static void proc_lockUnlock(lock_t *lock);
+
+
+static void thread_destroy(thread_t *thread)
 {
 	process_t *process;
 	spinlock_ctx_t sc;
 
-	perf_end(t);
+	perf_end(thread);
 
-	vm_kfree(t->kstack);
+	/* No need to protect thread->locks access with threads_common.spinlock */
+	/* The destroyed thread is a ghost and no thread (except for the current one) can access it */
+	while (thread->locks != NULL) {
+		proc_lockUnlock(thread->locks);
+	}
+	vm_kfree(thread->kstack);
 
-	if ((process = t->process) != NULL) {
+	process = thread->process;
+	if (process != NULL) {
 		hal_spinlockSet(&threads_common.spinlock, &sc);
-		LIST_REMOVE_EX(&process->threads, t, procnext, procprev);
-		LIST_ADD_EX(&process->ghosts, t, procnext, procprev);
-		_proc_threadWakeup(&process->reaper);
-		hal_spinlockClear(&threads_common.spinlock, &sc);
 
+		LIST_REMOVE_EX(&process->threads, thread, procnext, procprev);
+		LIST_ADD_EX(&process->ghosts, thread, procnext, procprev);
+		_proc_threadWakeup(&process->reaper);
+
+		hal_spinlockClear(&threads_common.spinlock, &sc);
 		proc_put(process);
 	}
-	else
-		vm_kfree(t);
+	else {
+		vm_kfree(thread);
+	}
 }
 
 
 thread_t *threads_findThread(int tid)
 {
-	thread_t *r, t;
-	t.id = tid;
+	thread_t *thread, t;
 
+	t.id = tid;
 	proc_lockSet(&threads_common.lock);
-	if ((r = lib_treeof(thread_t, idlinkage, lib_rbFind(&threads_common.id, &t.idlinkage))) != NULL)
-		r->refs++;
+
+	thread = lib_treeof(thread_t, idlinkage, lib_rbFind(&threads_common.id, &t.idlinkage));
+	if (thread != NULL) {
+		thread->refs++;
+	}
+
 	proc_lockClear(&threads_common.lock);
 
-	return r;
+	return thread;
 }
 
 
-void threads_put(thread_t *t)
+void threads_put(thread_t *thread)
 {
-	int remaining;
+	int refs;
 
 	proc_lockSet(&threads_common.lock);
-	if (!(remaining = --t->refs))
-		lib_rbRemove(&threads_common.id, &t->idlinkage);
+
+	refs = --thread->refs;
+	if (refs <= 0) {
+		lib_rbRemove(&threads_common.id, &thread->idlinkage);
+	}
+
 	proc_lockClear(&threads_common.lock);
 
-	if (!remaining)
-		thread_destroy(t);
+	if (refs <= 0) {
+		thread_destroy(thread);
+	}
 }
 
 
@@ -568,6 +588,7 @@ int threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
 		if (!selected->exit /*|| hal_cpuSupervisorMode(selected->context)*/)
 			break;
 
+		selected->state = GHOST;
 		LIST_ADD(&threads_common.ghosts, selected);
 		_proc_threadWakeup(&threads_common.reaper);
 	}
@@ -778,7 +799,6 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 		vm_kfree(t);
 		return -ENOMEM;
 	}
-
 	hal_memset(t->kstack, 0xba, t->kstacksz);
 
 	t->state = READY;
@@ -791,6 +811,7 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 	t->exit = 0;
 	t->execdata = NULL;
 	t->wait = NULL;
+	t->locks = NULL;
 
 	thread_alloc(t);
 
@@ -799,6 +820,7 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 
 	t->stick = 0;
 	t->utick = 0;
+	t->priorityBase = priority;
 	t->priority = priority;
 
 	if (process != NULL) {
@@ -809,7 +831,7 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 
 	t->execdata = NULL;
 
-	/* Insert thread to global quee */
+	/* Insert thread to global queue */
 	proc_lockSet(&threads_common.lock);
 	lib_rbInsert(&threads_common.id, &t->idlinkage);
 
@@ -841,6 +863,92 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 }
 
 
+static unsigned int _proc_lockGetPriority(lock_t *lock)
+{
+	unsigned int priority = sizeof(threads_common.ready) / sizeof(threads_common.ready[0]) - 1;
+	thread_t *thread = lock->queue;
+
+	if (thread != NULL) {
+		do {
+			if (thread->priority < priority) {
+				priority = thread->priority;
+			}
+			thread = thread->next;
+		} while (thread != lock->queue);
+	}
+
+	return priority;
+}
+
+
+static unsigned int _proc_threadGetPriority(thread_t *thread)
+{
+	unsigned int ret, priority = thread->priorityBase;
+	lock_t *lock = thread->locks;
+
+	if (lock != NULL) {
+		do {
+			ret = _proc_lockGetPriority(lock);
+			if (ret < priority) {
+				priority = ret;
+			}
+			lock = lock->next;
+		} while (lock != thread->locks);
+	}
+
+	return priority;
+}
+
+
+static void _proc_threadSetPriority(thread_t *thread, unsigned int priority)
+{
+	unsigned int i;
+
+	if (thread->state == READY) {
+		for (i = 0; i < hal_cpuGetCount(); i++) {
+			if (thread == threads_common.current[i]) {
+				break;
+			}
+		}
+
+		if (i == hal_cpuGetCount()) {
+			LIST_REMOVE(&threads_common.ready[thread->priority], thread);
+			LIST_ADD(&threads_common.ready[priority], thread);
+		}
+	}
+
+	thread->priority = priority;
+}
+
+
+int proc_threadPriority(int priority)
+{
+	thread_t *current;
+	spinlock_ctx_t sc;
+	int ret;
+
+	if ((priority < -1) || (priority >= sizeof(threads_common.ready) / sizeof(threads_common.ready[0]))) {
+		return -EINVAL;
+	}
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+
+	current = _proc_current();
+	if (priority >= 0) {
+		/* Don't lower current thread priority if it holds any locks */
+		if ((priority < current->priority) || (current->locks == NULL)) {
+			current->priority = priority;
+		}
+		current->priorityBase = priority;
+	}
+	ret = current->priorityBase;
+
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+
+	return ret;
+}
+
+
 static void _thread_interrupt(thread_t *t)
 {
 	_proc_threadDequeue(t);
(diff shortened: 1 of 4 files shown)


proc/threads.c
```diff
@@ -1275,88 +1353,201 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
 
-int _proc_lockSet(lock_t *lock, int interruptible, spinlock_ctx_t *sc)
+static unsigned int _proc_lockGetPriority(lock_t *lock)
 {
-	while (lock->v == 0) {
-		if (proc_threadWaitEx(&lock->queue, &lock->spinlock, 0, interruptible, sc) == -EINTR)
+	unsigned int ret = -1;
+	thread_t *t;
+
+	if ((t = lock->queue) != NULL) {
```
agkaminski, line 1361: MISRA

agkaminski: and other instances of if ((var = func()) != const)

proc/threads.c
```diff
@@ -1275,88 +1353,201 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+
+		if (refs == 0) {
+			thread_destroy(owner);
+		}
+
+		hal_spinlockSet(&threads_common.spinlock, &sc);
+
+		_proc_threadSetPriority(owner, _proc_threadGetPriority(owner));
+
+		hal_spinlockClear(&threads_common.spinlock, &sc);
```
agkaminski, line 1528:
```suggestion
		if (refs == 0) {
			thread_destroy(owner);
		}
		else {
			hal_spinlockSet(&threads_common.spinlock, &sc);

			_proc_threadSetPriority(owner, _proc_threadGetPriority(owner));

			hal_spinlockClear(&threads_common.spinlock, &sc);
		}
```

proc/threads.c
```diff
@@ -1366,47 +1557,81 @@ int proc_lockSet2(lock_t *l1, lock_t *l2)
...
+		hal_spinlockSet(&owner->spinlock, &sc);
+
+		LIST_REMOVE(&owner->locks, lock);
+		refs = --owner->refs;
+
+		hal_spinlockClear(&owner->spinlock, &sc);
+
+		if (refs == 0) {
+			thread_destroy(owner);
+		}
```
agkaminski, line 1610: We also need to wake up everyone waiting to take the lock and cause them to exit with error

proc/threads.c
```diff
@@ -799,6 +821,7 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 
 	t->stick = 0;
 	t->utick = 0;
+	t->base_priority = priority;
```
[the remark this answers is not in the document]
Maxez: Done.

proc/threads.c
```diff
@@ -841,6 +864,61 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 }
 
 
+static unsigned int _proc_threadGetPriority(thread_t *thread)
+{
+	unsigned int priority, ret = thread->base_priority;
+	lock_t *l;
+
+	if ((l = thread->locks) != NULL) {
```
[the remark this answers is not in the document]
Maxez: Done.

proc/threads.c
```diff
@@ -841,6 +864,61 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
...
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+
+	thread = _proc_current();
+	if (priority == -1) {
+		ret = thread->base_priority;
+	}
+	else {
+		if (thread->priority > priority)
+			thread->priority = priority;
```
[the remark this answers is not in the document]
Maxez: Thanks for spotting this case. Fixed by checking if the current thread holds any locks - if it doesn't we can safely lower its priority.

proc/threads.c
Maxez: Done.

proc/threads.c
Maxez: Done but I'm not sure if my solution really works - I'm going to write a specific test for this scenario.

proc/threads.c
```diff
@@ -467,49 +467,70 @@ static void thread_destroy(thread_t *t)
...
 	proc_lockClear(&threads_common.lock);
 
-	return r;
+	if (thread != NULL) {
+		hal_spinlockSet(&thread->spinlock, &sc);
+
+		thread->refs++;
+
+		hal_spinlockClear(&thread->spinlock, &sc);
+	}
```
agkaminski, line 514: Can we move this under threads_common.lock? There is small window (between lockClear and spinlockSet) during which thread is not protected from being destoyed

proc/threads.c
Maxez: It turned out that `proc_lockDone()` is always called with empty `lock->queue` and there are no waiting threads to wake up.

proc/threads.c
Maxez: Done. I've also added a check for soon to be destroyed thread in `threads_findThread()`.

proc/threads.c
```diff
@@ -1275,88 +1387,228 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+	if (owner != NULL) {
+		hal_spinlockSet(&owner->spinlock, &sc);
+
+		LIST_REMOVE(&owner->locks, lock);
+		refs = --owner->refs;
+
+		hal_spinlockClear(&owner->spinlock, &sc);
+
+		lock->owner = NULL;
+		if (lock->queue != NULL) {
```
agkaminski, line 1568: queue can be -1 - if threadWakeup has been performed on an empty queue (feature of sticky condtionals).

proc/threads.c
```diff
@@ -1275,88 +1387,228 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+		}
+	}
+
+	return ret;
 }
 
 
 int proc_lockClear(lock_t *lock)
 {
+	thread_t *thread = NULL;
```
agkaminski, line 1591: Call it "victim", "toDestroy" or something along these lines imho

proc/threads.c
Maxez: I think that the lock queue can never be equal -1 but I'm adding the change just in case I've missed something. I've also restored calling threads_common.spinlock proteced version of threadWakeup.

proc/threads.c
Maxez: Done.

proc/threads.c
```diff
@@ -841,6 +876,97 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 }
 
 
+static unsigned int _proc_lockGetPriority(lock_t *lock);
```
agkaminski, line 879: No need for this, just move funtion body here

proc/threads.c
```diff
@@ -841,6 +876,97 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
...
+	for (i = 0; i < hal_cpuGetCount(); i++) {
+		if (thread == threads_common.current[i]) {
+			break;
+		}
+	}
+
+	if ((i == hal_cpuGetCount()) && (thread->state == READY)) {
+		LIST_REMOVE(&threads_common.ready[thread->priority], thread);
+		LIST_ADD(&threads_common.ready[priority], thread);
+	}
```
agkaminski, line 920: Do this whole block under if (thread->state == READY) - minor performance improvement

proc/threads.c
```diff
@@ -1275,138 +1402,309 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
  */
 
 
-int _proc_lockSet(lock_t *lock, int interruptible, spinlock_ctx_t *sc)
+static unsigned int _proc_lockGetPriority(lock_t *lock)
 {
-	while (lock->v == 0) {
-		if (proc_threadWaitEx(&lock->queue, &lock->spinlock, 0, interruptible, sc) == -EINTR)
+	unsigned int priority = -1;
```
agkaminski, line 1407: We can safely start with e.g. 7 (sizeof ready list to avoid magic numbers) or thread->priority. -1 is somewhat misleading and can be dangerous if result of this function is used directly

proc/threads.c
```diff
@@ -1275,138 +1402,309 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&lock->spinlock, &sc);
+
+	if (_proc_lockUnlock(lock)) {
+		hal_cpuReschedule(&lock->spinlock, &sc);
+		return;
+	}
+
+	hal_spinlockClear(&lock->spinlock, &sc);
```
agkaminski, line 1588:
```suggestion
	if (_proc_lockUnlock(lock)) {
		hal_cpuReschedule(&lock->spinlock, &sc);
	}
	else {
		hal_spinlockClear(&lock->spinlock, &sc);
	}
```

proc/threads.c
```diff
@@ -1275,138 +1402,309 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
 
 	hal_spinlockSet(&lock->spinlock, &sc);
-	_proc_lockClear(lock);
-	if ((err = proc_threadWaitEx(queue, &lock->spinlock, timeout, 1, &sc)) != -EINTR)
-		_proc_lockSet(lock, 0, &sc);
+
+	err = _proc_lockClear(lock);
+	if (err >= 0) {
+		err = proc_threadWaitEx(queue, &lock->spinlock, timeout, 1, &sc);
+		if (err == EOK) {
```
agkaminski, line 1674: Uh, no, we need to take the lock if -ETIME. Don't take the lock only in -EINTR case.

agkaminski:
Addid @HBuczynski, I don't want to miss anything, this isn't particularly big PR, but it touches complex code

proc/threads.c
Maxez: Done.

proc/threads.c
Maxez: I've changed the starting lock priority to `sizeof(threads_common.ready) / sizeof(threads_common.ready[0]) - 1`. The function can be called on a lock without any waiting threads and should return lowest possible priority in that case.

proc/threads.c
Maxez: Done.

proc/threads.c
Maxez: Done.

proc/threads.c
Maxez: Done.

agkaminski (approved):
LGTM, but let's wait for @HBuczynski to take a look

Maxez:
> LGTM, but let's wait for @HBuczynski to take a look

In the meantime, I've slightly changed `_proc_lockUnlock(lock_t *lock)` to look cleaner.

proc/threads.c
```diff
@@ -841,6 +876,115 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
...
+	thread_t *current = _proc_current();
+	spinlock_ctx_t sc;
+
+	if (priority >= 0) {
+		if (priority <= current->priority) {
+			current->priority = priority;
+		}
+		else {
+			/* Don't lower current thread priority if it holds any locks */
+			hal_spinlockSet(&current->spinlock, &sc);
```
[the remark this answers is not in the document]
Maxez: `current->priority` access is protected by `threads_common.spinlock` which should be set before calling the function.
`current->spinlock` is used for protecting thread `refs` and `locks` members access.

proc/threads.c
```diff
@@ -841,6 +876,115 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
...
+int proc_threadPriority(int priority)
+{
+	spinlock_ctx_t sc;
+	int ret;
+
+	if ((priority < -1) || (priority >= sizeof(threads_common.ready) / sizeof(threads_common.ready[0]))) {
+		return -EINVAL;
+	}
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
```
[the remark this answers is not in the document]
Maxez: Every `thread->priority` access should be protected with the global lock (required by `threads_schedule() to work correctly).
I don't see any case where `_proc_current()` is called without the global lock. Can you be more specific?

proc/threads.c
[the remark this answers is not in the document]
agkaminski: No, spinlock/lock is needed to be held to access resource it protects - so when accessing each resource, appropriate lock needs to be held. Spinlocks are no different than locks on multicore cpus (which we support)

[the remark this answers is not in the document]
agkaminski: If we always have threads_common.spinlock when we're taking thread->spinlock - it means that thread->spinlock is redundant and can be removed altogether.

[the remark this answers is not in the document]
HBuczynski: https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/c4950e6f50ad05b5c99c18d351c03473d9fb53e8/proc/threads.c#L977-L982

https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/c4950e6f50ad05b5c99c18d351c03473d9fb53e8/proc/threads.c#L954-L960

According to the code above I understood that `threads_common.spinlock` protects the global data and `thread->spinlock` protects the majority of thread's data what is a mistake.

Maybe some comment would be helpful in the `thread_t` structure which data are protected by this thread's spinlock.

proc/threads.c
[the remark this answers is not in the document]
HBuczynski: Take a look:
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/c4950e6f50ad05b5c99c18d351c03473d9fb53e8/proc/threads.c#L719

https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/c4950e6f50ad05b5c99c18d351c03473d9fb53e8/proc/threads.c#L1475-L1491

`_proc_lockSet` invokes `_proc_current()`. In this case is set `threads_common.lock.spinlock`.

However, here is set a `threads_common.spinlock`:
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/c4950e6f50ad05b5c99c18d351c03473d9fb53e8/proc/threads.c#L978-L982

Both functions invoke `_proc_current()` with different protection.

proc/threads.c
```diff
@@ -841,6 +876,115 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
 }
 
 
+static unsigned int _proc_lockGetPriority(lock_t *lock)
+{
+	unsigned int priority = sizeof(threads_common.ready) / sizeof(threads_common.ready[0]) - 1;
```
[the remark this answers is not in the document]
HBuczynski: I agree with you that it is compile-time computation but for me it would be more readable to use one define instead of writing each time - `sizeof(threads_common.ready) / sizeof(threads_common.ready[0])` what is prone to mistakes.

proc/threads.c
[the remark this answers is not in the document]
Maxez: > `_proc_lockSet` invokes `_proc_current()`

Where?
`_proc_current()` is invoked only in `_proc_threadPriority()`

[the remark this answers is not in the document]
HBuczynski: `_proc_lockSet` invokes `proc_current()` instead of `_proc_current()`.
I haven't notice that we use two different functions.

Maxez:
After the discussion we've decided to completely remove the `thread->spinlock` field. The `thread->locks` field is now protected by the `threads_common.spinlock` and the `thread->refs` is protected by the `threads_common.lock`.

agkaminski (approved):
Looks good to me. Let's wait for @HBuczynski to review PR again

HBuczynski (approved):

PR #304, opened 2022-02-21
posix: implement sys_futimes() (previously sys_utimes())

ziemleszcz:
JIRA: DTR-276

## Types of changes

## How Has This Been Tested?

- [x] New test added: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/94).

## Checklist:

## Special treatment

### DIFF
--- a/include/syscalls.h
+++ b/include/syscalls.h
@@ -108,6 +108,7 @@
 	\
 	ID(sys_ioctl) \
 	ID(sys_utimes) \
+	ID(sys_futimens) \
 	ID(sys_poll) \
 	\
 	ID(sys_tkill) \
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -1993,11 +1993,44 @@ int posix_setsockopt(int socket, int level, int optname, const void *optval, soc
 int posix_utimes(const char *filename, const struct timeval *times)
 {
 	oid_t oid;
-
 	if (proc_lookup(filename, NULL, &oid) < 0)
 		return -ENOENT;
 
 	return 0;
+}
+
+
+int posix_futimens(int fildes, const struct timespec *times)
+{
+	TRACE("futimens(%d)", fildes);
+
+	open_file_t *f;
+	msg_t msg;
+	int err;
+
+	err = posix_getOpenFile(fildes, &f);
+	if (err < 0)
+		return err;
+
+	hal_memset(&msg, 0, sizeof(msg_t));
+
+	msg.type = mtSetAttr;
+	hal_memcpy(&msg.i.attr.oid, &f->oid, sizeof(oid_t));
+
+	msg.i.attr.type = atMTime;
+	msg.i.attr.val = times[1].tv_sec;
+	err = proc_send(f->oid.port, &msg);
+	if ((err >= 0) && (msg.o.attr.err >= 0)) {
+		msg.i.attr.type = atATime;
+		msg.i.attr.val = times[0].tv_sec;
+		err = proc_send(f->oid.port, &msg);
+	}
+	if (err >= 0)
+		err = msg.o.attr.err;
+
+	posix_fileDeref(f);
+
+	return err;
 }
 
 
--- a/include/posix.h
+++ b/include/posix.h
@@ -136,6 +136,12 @@ struct timeval {
 };
 
 
+struct timespec {
+	time_t tv_sec;
+	long tv_nsec;
+};
+
+
 #define AF_UNSPEC 0
 #define AF_UNIX   1
 #define AF_INET   2
--- a/posix/posix.h
+++ b/posix/posix.h
@@ -143,6 +143,9 @@ extern int posix_poll(struct pollfd *fds, nfds_t nfds, int timeout_ms);
 extern int posix_utimes(const char *filename, const struct timeval *times);
 
 
+extern int posix_futimens(int fildes, const struct timespec *times);
+
+
 extern int posix_tkill(pid_t pid, int tid, int sig);
 
 
--- a/syscalls.c
+++ b/syscalls.c
@@ -1350,6 +1350,18 @@ int syscalls_sys_utimes(char *ustack)
 }
 
 
+int syscalls_sys_futimens(char *ustack)
+{
+	int fildes;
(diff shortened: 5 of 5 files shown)


ziemleszcz:
sys_utimes to be removed later (syscall count will not change).

posix/posix.c
```diff
@@ -1993,14 +1993,45 @@ int posix_setsockopt(int socket, int level, int optname, const void *optval, soc
...
+		return err;
+
+	hal_memset(&msg, 0, sizeof(msg_t));
+
+	msg.type = mtSetAttr;
+	hal_memcpy(&msg.i.attr.oid, &f->oid, sizeof(oid_t));
+
+	msg.i.attr.type = atMTime;
+	msg.i.attr.val = times[1].tv_sec;
+	if (((err = proc_send(f->oid.port, &msg)) < 0) || ((err = msg.o.attr.err) < 0))
```
nalajcie, line 2021: missing `flieDeref`

posix/posix.c
```diff
@@ -1993,14 +1993,45 @@ int posix_setsockopt(int socket, int level, int optname, const void *optval, soc
...
+		return err;
+
+	msg.i.attr.type = atATime;
+	msg.i.attr.val = times[0].tv_sec;
+	if (((err = proc_send(f->oid.port, &msg)) < 0) || ((err = msg.o.attr.err) < 0))
+		return err;
```
nalajcie, line 2027: missing `flieDeref`

include/syscalls.h
```diff
@@ -108,6 +108,7 @@
 	\
 	ID(sys_ioctl) \
 	ID(sys_utimes) \
+	ID(sys_futimes) \
```
nalajcie, line 111: maybe we can remove `sys_utimes` now that it's implemented in userspace?

posix/posix.c
```diff
@@ -1993,14 +1993,45 @@ int posix_setsockopt(int socket, int level, int optname, const void *optval, soc
...
+int posix_futimes(int fildes, const struct timeval *times)
+{
+	TRACE("futimes(%d)", fildes);
+
+	open_file_t *f;
+	msg_t msg;
+	int err;
+
+	if ((err = posix_getOpenFile(fildes, &f)) < 0)
+		return err;
```
nalajcie, line 2012: maybe we should guard against `times` being null (somebody calling syscall directly) and return error (`futimes(_, NULL)` function call handled correctly in libphoenix)

nalajcie (changes requested):
IMHO the decision regarding removing `utimes` syscall should be taken by @agkaminski.

nitpicking: `futimes` is non-posix (but exists on BSD also)... alas `futimens` (nanosecond precision variant) is specified in POSIX.1-2008. Maybe we should use this syscall as a part of 
POSIX compat layer (even if it would support only seconds - precision for now)?

include/syscalls.h
ziemleszcz: I will do it in separate pull request to not break libphoenix building.

posix/posix.c
ziemleszcz: We can assume that this pointer should not be NULL.

include/syscalls.h
nalajcie: ack

posix/posix.c
agkaminski: @nalajcie We should really sanitize all pointers passed to the kernel - not only check if it's not NULL but also if it belongs to the caller's map

ziemleszcz: Agreed but this applies to all syscalls so we should have some common sanitizing code.

agkaminski:
@nalajcie change from utimes to futimes is ok - kernel knows about file descriptors, everything can be done in user space. Do we have already clear goal of specific POSIX version compliance? If we want to introduce nanosecond precision much of existing API need to change - imho thats something to do in the future, for now futimes is ok to me.

nalajcie:
> @nalajcie change from utimes to futimes is ok - kernel knows about file descriptors, everything can be done in user space. Do we have already clear goal of specific POSIX version compliance? If we want to introduce nanosecond precision much of existing API need to change - imho thats something to do in the future, for now futimes is ok to me.

It's OK for now pro me also, just doesn't sit well with me that we're adding non-posix function to posix.c :).

FYI, on phoenix-rtos.com webpage it's written that we're aiming for  IEEE Std 1003.1-2017 (so POSIX.1-2017) compliance (but we may change that ofc).

Other than that, LGTM (but assuming @agkaminski should give approval :man_mechanic:)

agkaminski (approved):

posix/posix.c
```diff
@@ -1993,14 +1993,60 @@ int posix_setsockopt(int socket, int level, int optname, const void *optval, soc
...
+	msg.i.attr.val = times[1].tv_sec;
+	err = proc_send(f->oid.port, &msg);
+	if (err < 0) {
+		posix_fileDeref(f);
+		return err;
+	}
+	if (msg.o.attr.err < 0) {
+		posix_fileDeref(f);
+		return msg.o.attr.err;
+	}
```
kemonats, line 2030: You can simplify it a litle bit.
```suggestion
	if ((err >= 0) && (msg.o.attr.err < 0)) {
		err = msg.o.attr.err;
	}
	if (err < 0) {
		posix_fileDeref(f);
		return err;
	}
```

posix/posix.c
```diff
@@ -1993,14 +1993,60 @@ int posix_setsockopt(int socket, int level, int optname, const void *optval, soc
...
+	msg.i.attr.val = times[0].tv_sec;
+	err = proc_send(f->oid.port, &msg);
+	if (err < 0) {
+		posix_fileDeref(f);
+		return err;
+	}
+	if (msg.o.attr.err < 0) {
+		posix_fileDeref(f);
+		return msg.o.attr.err;
+	}
```
kemonats, line 2042: and this
```suggestion
	if (err >= 0) {
		err = EOK;
		if (msg.o.attr.err < 0) {
			err = msg.o.attr.err;
		}
	}
```

nalajcie (approved):

PR #317, opened 2022-03-24
proc: change process map under threads_common.spinlock

HBuczynski:
## Description

Process map is changed under threads_common.spinlock to prevent scheduling.

## Motivation and Context

PD-250

## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: (zynq7000, ia32-generic).

## Checklist:

## Special treatment

### DIFF
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -1003,6 +1003,17 @@ void proc_reap(void)
 }
 
 
+void proc_changeMap(process_t *proc, vm_map_t *map, pmap_t *pmap)
+{
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	proc->mapp = map;
+	proc->pmapp = pmap;
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+}
+
+
 /*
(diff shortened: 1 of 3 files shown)


agkaminski (commented):
Sadly, pmapp != mmapp->pmapp on NOMMU targets. It's separated, as by design mmap is a global, shared entity, but pmap is a per process entity (to have MPU entries stored in it in the future)

proc/threads.c
```diff
@@ -1003,6 +1003,23 @@ void proc_reap(void)
...
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	if (map != NULL && pmap != NULL) {
+		proc->mapp = map;
+		proc->pmapp = pmap;
+	}
+	else {
+		proc->mapp = NULL;
+		proc->pmapp = NULL;
+	}
```
agkaminski, line 1018: This is equivalent to simple:
```suggestion
	proc->mapp = map;
	proc->pmapp = pmap;
```

agkaminski: We have no use case of changing only mapp or pmapp, both are always changed at the same time

PR #318, opened 2022-03-25
proc: fix condition for lowering thread priority in proc_threadPriority()

Maxez:
## Description

Thread which holds locks can lower its priority (up to the highest priority of the threads waiting for any of the locks it took).

## Motivation and Context

[JIRA: DTR-281](https://jira.phoenix-rtos.com/browse/DTR-281)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic, armv7a7-imx6ull.

## Checklist:

## Special treatment

### DIFF
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -870,9 +870,9 @@ static unsigned int _proc_lockGetPriority(lock_t *lock)
 }
 
 
-static unsigned int _proc_threadGetPriority(thread_t *thread)
+static unsigned int _proc_threadGetLockPriority(thread_t *thread)
 {
-	unsigned int ret, priority = thread->priorityBase;
+	unsigned int ret, priority = sizeof(threads_common.ready) / sizeof(threads_common.ready[0]) - 1;
 	lock_t *lock = thread->locks;
 
 	if (lock != NULL) {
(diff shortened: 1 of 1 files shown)


proc/threads.c
```diff
@@ -924,8 +924,7 @@ int proc_threadPriority(int priority)
 
 	current = _proc_current();
 	if (priority >= 0) {
-		/* Don't lower current thread priority if it holds any locks */
-		if ((priority < current->priority) || (current->locks == NULL)) {
+		if (priority < _proc_threadGetPriority(current)) {
```
agkaminski, line 927: Hmm, `_proc_threadGetPriority` cannot return priority lower than `current->priority` (because `current-priority` is either `current->priorityBase` or it has been set by previous call to `_proc_threadGetPriority)`, so if `priority < _proc_threadGetPriority(current)` then also `priority < current->priority`. Or am I missing something?

Maxez: You're right, the current thread priority shouldn't be taken into account here at all. Only priority of the threads waiting on the locks taken by the current thread matter. I've added new function `_proc_threadGetLockPriority()` which returns thread `lock priority` - max priority of the threads waiting for any of the locks taken by the given thread.

agkaminski (approved):
LGTM

PR #322, opened 2022-04-26
vm: Added shared memory maps info to vm_mapinfo

agkaminski:
JIRA: ISC-163

## Motivation and Context

Currently there is a problem with estimation of allocated/available memory on NOMMU targets. Shared maps parameters provide adequate information about total system memory.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4, ia32-generic).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/129).

### DIFF
--- a/vm/map.c
+++ b/vm/map.c
@@ -36,6 +36,7 @@ struct {
 	map_entry_t *entries;
 
 	vm_map_t **maps;
+	size_t mapssz;
 } map_common;
 
 
@@ -861,6 +862,7 @@ void vm_mapinfo(meminfo_t *info)
 	unsigned int size;
 	process_t *process;
 	int i;
+	size_t total, free;
 
 
 	proc_lockSet(&map_common.lock);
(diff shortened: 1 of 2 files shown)


agkaminski:
This change will compile without changes in phoenix-rtos-utils, but psh/mem will not work - wrong parameters would be passed to the kernel which will cause psh memory to be overwritten.

vm/map.c
```diff
@@ -1003,6 +1005,48 @@ void vm_mapinfo(meminfo_t *info)
...
 	}
+
+	if (info->maps.mapsz != -1) {
+		info->maps.total = 0;
+		info->maps.free = 0;
+
+		for (size = 0; size < map_common.mapssz; ++size) {
+			map = vm_getSharedMap(NULL, (int)size);
+			if (map == NULL) {
+				if ((int)size < info->maps.mapsz) {
```
HBuczynski, line 1016: If `size >= info->maps.mapsz`, there is nothing to do and You can exit the loop.
If You accept this suggestion, You can remove `if` in line `1016` and `1037`.

```suggestion
        for (size = 0; size < map_common.mapssz; ++size) {
                if ((int)size >= info->maps.mapsz) {
                        break;
                }
		map = vm_getSharedMap(NULL, (int)size);
```

HBuczynski (changes requested):

vm/map.c
agkaminski: Nope, there is something to do - e.g. use case with info->map.mapsz = 0 exist - if I only want total summary, without additional information about specific maps.

vm/map.c
```diff
@@ -861,6 +862,7 @@ void vm_mapinfo(meminfo_t *info)
 	unsigned int size;
```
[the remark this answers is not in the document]
agkaminski: It kinda is a `size` - see line 1048. Also, I didn't want to change old code too much, it's use of `size` is identical.

HBuczynski (approved):

vm/map.c
```diff
@@ -1054,19 +1098,31 @@ void vm_mapGetStats(size_t *allocsz)
...
-	if (map < 0)
-		map = prog->dmaps[0];
+	if (prog == NULL) {
+		if (map >= 0 && (size_t)map < map_common.mapssz) {
+			ret = map_common.maps[map];
+		}
+	}
+	else {
+		if (map < 0) {
+			map = prog->dmaps[0];
```
anglov, line 1113: TBH I don' t like that here we assigning variable and two lines below checking if it's equal.  I have to read it twice to recognize what was intended.

But it's old code and changing it increase indentation without early return. So approving anyway.

anglov (approved):

PR #331, opened 2022-07-04
zynq7000/console: Use UART pin definitions from board_config.h

MaciejPurski:
JIRA: RTOS-203

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (armv7a9-zynq7000-zturn).

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/zynq7000/zynq.c
+++ b/hal/armv7a/zynq7000/zynq.c
@@ -89,7 +89,7 @@ static void _zynq_slcrUnlock(void)
 }
 
 
-static int _zynq_setAmbaClk(u32 dev, u32 state)
+int _zynq_setAmbaClk(u32 dev, u32 state)
 {
 	/* Check max dev position in amba register */
 	if (dev > 24)
@@ -306,7 +306,7 @@ static int _zynq_getMioClk(char *ref0, char *mux0, char *ref1, char *mux1)
 }
 
 
-static int _zynq_setMIO(unsigned int pin, char disableRcvr, char pullup, char ioType, char speed, char l0, char l1, char l2, char l3, char triEnable)
+int _zynq_setMIO(unsigned int pin, char disableRcvr, char pullup, char ioType, char speed, char l0, char l1, char l2, char l3, char triEnable)
 {
 	u32 val = 0;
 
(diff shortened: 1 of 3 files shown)


agkaminski (dismissed):

hal/armv7a/zynq7000/zynq.h
```diff
@@ -0,0 +1,25 @@
...
+ * %LICENSE%
+ */
+
+#ifndef _HAL_ZYNQ_H_
+#define _HAL_ZYNQ_H_
+
+#include "../../cpu.h"
+
+
+int _zynq_setMIO(unsigned int pin, char disableRcvr, char pullup, char ioType, char speed, char l0, char l1, char l2, char l3, char triEnable);
```
HBuczynski, line 22: Missing `extern`.

hal/armv7a/zynq7000/zynq.h
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#ifndef _HAL_ZYNQ_H_
+#define _HAL_ZYNQ_H_
+
+#include "../../cpu.h"
```
HBuczynski, line 19: Probably, it is not needed.

hal/armv7a/zynq7000/zynq.c
```diff
@@ -595,32 +595,7 @@ static void _zynq_peripherals(void)
...
-
-
-	/* Define UARTs' clocks speed
-	 * IO_PLL / 20 :  1000 MHz / 20 = 50 MHz
-	 * CLKACT0 = 0x0; CLKACT1 = 0x1; SRCSEL = 0x0; DIVISOR = 0x14 */
-	*(zynq_common.slcr + slcr_uart_clk_ctrl) = (*(zynq_common.slcr + slcr_uart_clk_ctrl) & ~0x00003f33) | 0x00001402 | 0x1;
-
-	/* Enable UART_0 & UART_1 clock */
+	/* Enable UART_0 & UART_1 clock simultaneously to avoid race condition */
 	*(zynq_common.slcr + slcr_aper_clk_ctrl) = *(zynq_common.slcr + slcr_aper_clk_ctrl) | (1 << 21) | (1 << 20);
```
HBuczynski, line 599: You can make `_zynq_setAmbaClk` as a part of the interface, similar to the `_zynq_setMIO` and `_zynq_peripherals` would be removed. Then everything related to uart will be done in `console`.

MaciejPurski: I was considering using `_zynq_setAmbaClk` here, however this function sets only one clock at a time. From, what I understood from @agkaminski, we want to set both clocks simultaneously with just one write, so I left it like it is.

hal/armv7a/zynq7000/zynq.h
MaciejPurski: I know we use `extern` in many places, but not everywhere and also it has no effect here.

HBuczynski: Take a look on other kernel modules. In the majority of headers there is an `extern` keyword. I think that code should be consistent.

PR #337, opened 2022-07-25
Replace usleep with nsleep to allow nanosleep implementation

agkaminski:
## Description

Replace usleep with nsleep:
- accept greater time resolution (us -> ns)
- allow returning "unsleept" time

## Motivation and Context

Greater time resolution is needed by `nanosleep()`, returning unsleept time is needed by both `sleep()` and `nanosleep()`.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (is dependent on https://github.com/phoenix-rtos/libphoenix/pull/192).

### DIFF
--- a/syscalls.c
+++ b/syscalls.c
@@ -261,12 +261,45 @@ int syscalls_endthread(void *ustack)
 }
 
 
-int syscalls_usleep(void *ustack)
+int syscalls_nsleep(void *ustack)
 {
-	unsigned int us;
+	time_t *sec;
+	long int *nsec;
+	time_t start, us, stop, diff;
+	int ret;
 
-	GETFROMSTACK(ustack, unsigned int, us, 0);
-	return proc_threadSleep((unsigned long long)us);
+	GETFROMSTACK(ustack, time_t *, sec, 0);
+	GETFROMSTACK(ustack, long int *, nsec, 1);
+
+	/* TODO - user pointer check */
+	if (sec == NULL || nsec == NULL) {
+		return -EINVAL;
+	}
+
+	/* FIXME - time_t should be signed and we should check for *sec < 0 */
+	if (*nsec < 0 || *nsec >= 1000 * 1000 * 1000) {
+		return -EINVAL;
+	}
+
+	proc_gettime(&start, NULL);
+
+	us = *sec * 1000 * 1000 + (*nsec + 999) / 1000;
+
+	ret = proc_threadSleep(us);
+
+	*sec = 0;
+	*nsec = 0;
+
+	if (ret == -EINTR) {
+		proc_gettime(&stop, NULL);
+		diff = stop - start;
+		if (diff < us) {
+			*sec = diff / (1000 * 1000);
+			*nsec = (diff % (1000 * 1000)) * 1000;
+		}
+	}
+
+	return ret;
 }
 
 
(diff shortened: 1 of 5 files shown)


syscalls.c
```diff
@@ -261,12 +261,45 @@ int syscalls_endthread(void *ustack)
 }
 
 
-int syscalls_usleep(void *ustack)
+int syscalls_nsleep(void *ustack)
 {
-	unsigned int us;
+	time_t *sec;
+	long int *nsec;
+	time_t start, sleep, stop, diff;
```
nalajcie, line 268: maybe don't use `sleep` as a variable name as it's a c-standard-library symbol (thus highlighted here)

syscalls.c
```diff
@@ -261,12 +261,45 @@ int syscalls_endthread(void *ustack)
...
+	}
+
+	/* FIXME - time_t should be signed and we should check for *sec < 0 */
+	if (*nsec < 0 || *nsec >= 1000 * 1000 * 1000) {
+		return -EINVAL;
+	}
+
+	proc_gettime(&start, NULL);
+
+	sleep = *sec * 1000 * 1000 + (*nsec + 999) / 1000;
```
nalajcie, line 286: this doesn't fix the problem with overflowing microseconds when `sleep(4295)` is being used, can this be taken care of (probably better in kernel to avoid ^C^V in libphoenix `nanosleep` and `sleep` implementations)

see: https://github.com/phoenix-rtos/libphoenix/issues/93

nalajcie (commented):
ok, but maybe You could fix https://github.com/phoenix-rtos/libphoenix/issues/93 as a side effect? :)

IMHO we should write "ABI CHANGE" in the commit as it breaks userspace API

syscalls.c
nalajcie: edit: the `proc_threadSleep` takes `unsigned long long us` as a param, so maybe this is fixed. Please use the expected type directly, don't assume `time_t` would be wide enough.

agkaminski: Hmmm, I think this indeed solves it already - both `*sec` and `sleep` are of `time_t` type (`unsigned long long`).

syscalls.c
agkaminski: Done, thx

syscalls.c
agkaminski: Done

nalajcie: yeah, confused by the types... but `time_t` should probably be not arch-dependant for it to be always true. Still `proc_threadSleep` should probably take `time_t` as a param (or do we have the convention to use `time_t` for absolute times and `unsigned long long` for periods? If so, some types should be changed here ;))

agkaminski: I agree that there is no point for time_t being platform dependent, but this is outside of this PR scope.

Also, it should be shared between kernel and libc and be signed, all these things should be fixed in the future.

nalajcie:
side note: You can write in the commit message body:
Fixes: phoenix-rtos/libphoenix#93

to link commit / PR to issue and auto-close it on merge

agkaminski:
@nalajcie I've applied both remarks regarding commit msg

nalajcie (approved):

PR #339, opened 2022-08-03
Implement thread local storage

lukileczo:
## Description

- Implements thread local storage on all supported architectures.
## Motivation and Context

JIRA: RTOS-209
## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/125.
- [x] Tested by hand on: `ia32-generic-qemu`, `armv7m7-imxrt106x-evk`, `riscv64-generic-qemu`, `armv7a9-zynq7000-qemu`

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/340, https://github.com/phoenix-rtos/phoenix-rtos-build/pull/127, https://github.com/phoenix-rtos/libphoenix/pull/195.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 CMakeLists.txt
 .idea/
 cmake-build-debug/
+.vscode/*
--- a/hal/armv7a/cpu.c
+++ b/hal/armv7a/cpu.c
@@ -19,6 +19,7 @@
 #include "../cpu.h"
 #include "../string.h"
 #include "../spinlock.h"
+#include "../hal.h"
 
 
 /* Function creates new cpu context on top of given thread kernel stack */
@@ -198,6 +199,16 @@ char *hal_cpuFeatures(char *features, unsigned int len)
 		features[0] = '\0';
 
 	return features;
+}
+
+
+void hal_cpuTlsSet(hal_tls_t *tls, cpu_context_t *ctx)
+{
+	/* In theory there should be 8-byte thread control block but
+	 * it's stored elsewhere so we need to subtract 8 from the pointer
+	 */
+	ptr_t ptr = tls->tls_base - 8;
+	__asm__ volatile("mcr p15, 0, %[value], cr13, cr0, 3;" ::[value] "r"(ptr));
 }
 
 
--- a/hal/ia32/cpu.c
+++ b/hal/ia32/cpu.c
@@ -23,6 +23,7 @@
 #include "pci.h"
 #include "ia32.h"
 #include "halsyspage.h"
+#include "../hal.h"
 
 
 struct cpu_feature_t {
@@ -144,7 +145,7 @@ int hal_cpuCreateContext(cpu_context_t **nctx, void *start, void *kstack, size_t
 	ctx->ecx = 0;
 	ctx->ebx = 0;
 	ctx->eax = 0;
-	ctx->gs = ustack ? SEL_UDATA : SEL_KDATA;
+	ctx->gs = SEL_TLS;
 	ctx->fs = ustack ? SEL_UDATA : SEL_KDATA;
 	ctx->es = ustack ? SEL_UDATA : SEL_KDATA;
 	ctx->ds = ustack ? SEL_UDATA : SEL_KDATA;
@@ -315,7 +316,6 @@ void hal_jmp(void *f, void *kstack, void *stack, int argc)
 			movw %%dx, %%ds;\
 			movw %%dx, %%es;\
 			movw %%dx, %%fs;\
-			movw %%dx, %%gs;\
 			pushl %%eax;\
 			iret"
 		:
@@ -416,7 +416,7 @@ void *_cpu_initCore(void)
 
 	hal_memset(&cpu.tss[hal_cpuGetID()], 0, sizeof(tss_t));
 
-	_cpu_gdtInsert(4 + cpu.ncpus, (u32)&cpu.tss[hal_cpuGetID()], sizeof(tss_t), DESCR_TSS);
+	_cpu_gdtInsert(TLS_DESC_IDX + cpu.ncpus, (u32)&cpu.tss[hal_cpuGetID()], sizeof(tss_t), DESCR_TSS);
 
 	cpu.tss[hal_cpuGetID()].ss0 = SEL_KDATA;
 	cpu.tss[hal_cpuGetID()].esp0 = (u32)&cpu.stacks[hal_cpuGetID()][511];
@@ -424,7 +424,7 @@ void *_cpu_initCore(void)
 	/* Set task register */
 	__asm__ volatile(
 		"ltr %0; "
-	:: "r" ((u16)((4 + cpu.ncpus) * 8)));
+	:: "r" ((u16)((TLS_DESC_IDX + cpu.ncpus) * 8)));
 
 	return (void *)cpu.tss[hal_cpuGetID()].esp0;
 }
@@ -437,6 +437,7 @@ static void _hal_cpuInitCores(void)
 	/* Prepare descriptors for user segments */
 	_cpu_gdtInsert(3, 0x00000000, VADDR_KERNEL, DESCR_UCODE);
 	_cpu_gdtInsert(4, 0x00000000, VADDR_KERNEL, DESCR_UDATA);
+	_cpu_gdtInsert(TLS_DESC_IDX, 0x00000000, VADDR_KERNEL, DESCR_TLS);
 
 	/* Initialize BSP */
 	cpu.ncpus = 0;
@@ -615,4 +616,26 @@ void _hal_cpuInit(void)
 //	hal_cpuDebugGuard(1, 2);
 //	hal_cpuDebugGuard(1, 3);
 #endif
+}
+
+
+void hal_cpuTlsSet(hal_tls_t *tls, cpu_context_t *ctx)
+{
+	ptr_t descrh, descrl;
+	u32 type = DESCR_TLS;
+	u32 *gdt;
+	ptr_t base;
+	ptr_t limit = VADDR_KERNEL;
+
+	base = tls->tls_base + tls->tbss_sz + tls->tdata_sz;
+
+	/* Update TLS entry in GDT with allocated page address */
+	descrh = (base & 0xff000000) | (type & 0x00c00000) | (limit & 0x000f0000) |
+		(type & 0x0000ff00) | ((base >> 16) & 0x000000ff);
+	descrl = (base << 16) | (limit & 0xffff);
+
+	gdt = (void *)syspage->hs.gdtr.addr;
+
+	gdt[TLS_DESC_IDX * 2] = descrl;
+	gdt[TLS_DESC_IDX * 2 + 1] = descrh;
 }
--- a/proc/process.c
+++ b/proc/process.c
@@ -288,6 +288,11 @@ int proc_start(void (*initthr)(void *), void *arg, const char *path)
 	process->sigpend = 0;
 	process->sigmask = 0;
 	process->sighandler = NULL;
+	process->tls.tls_base = NULL;
+	process->tls.tbss_sz = 0;
+	process->tls.tdata_sz = 0;
+	process->tls.tls_sz = 0;
+	process->tls.arm_m_tls = NULL;
 
 #ifndef NOMMU
 	process->lazy = 0;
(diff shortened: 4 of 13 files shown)


hal/armv7a/cpu.c
```diff
@@ -19,6 +19,7 @@
 #include "../cpu.h"
 #include "../string.h"
 #include "../spinlock.h"
+#include "../../proc/process.h"
```
agkaminski, line 22: top -> down module referencing (i.e. using proc API in HAL) is forbidden. `tls_t` definition should be located in HAL (and named `hal_tls_t`), this will solve this issue.

.gitignore
```diff
@@ -11,3 +11,4 @@
 CMakeLists.txt
 .idea/
 cmake-build-debug/
+.vscode/*
```
agkaminski, line 14: This change is ok, but should be in a separate commit

hal/armv7a/cpu.c
```diff
@@ -199,3 +200,15 @@ char *hal_cpuFeatures(char *features, unsigned int len)
...
 	return features;
 }
+
+
+void hal_cpuTlsSet(tls_t *tls, cpu_context_t *ctx)
+{
+	/* In theory there should be 8-byte thread control block but
+	 * it's stored elsewhere so we need to subtract 8 from the pointer
+	 */
+	u32 ptr = (u32)tls->tls_base - 8;
```
agkaminski, line 210: Why is it stored in a different place than expected? This might be covering some hidden bug elsewhere.
Also `ptr_t` should be used in place of `u32`.

hal/ia32/cpu.c
```diff
@@ -23,6 +23,8 @@
 #include "pci.h"
 #include "ia32.h"
 #include "halsyspage.h"
+#include "../../lib/printf.h"
```
agkaminski, line 26: Debugging left-over, I guess
```suggestion
```

hal/ia32/cpu.c
```diff
@@ -416,15 +417,14 @@ void *_cpu_initCore(void)
...
+	_cpu_gdtInsert(TLS_DESC_IDX + cpu.ncpus, (u32)&cpu.tss[hal_cpuGetID()], sizeof(tss_t), DESCR_TSS);
 
 	cpu.tss[hal_cpuGetID()].ss0 = SEL_KDATA;
 	cpu.tss[hal_cpuGetID()].esp0 = (u32)&cpu.stacks[hal_cpuGetID()][511];
 
 	/* Set task register */
 	__asm__ volatile(
-		"ltr %0; "
-	:: "r" ((u16)((4 + cpu.ncpus) * 8)));
+		"ltr %0; " ::"r"((u16)((TLS_DESC_IDX + cpu.ncpus) * 8)));
```
agkaminski, line 427: Keep previous formatting

hal/ia32/cpu.c
```diff
@@ -605,3 +606,25 @@ void _hal_cpuInit(void)
 //	hal_cpuDebugGuard(1, 3);
 #endif
 }
+
+
+void hal_cpuTlsSet(tls_t *tls, cpu_context_t *ctx)
+{
+	/* At the end of TLS there must be pointer to itself */
```
agkaminski, line 613: Coding style: in kernel all variables must be declared before logic.

hal/ia32/cpu.c
```diff
@@ -605,3 +606,25 @@ void _hal_cpuInit(void)
...
+	*(ptr_t *)((ptr_t)tls->tls_base + tls->tdata_sz + tls->tbss_sz) = (ptr_t)tls->tls_base + tls->tdata_sz + tls->tbss_sz;
+
+	ctx->gs = SEL_TLS;
+	u32 base = (u32)((char *)tls->tls_base + tls->tbss_sz + tls->tdata_sz);
+	u32 limit = VADDR_KERNEL;
```
agkaminski, line 618: These should be of `ptr_t` type

agkaminski: I think descriptors should be also, I guess these also scale up to 64 bits on x64, but I'm not sure

proc/process.c
```diff
@@ -409,19 +432,49 @@ int process_load32(vm_map_t *map, vm_object_t *o, offs_t base, void *iehdr)
 			hal_memset(vaddr + filesz, 0, round_page((unsigned long)vaddr + memsz) - (unsigned long)vaddr - filesz);
 		}
 	}
+	shdr = (void *)((char *)ehdr + ehdr->e_shoff);
+	shdr += ehdr->e_shstrndx;
+
+	char *snameTab = (char *)ehdr + shdr->sh_offset;
```
agkaminski, line 438: Variables should be declared on the beginning of the function

proc/process.c
```diff
@@ -361,18 +364,38 @@ static void process_illegal(unsigned int n, exc_context_t *ctx)
...
+	process->tls.tls_sz = tls_size(&process->tls);
+	process->tls.arm_m_tls = tls->arm_m_tls;
+}
 
 
+#ifndef NOMMU
+
 /* TODO - adding error handling and unmapping of already mapped segments */
-int process_load32(vm_map_t *map, vm_object_t *o, offs_t base, void *iehdr)
+int process_load32(process_t *process, vm_map_t *map, vm_object_t *o, offs_t base, void *iehdr)
```
agkaminski, line 388: Imho it would be better to pass a pointer to tls here, instead of process. Then we can easily make tls init and assignment code common between load32 and load64

proc/process.c
```diff
@@ -1328,3 +1446,31 @@ int _process_init(vm_map_t *kmap, vm_object_t *kernel)
...
+void tls_init(tls_t *dest, tls_t *source, vm_map_t *map)
+{
+	dest->tdata_sz = source->tdata_sz;
+	dest->tbss_sz = source->tbss_sz;
+	dest->tls_sz = round_page(source->tls_sz);
+	dest->arm_m_tls = source->arm_m_tls;
+
+	dest->tls_base = vm_mmap(map, NULL, NULL, dest->tls_sz, PROT_READ | PROT_WRITE | PROT_USER, NULL, 0, MAP_NONE);
+
+	if (dest->tls_base != NULL) {
```
agkaminski, line 1460: We should return -ENOMEM if it was indeed NULL - we shouldn't "successfully" spawn a thread that expects TLS without TLS

hal/armv7a/cpu.c
lukileczo: Ok, maybe I've misunderstood something, but GCC generates calls for first TLS variable 8 bytes over the address in the register. ARM abi only specifies TLS for Linux.

.gitignore
lukileczo: It is

agkaminski: Oops, sorry!

hal/armv7a/cpu.c
```diff
@@ -199,3 +200,15 @@ char *hal_cpuFeatures(char *features, unsigned int len)
...
+
+void hal_cpuTlsSet(hal_tls_t *tls, cpu_context_t *ctx)
+{
+	/* In theory there should be 8-byte thread control block but
+	 * it's stored elsewhere so we need to subtract 8 from the pointer
+	 */
+	ptr_t ptr = (ptr_t)tls->tls_base - 8;
+	__asm__ volatile("mov r0, %[value];"
+					 "mcr p15, 0, r0, cr13, cr0, 3;" ::[value] "r"(ptr)
+					 : "r0");
```
kemonats, line 213: One `mov` instruction less
```suggestion
	__asm__ volatile("mcr p15, 0, %[value], cr13, cr0, 3;" ::[value] "r"(ptr));
```

proc/process.c
```diff
@@ -361,18 +364,39 @@ static void process_illegal(unsigned int n, exc_context_t *ctx)
 }
 
 
-#ifndef NOMMU
+static void assign_tls(hal_tls_t *process_tls, hal_tls_t *tls, void *tbss_addr)
```
agkaminski, line 367: Use module prefix, on what -> do what nomenclature and camelCase:
`process_tlsAssign`

proc/process.c
```diff
@@ -1328,3 +1450,37 @@ int _process_init(vm_map_t *kmap, vm_object_t *kernel)
 	hal_exceptionsSetHandler(EXC_UNDEFINED, process_illegal);
 	return EOK;
 }
+
+
+int tls_init(hal_tls_t *dest, hal_tls_t *source, vm_map_t *map)
```
agkaminski, line 1455:
```suggestion
int process_tlsInit(hal_tls_t *dest, hal_tls_t *source, vm_map_t *map)
```

proc/process.c
```diff
@@ -809,6 +924,13 @@ static void process_exec(thread_t *current, process_spawn_t *spawn)
 		hal_spinlockClear(&spawn->sl, &sc);
 	}
 
+	if (current->process->tls.tls_base != NULL) {
+		err = tls_init(&current->tls, &current->process->tls, current->process->mapp);
+		if (err == EOK) {
+			hal_cpuTlsSet(&current->tls, current->context);
```
agkaminski, line 930: What will happen if user code that expects TLS to be present gets a process without TLS? I think it will crash or malfunction in a bad way (i.e. variables we expect to be thread safe are not). I think exec should fail if tls_base != NULL and tls_init failed.

proc/process.c
```diff
@@ -1328,3 +1450,37 @@ int _process_init(vm_map_t *kmap, vm_object_t *kernel)
...
+		err = EOK;
+	}
+	else {
+		err = -ENOMEM;
+	}
+	return err;
+}
+
+
+size_t tls_size(hal_tls_t *tls)
```
agkaminski, line 1477: Used in one place, this function is not needed

proc/threads.c
```diff
@@ -787,68 +791,82 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
...
-		return -EINVAL;
+	if (priority >= sizeof(threads_common.ready) / sizeof(thread_t *)) {
+		err = -EINVAL;
+	}
 
-	if ((t = vm_kmalloc(sizeof(thread_t))) == NULL)
-		return -ENOMEM;
+	if ((t = vm_kmalloc(sizeof(thread_t))) == NULL) {
+		err = -ENOMEM;
+	}
```
agkaminski, line 802: This code refactor is out-of-scope of this PR (and not ideal too, if we fail here there is no need for stack allocation below, also we should extract vm_kmalloc call from if to be MISRA friendly).

proc/threads.c
```diff
@@ -787,68 +791,82 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
...
+			hal_cpuSetCtxGot(t->context, process->got);
 
-		hal_spinlockClear(&threads_common.spinlock, &sc);
-		hal_cpuSetCtxGot(t->context, process->got);
-	}
+			if (process->tls.tdata_sz != 0 || process->tls.tbss_sz != 0) {
+				err = tls_init(&t->tls, &process->tls, process->mapp);
+				if (err == EOK) {
+					hal_cpuTlsSet(&t->tls, t->context);
+				}
```
agkaminski, line 854: We should fail if we cannot provide TLS for a thread that expects it

proc/process.c
lukileczo: I think the check `if (err < 0) {` lower, in line 934 covers tls_init failure

agkaminski (approved):
LGTM, let's wait for test in CI

hal/armv7a/cpu.c
```diff
@@ -199,3 +200,14 @@ char *hal_cpuFeatures(char *features, unsigned int len)
...
+
+
+void hal_cpuTlsSet(hal_tls_t *tls, cpu_context_t *ctx)
+{
+	/* In theory there should be 8-byte thread control block but
+	 * it's stored elsewhere so we need to subtract 8 from the pointer
+	 */
+	ptr_t ptr = (ptr_t)tls->tls_base - 8;
+	__asm__ volatile("mcr p15, 0, %[value], cr13, cr0, 3;" ::[value] "r"(ptr)
+					 : "r0");
```
kemonats, line 212: The `r0` is not needded in the clobbers list. You are not using the register `r0` explicitly.
```suggestion
	__asm__ volatile("mcr p15, 0, %[value], cr13, cr0, 3;" ::[value] "r"(ptr));
```

lukileczo: Removed `r0` from clobbers

agkaminski (approved):

PR #344, opened 2022-09-22
stm32l4: Fix two race conditions in low power

agkaminski:
- cmp event could occur before reaching wfi causing wakeup event to be missed and sleep taking much longer (until other wakeup event occured),
- arr event could occur between spinlock set and deep sleep causing timer overflow to be missed (causing ~15 s being lost on system timer).

Remove manual SYSTICK pending flag set, not needed anymore as rescheduling is triggered by timer irqs.

JIRA: ISC-199

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4).

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/stm32/l4/timer.c
+++ b/hal/armv7m/stm32/l4/timer.c
@@ -137,7 +137,9 @@ void timer_jiffiesAdd(time_t t)
 
 void timer_setAlarm(time_t us)
 {
-	u32 setval;
+	const u32 mintime = hal_timerUs2Cyc(1000);
+	u32 setval, timerval[2];
+	int arrpend;
 	spinlock_ctx_t sc;
 	time_t ticks = hal_timerUs2Cyc(us);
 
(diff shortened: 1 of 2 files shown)


hal/armv7m/stm32/l4/timer.c
```diff
@@ -203,13 +231,13 @@ void _hal_timerInit(u32 interval)
...
 
-	*(timer_common.lptim + lptim_cr) |= 4;
-
 	timer_common.overflowh.f = timer_irqHandler;
 	timer_common.overflowh.n = lptim1_irq;
 	timer_common.overflowh.got = NULL;
 	timer_common.overflowh.data = NULL;
 	hal_interruptsSetHandler(&timer_common.overflowh);
 
+	*(timer_common.lptim + lptim_cr) |= 4;
```
HBuczynski, line 240: I assume that this line enables timer. To keep the order of execution, don't we need a memory barrier?

hal/armv7m/stm32/l4/timer.c
```diff
@@ -146,18 +147,43 @@ void timer_setAlarm(time_t us)
...
+		setval = (timerval[1] - 1) & 0xffff;
+	}
+	else {
+		setval = (timerval[1] + (u32)ticks) & 0xffff;
+
+		if (((setval - timerval[1]) & 0xffff) < hal_timerUs2Cyc(1000)) {
+			/* There is too much risk that interrupt will arrive before wfi
+			 * This can happen regardless of previous checks - can be
+			 * caused by a modulo operation few lines above */
+			setval = (timerval[1] + hal_timerUs2Cyc(1000)) & 0xffff;
```
HBuczynski, line 172: You can define value from `hal_timerUs2Cyc(1000)` as a const to not invoke function with the same argument two times.

HBuczynski (commented):
General, the code LGTM, my comments are rather nitpickings. 
I have not dived into the documentation of this controller so I have only focused on the code structure.

hal/armv7m/stm32/l4/timer.c
```diff
@@ -146,18 +147,43 @@ void timer_setAlarm(time_t us)
 	}
 
 	hal_spinlockSet(&timer_common.sp, &sc);
-	setval = (timer_getCnt() + (u32)ticks) & 0xffff;
+
+	/* Check if there's pending overflow to be handled
+	 * to not lose it should there be next when sleeping */
+	timerval[0] = timer_getCnt();
+	arrpend = (*(timer_common.lptim + lptim_isr) & 2) != 0 ? 1 : 0;
```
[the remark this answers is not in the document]
agkaminski: On read I don't need to

hal/armv7m/stm32/l4/timer.c
agkaminski: Not strictly needed here, but won't hurt, added

HBuczynski (approved):

anglov (approved):

PR #349, opened 2022-11-17
add armv8m architecture port

damianloew:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


hal/armv8m/arch/pmap.h
```diff
@@ -0,0 +1,70 @@
...
+} pmap_t;
+
+
+typedef struct _mpur_t {
+	u8 region;
+	u32 base;
+	u32 size;
+	u8 subregions;
+	int attr;
+} mpur_t;
```
agkaminski, line 66: Hmm, is it the same stuff as on armv7m? We can skip this stuff for now, it's not used for now on armv7m

hal/armv8m/arch/cpu.h
```diff
@@ -0,0 +1,233 @@
...
+
+#if defined(CPU_NRF9160)
+#define CPU_NRF91
+#endif
+
+#include "types.h"
+
+#if defined CPU_NRF91
+#define SIZE_PAGE 0x200
+#endif
```
agkaminski, line 28: Shouldn't be needed, will cause SIZE_PAGE to be not defined in `#else `case
```suggestion
#define SIZE_PAGE 0x200
```

agkaminski (commented):
Hmm, `nrf91` specific files seem to be missing in this PR

hal/armv8m/arch/pmap.h
damianloew: Yes, that's right it's not necessary to have mpur_t defined, removed.

damianloew:
> Hmm, `nrf91` specific files seem to be missing in this PR

Yeah, as consulted with @HBuczynski I've decided to split nrf9160 kernel port to 2 PR's - one for the initial armv8m file structure (things depending only on armv8m architecture) and next will be nrf9160 dependent code, there are files to add in the next PR:
Screenshot from 2022-11-25 18-33-35

Of course, in this PR are also some defines etc. needed for nrf target, but I hope it isn't a problem.

hal/armv8m/nrf/91/nrf91.c
```diff
@@ -0,0 +1,274 @@
...
+
+void _nrf91_nvicSystemReset(void)
+{
+	*(nrf91_common.scb + scb_aircr) = ((0x5fau << 16) | (*(nrf91_common.scb + scb_aircr) & (0x700u)) | (1u << 2));
+
+	__asm__ volatile("dsb");
+
+	for (;;)
+		;
+}
```
HBuczynski, line 221: `nvic` controller should be moved to the interrupt unit

hal/armv8m/nrf/91/nrf91.c
```diff
@@ -0,0 +1,274 @@
...
+	/* We set only group priority field */
+	*ptr = (priority << (8 * (irqn % 4) + 4));
+}
+
+
+void _nrf91_nvicSystemReset(void)
+{
+	*(nrf91_common.scb + scb_aircr) = ((0x5fau << 16) | (*(nrf91_common.scb + scb_aircr) & (0x700u)) | (1u << 2));
+
+	__asm__ volatile("dsb");
```
HBuczynski, line 217: Be consistent with the rest of the code. Please use `hal_cpuDataSyncBarrier`

include/arch/nrf9160.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+#ifndef _PHOENIX_ARCH_NRF9160_H_
+#define _PHOENIX_ARCH_NRF9160_H_
+
+
+/* clang-format off */
+enum { gpio_input = 0, gpio_output };
+enum { gpio_low = 0, gpio_high };
+enum { gpio_nopull = 0, gpio_pulldown, gpio_pullup = 3};
```
HBuczynski, line 23: These enums should not be in this header.

include/arch/nrf9160.h
```diff
@@ -0,0 +1,36 @@
...
+
+
+/* nRF9160 peripheral id's - same as irq numbers */
+enum { spu = 3, regulators, clock = 5, power = 5, ctrlapperi, spi0 = 8, twi0 = 8, uarte0 = 8,
+spi1 = 9, twi1 = 9, uarte1 = 9, spi2 = 10, twi2 = 10, uarte2 = 10, spi3 = 11, twi3 = 11, uarte3 = 11,
+gpiote0 = 13, saadc, timer0, timer1, timer2, rtc0 = 20, rtc1, ddpic = 23, wdt, 
+egu0 = 27, egu1, egu2, egu3, egu4, egu5, pwm0, pwm1, pwm2, pwm3, pdm = 38, 
+i2s = 40, ipc = 42, fpu = 44, gpiote1 = 49, kmu = 57, nvmc = 57, vmc, cc_host_rgf = 64,
+cryptocell = 64, gpio = 66 };
```
HBuczynski, line 32: Please add  postfix `*_irq` to each of the enum.

HBuczynski (changes requested):

hal/armv8m/exceptions.c
```diff
@@ -0,0 +1,116 @@
...
+	size_t i = 0;
+
+	n &= 0xf;
+
+	hal_strcpy(buff, "\nException: ");
+	hal_strcpy(buff += hal_strlen(buff), mnemonics[n]);
+	hal_strcpy(buff += hal_strlen(buff), "\n");
+	buff += hal_strlen(buff);
+
+	i += hal_i2s(" r0=", &buff[i], ctx->r0, 16, 1);
```
kemonats, line 40:
```suggestion
	hal_strcpy(buff, "\nException: ");
	i = sizeof("\nException: ") - 1;
	hal_strcpy(&buff[i], mnemonics[n]);
	i += hal_strlen(mnemonics[n]);

	i += hal_i2s("\n r0=", &buff[i], ctx->r0, 16, 1);
```
`hal_strlen()` only on `mnemonics[n]`, without 3 times on  "\nExceptions: "

General comment.
Since the `hal_i2s()` function is called with afixed 4th and 5th argument, you can reduce code size by defining, for example
```
hal_i2sHex(const char *prefix, char *buff, unsigned long i);
```
and use it.

You can define this function as:
```
size_t hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
```
and eliminate `hal_strlen()` in this code:
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/428d509d25d24967a4d88235effe0d7390a6677a/proc/process.c#L324-L330

hal/armv8m/exceptions.c
```diff
@@ -0,0 +1,116 @@
...
+
+	buff[i++] = '\n';
+
+	buff[i] = 0;
+}
+
+
+void exceptions_dispatch(unsigned int n, exc_context_t *ctx)
+{
+	char buff[512];
```
kemonats, line 75:
```suggestion
	char buff[SIZE_CTXDUMP];
```

hal/armv8m/spinlock.c
```diff
@@ -0,0 +1,120 @@
...
+		ldrexb r1, [%1]; \
+		cmp r1, #0; \
+		beq 1b; \
+		strexb r1, r2, [%1]; \
+		cmp r1, #0; \
+		bne 1b; \
+		dmb"
+	:
+	: "r" (sc), "r" (&spinlock->lock)
+	: "r1", "r2", "memory", "cc");
```
kemonats, line 42:
```suggestion
		mrs r2, primask; \
		cpsid i; \
		str r2, [%0]; \
		mov r3, #0; \
	1: \
		ldrexb r2, [%1]; \
		cmp r2, #0; \
		beq 1b; \
		strexb r2, r3, [%1]; \
		cmp r2, #0; \
		bne 1b; \
		dmb"
	:
	: "r" (sc), "r" (&spinlock->lock)
	: "r2", "r3", "memory", "cc");
```

hal/armv8m/spinlock.c
```diff
@@ -0,0 +1,120 @@
...
+		add r1, r1, #1; \
+		dmb; \
+		strexb r2, r1, [%0]; \
+		cmp r2, #0; \
+		bne 1b; \
+		ldr r1, [%1]; \
+		msr primask, r1;"
+	:
+	: "r" (&spinlock->lock), "r" (sc)
+	: "r1", "r2", "memory");
```
kemonats, line 62:
```suggestion
		ldrexb r2, [%0]; \
		add r2, r2, #1; \
		dmb; \
		strexb r3, r2, [%0]; \
		cmp r3, #0; \
		bne 1b; \
		ldr r2, [%1]; \
		msr primask, r2;"
	:
	: "r" (&spinlock->lock), "r" (sc)
	: "r2", "r3", "memory", "cc");
```

hal/armv8m/spinlock.c
```diff
@@ -0,0 +1,120 @@
...
+	/* clang-format on */
+}
+
+
+void _hal_spinlockCreate(spinlock_t *spinlock, const char *name)
```
kemonats, line 67:
```suggestion
static void _hal_spinlockCreate(spinlock_t *spinlock, const char *name)
```

hal/armv8m/string.c
```diff
@@ -0,0 +1,256 @@
...
+		ldrb r1, [%1, %0]; \
+		cbz r1, 2f; \
+		add %0, #1; \
+		b 1b; \
+	2:"
+	: "+r" (k), "+r" (s)
+	:
+	: "r1", "memory", "cc");
+
+	return k;
```
kemonats, line 123: This function can be improved, i.e., one instruction can be reduced in the loop.
If you are interested I can present it.

hal/armv8m/string.c
```diff
@@ -0,0 +1,256 @@
...
+	char c;
+	unsigned long l, k, m;
+
+	m = hal_strlen(prefix);
+	hal_memcpy(s, prefix, m);
+
+	for (k = m, l = (unsigned long)-1; l; i /= b, l /= b) {
+		if (!zero && !i) {
+			break;
+		}
```
kemonats, line 243:
```suggestion
		if (!zero && !i) {
			if (k == m) {
				s[k++] = '0';
			}
			break;
		}
```
Look at PR #231

hal/armv8m/nrf/91/console.c
```diff
@@ -0,0 +1,182 @@
...
+	blocks = len / TX_DMA_SIZE;
+
+	for (iter = 0; iter <= blocks; iter++) {
+		cnt = 0;
+		do {
+			pos = (iter * TX_DMA_SIZE) + cnt;
+			tx_dma_buff[cnt] = s[pos];
+			cnt++;
+			pos++;
+		} while ((s[pos - 1] != '\0') && (cnt < console_common.txDmaSize));
```
kemonats, line 76:
```suggestion
	len = hal_strlen(s);
	blocks = len /  console_common.txDmaSize;

	p = 0;
	for (iter = 0; iter <= blocks; iter++) {
		cnt = 0;
		while (pos < len) && (cnt < console_common.txDmaSize)) {
			tx_dma_buff[cnt++] = s[pos++];
		}
		if (cnt == 0) {
			break;
		}
```
You send  '\0' char.

hal/armv8m/cpu.c
```diff
@@ -0,0 +1,231 @@
...
+
+	if (((cpuinfo >> 24) & 0xffu) == 0x41u) {
+		hal_strcpy(info + i, "ARMv8 ");
+		i += 6;
+	}
+
+	if (((cpuinfo >> 4) & 0xfffu) == 0xc23u) {
+		hal_strcpy(info + i, "Cortex-M33 ");
+		i += 10;
+	}
```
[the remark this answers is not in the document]
damianloew: Yeah, thanks for being watchful!

hal/armv8m/exceptions.c
damianloew: Thanks! I think changing `hal_exceptionsDumpContext()` definition can be very good idea, but maybe for another PR.

hal/armv8m/spinlock.c
damianloew: Why do we want to avoid using r1?

damianloew: Now I am thinking whether "cc" is even  needed in clobbers

hal/armv8m/string.c
damianloew: I would like to add it there after #231 merge, but thanks for reminder

hal/armv8m/string.c
damianloew: What do you mean?

hal/armv8m/nrf/91/console.c
damianloew: I propose sth like this then:

```C
	len = hal_strlen(s);
	blocks = len / console_common.txDmaSize;

	for (iter = 0; iter <= blocks; iter++) {
		cnt = 0;
		do {
			pos = (iter * TX_DMA_SIZE) + cnt;
			tx_dma_buff[cnt++] = s[pos++];
		} while ((s[pos] != '\0') && (cnt < console_common.txDmaSize));
```

damianloew:
Changes regarding last suggestions applied in the force-push.

hal/armv8m/nrf/91/console.c
kemonats: Two comments:
- what if `len == 0` (empty string),
- in the loop `do .. while` you calculate the `pos` variable and then increase it by one, and so on in the next steeps. This is inefficient.

damianloew: You are right, I didn't see this issue before, thanks!

hal/armv8m/Makefile
```diff
@@ -0,0 +1,13 @@
+#
+# Makefile for Phoenix-RTOS kernel (ARMv8M HAL)
+#
+# Copyright 2016-2017, 2019-2020, 2022 Phoenix Systems
```
HBuczynski, line 4:
```suggestion
# Copyright 2022 Phoenix Systems
```

hal/armv8m/exceptions.c
HBuczynski: Please update

hal/armv8m/nrf/interrupts.c
```diff
@@ -0,0 +1,228 @@
...
+			t->next->prev = t->prev; \
+			if (t == (*list)) \
+				(*list) = t->next; \
+		} \
+		t->next = NULL; \
+		t->prev = NULL; \
+	} while (0)
+
+
+struct {
```
HBuczynski, line 72:
```suggestion
static struct {
```

hal/armv8m/spinlock.c
```diff
@@ -0,0 +1,120 @@
...
+ * Author: Pawel Pisarczyk, Damian Loewnau
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "../spinlock.h"
+
+struct {
```
HBuczynski, line 18:
```suggestion
static struct {
```

include/arch/nrf9160.h
HBuczynski: I see that you like the comment but `enums` are still in this header.

damianloew: Yeah, sorry. Now it is applied.

hal/armv8m/exceptions.c
damianloew: Thanks, now it is applied.

HBuczynski (approved):
LGTM

hal/armv8m/cpu.c
```diff
@@ -0,0 +1,237 @@
...
```
kemonats, line 172:
```suggestion
	if (((cpuinfo >> 4) & 0xfffu) == 0xD21u) {
		hal_strcpy(info + i, "Cortex-M33 ");
		i += 11;
	}
```
Update at least this.

damianloew: Oh, now I see that remote branch hasn't been updated with changes from your review @kemonats. Now they are present, thanks!

hal/armv8m/nrf/91/nrf91.c
```diff
@@ -0,0 +1,224 @@
...
+/* clang-format on */
+
+
+/* platformctl syscall */
+
+
+/* TODO: add platformctl implementation */
+int hal_platformctl(void *ptr)
+{
+	return 0;
```
agkaminski, line 56:
```suggestion
	return -ENOSYS;
```

hal/armv8m/nrf/91/nrf91.c
```diff
@@ -0,0 +1,224 @@
...
+int _nrf91_gpioConfig(u8 pin, u8 dir, u8 pull)
+{
+	if (pin > 31) {
+		return -1;
+	}
+
+	if (dir == gpio_output) {
+		*(nrf91_common.gpio + gpio_dirsetout) = (1u << pin);
+		while ((*(nrf91_common.gpio + gpio_dir) & (1u << pin)) != (1u << pin)) {
+		}
```
agkaminski, line 96: Wouldn't memory barier (`dmb` instruction) suffice? I don't really know, but I suspect so. Applies to other instances.

agkaminski: If no, then maybe consider using static inline function to perform write, read back and compare operation.

hal/armv8m/nrf/interrupts.c
```diff
@@ -0,0 +1,228 @@
...
+
+
+void _interrupts_nvicSystemReset(void)
+{
+	*(interrupts.scb + scb_aircr) = ((0x5fau << 16) | (*(interrupts.scb + scb_aircr) & (0x700u)) | (1u << 2));
+
+	hal_cpuDataSyncBarrier();
+
+	for (;;)
+		;
```
agkaminski, line 118:
```suggestion
	for (;;) {
	}
```

hal/armv8m/pmap.c
```diff
@@ -0,0 +1,103 @@
...
+
+int _pmap_kernelSpaceExpand(pmap_t *pmap, void **start, void *end, page_t *dp)
+{
+	return 0;
+}
+
+
+int pmap_segment(unsigned int i, void **vaddr, size_t *size, int *prot, void **top)
+{
+	if (i) {
```
agkaminski, line 82:
```suggestion
	if (i != 0) {
```

agkaminski (commented):
Please also take a look at https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/363 (after it has been approved of course)

hal/armv8m/nrf/91/console.c
```diff
@@ -0,0 +1,181 @@
...
+	_hal_consolePrint(s);
+	_hal_consolePrint(CONSOLE_NORMAL);
+}
+
+
+void hal_consolePutch(const char c)
+{
+	volatile char *tx_dma_buff = (volatile char *)console_common.txDma;
+
+	tx_dma_buff[0] = c;
```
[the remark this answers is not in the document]
damianloew: Please check the above answer.

hal/armv8m/nrf/91/console.c
```diff
@@ -0,0 +1,181 @@
...
+
+	len = hal_strlen(s);
+	blocks = len / TX_DMA_SIZE;
+
+	pos = 0;
+	for (iter = 0; iter <= blocks; iter++) {
+		cnt = 0;
+		while ((s[pos] != '\0') && (cnt < console_common.txDmaSize)) {
+			tx_dma_buff[cnt++] = s[pos++];
+		}
```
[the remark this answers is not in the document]
damianloew: I've verified that this solution with using s is not working

hal/armv8m/nrf/91/nrf91.c
damianloew: In this case `dmb` is sufficient, but when checking events I'd rather leave it as it was:
```C
		while (*(console_common.base + uarte_events_txstarted) != 1u) {
		}
```

agkaminski: Yeah, I was referring only to `x = a;` `while (x != a);`

hal/armv8m/nrf/91/console.c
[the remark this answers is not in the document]
agkaminski: Ok, that's a shame though, we could save both buffer space on copy operation :(

I don't quite understand though why `console_common.txDma` is working. There should be nothing special about this memory in .bss.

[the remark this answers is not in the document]
damianloew: What do you mean? Why bss is stored above `0x2000 0000` (RAM) and strings not?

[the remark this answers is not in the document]
nalajcie: from what I understood the string originates from `.rodata` in that case and resides on flash. the EasyDMA can't access directly the FLASH memory so we need to copy it to ram (https://infocenter.nordicsemi.com/pdf/nRF9160_PS_v2.0.pdf page 22)

Please note that on Zephyr the copying is done **only if the original string is not in RAM** so that would be the optimal solution imho.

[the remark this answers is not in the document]
damianloew: Yeah, that's right. I proposed sth similar in the last force push.

damianloew (commented):
Last two force pushes contain:

 - rebase to the master branch
 - tls support (analogous to armv7m)
 - hal_cpuLowPower() updated to the version consistent with other platforms

hal/armv8m/exceptions.c
```diff
@@ -0,0 +1,116 @@
...
+
+	i += hal_i2s("\nr12=", &buff[i], ctx->r12, 16, 1);
+	i += hal_i2s("  sp=", &buff[i], (u32)ctx - 17 * 4, 16, 1);
+	i += hal_i2s("  lr=", &buff[i], ctx->lr, 16, 1);
+	i += hal_i2s("  pc=", &buff[i], ctx->pc, 16, 1);
+
+	i += hal_i2s("\npsp_s=", &buff[i], ctx->psp_s, 16, 1);
+	i += hal_i2s(" psr=", &buff[i], ctx->psr, 16, 1);
+	i += hal_i2s(" exr=", &buff[i], ctx->excret, 16, 1);
+	i += hal_i2s(" bfa=", &buff[i], *(u32 *)0xe000ed38, 16, 1);
```
anglov, line 63: Please look at https://github.com/phoenix-rtos/phoenix-rtos-kernel/commit/c6f7d6c7355df92810e1116eaa93491954f94425 and https://github.com/phoenix-rtos/phoenix-rtos-kernel/commit/4c6ad6160c37bd1193c4357b3d43c160f8b7aad3

anglov (commented):
I don't have that target so not checking if it's work. Also don't looking on timer/uart/gpio parts of commit. I can confirm that common archiecture part  works.

I point some inconsistencies with armv7m targets, mainly things that was changed in armv7m during development of that port.

hal/armv8m/nrf/_init.S
```diff
@@ -0,0 +1,315 @@
...
+.type _hal_invokePendSV, %function
+_hal_invokePendSV:
+	mov r1, #(1 << 28)
+	ldr r2, =ADDR_SCB
+	str r1, [r2, #4]
+	bx lr
+.size _hal_invokePendSV, .-_hal_invokePendSV
+.ltorg
+
+/* analyze this function! */
```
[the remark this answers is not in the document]
damianloew: Yeah, I've missed them, thanks!

hal/armv8m/nrf/91/nrf91.c
```diff
@@ -0,0 +1,218 @@
...
+
+	/* Enable low power mode */
+	*(nrf91_common.power + power_tasks_lowpwr) = 1u;
+	hal_cpuDataMemoryBarrier();
+
+	/* Disable all power interrupts */
+	*(nrf91_common.power + power_intenclr) = 0x64u;
+
+	/* Disable all clock interrupts */
+	*(nrf91_common.power + power_intenclr) = 0x3u;
```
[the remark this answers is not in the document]
damianloew: Yeah, thanks!

damianloew:
> I don't have that target so not checking if it's work. Also don't looking on timer/uart/gpio parts of commit. I can confirm that common archiecture part works.
> 
> I point some inconsistencies with armv7m targets, mainly things that was changed in armv7m during development of that port.

Thanks for the review, all the changes have been applied - small corrections have been amended to a first commit, updates since last version are placed in a second one. Above that I have looked through last changes in kernel and haven't discovered anything else.

hal/armv8m/exceptions.c
```diff
@@ -0,0 +1,138 @@
...
+	i += hal_i2s("  r9=", &buff[i], ctx->r9, 16, 1);
+	i += hal_i2s(" r10=", &buff[i], ctx->r10, 16, 1);
+	i += hal_i2s(" r11=", &buff[i], ctx->r11, 16, 1);
+
+	i += hal_i2s("\nr12=", &buff[i], hwctx->r12, 16, 1);
+	i += hal_i2s(" psr=", &buff[i], hwctx->psr, 16, 1);
+	i += hal_i2s("  lr=", &buff[i], hwctx->lr, 16, 1);
+	i += hal_i2s("  pc=", &buff[i], hwctx->pc, 16, 1);
+
+	i += hal_i2s("\npsp_s=", &buff[i], psp_s, 16, 1);
```
anglov, line 73: Should it  be named with _s suffix? At least for now we have only non-secure mode, so according to DDI0553B.r B3.8 it should be named without _s/_ns suffix. And you don't add this suffix to msp.

hal/armv8m/exceptions.c
```diff
@@ -0,0 +1,138 @@
...
+		"8 #", "9 #", "10 #", "11 #SVC",
+		"12 #Debug", "13 #", "14 #PendSV", "15 #SysTick"
+	};
+	size_t i = 0;
+	u32 msp = (u32)ctx + sizeof(*ctx);
+	u32 psp_s = ctx->psp_s;
+	cpu_hwContext_t *hwctx;
+
+	/* If we came from userspace HW ctx in on psp stack */
+	if (ctx->excret == RET_THREAD_PSP) {
```
anglov, line 37: Not valid on armv8m, causing that userspace exceptions doesn't work correctly. 
I think that you should only check on bit 2 of EXC_RET to determine which stack have hardware context. 

DDI0553B.r D1.2.95 EXC_RETURN, Exception Return Payload:
```
SPSEL, bit [2]
Stack pointer selection. The value of this bit indicates the transitory value of the CONTROL.SPSEL bit
associated with the Security state of the exception as indicated by EXC_RETURN.ES.
The possible values of this bit are:
0
Main stack pointer.
1
Process stack pointer.

```

It's differ from armv7m mainly because of additional security features (trustzone etc.) I would advise to check other usage of that defines because it indicates secure mode

hal/armv8m/arch/cpu.h
```diff
@@ -0,0 +1,240 @@
...
+{
+	__asm__ volatile("cpsie if");
+}
+
+
+static inline void hal_cpuHalt(void)
+{
+	__asm__ volatile("\
+	wfi; \
+	nop; ");
```
anglov, line 113: clang-format misformated it.

hal/armv8m/exceptions.c
damianloew: Yeah, maybe it would be better, I wanted to show that there is only secure mode in our implementation, but let's be consistent with the docs.

anglov: Welp, It bring my attention mainly because it causes that exception will be misformated.

hal/armv8m/nrf/91/nrf91.h
```diff
@@ -0,0 +1,69 @@
...
+/* clang-format on */
+
+
+extern void _nrf91_platformInit(void);
+
+
+extern int hal_platformctl(void *);
+
+
+extern void _nrf91_platformInit(void);
```
lukileczo, line 42: Doubled function declaration

hal/armv8m/nrf/91/nrf91.c
```diff
@@ -0,0 +1,218 @@
...
+
+
+/* TODO: add platformctl implementation */
+int hal_platformctl(void *ptr)
+{
+	return -ENOSYS;
+}
+
+
+void _nrf91_platformInit(void)
```
lukileczo, line 60: I'd name this function `_hal_platformInit`, like on `armv7a` and `sparc` targets. This way there'll be no need for an `#ifdef` in `hal_cpuInit`.

hal/armv8m/exceptions.c
anglov: After discussion we have conclusions:
- on nrf9160 default execution state is secure state, so EXC_RET is set correctly
- it was still worth to switch to checking on bit 2  in case of exceptions - it's well defined in cpu spec. 

So current version is ok for me and it's proved to work.

hal/armv8m/nrf/interrupts.c
```diff
@@ -0,0 +1,228 @@
...
+	interrupts.counters[n]++;
+
+	if ((h = interrupts.handlers[n]) != NULL) {
+		do {
+			hal_cpuSetGot(h->got);
+			if (h->f(n, ctx, h->data)) {
+				reschedule = 1;
+			}
+		} while ((h = h->next) != interrupts.handlers[n]);
+	}
```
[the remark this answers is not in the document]
damianloew: Yeah, MISRA, thanks!

hal/armv8m/nrf/91/nrf91.c
damianloew: Applied, this and the other suggestions. Moreover, I've looked through the whole my code in kernel once again.

gerard5 (commented):
The code that supports the early console can be optimized by decomposing easydma parts, making the implementation more concise.

hal/armv8m/nrf/91/console.c
```diff
@@ -0,0 +1,198 @@
...
+		*(console_common.base + uarte_txd_ptr) = (volatile u32)s;
+		*(console_common.base + uarte_txd_maxcnt) = cnt;
+		*(console_common.base + uarte_starttx) = 1u;
+		while (*(console_common.base + uarte_events_txstarted) != 1u) {
+		}
+		*(console_common.base + uarte_events_txstarted) = 0u;
+
+		while (*(console_common.base + uarte_events_endtx) != 1u) {
+		}
+		*(console_common.base + uarte_events_endtx) = 0u;
```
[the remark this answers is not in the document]
damianloew: Thanks, I wanted to minimize the number of function calls in this module, but I think using such function may be a good idea. Applied with name changed a little bit.

hal/armv8m/nrf/91/console.c
```diff
@@ -0,0 +1,184 @@
...
+	volatile char *tx_dma_buff = (volatile char *)console_common.txDma;
+	int iter, blocks, pos = 0, len = 0;
+
+	while ((s[len] != '\0')) {
+		len++;
+	}
+
+	/* copy to RAM only if it's not there. (6.7.7. EasyDMA chapter in NRF9160 PS) */
+	if (((u32)s & 0xE0000000u) == 0x20000000u) {
+		console_dmaSend(s, len);
```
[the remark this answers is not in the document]
damianloew: Yeah, I'd leave discarding const intentionally.

hal/armv8m/nrf/91/console.c
```diff
@@ -0,0 +1,184 @@
...
+}
+
+
+void hal_consolePutch(const char c)
+{
+	volatile char *tx_dma_buff = (volatile char *)console_common.txDma;
+
+	tx_dma_buff[0] = c;
+
+	console_dmaSend(console_common.txDma, 1);
```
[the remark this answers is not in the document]
damianloew: As we discussed it's not necessary to have this pointer volatile here.

hal/armv8m/nrf/91/console.c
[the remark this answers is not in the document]
damianloew: The suggested approach with union seems to be the best solution here.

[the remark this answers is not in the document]
gerard5: Ok, for me.

damianloew:
> I see some weakness of the current approach, functions `_hal_consolePrint` and `hal_consolePutch` share the same common buffer `console_common.txDma[64]`, this approach is ok in `plo` where there is no multi-threading but in kernel is problematic, because there can be many writers to common buffer (txDma) so a locking is needed while rodata is copied to RAM, and if dma use that `txDma` buffer while it is transmitting, all the more necessary.

Thanks for the suggestions. As mentioned in our coding convention functions prefixed with `_` don't have to be thread safe, so I'd leave it there. However, in the rest of functions I've added locks.

hal/armv8m/Makefile
```diff
@@ -0,0 +1,13 @@
+#
+# Makefile for Phoenix-RTOS kernel (ARMv8M HAL)
+#
+# Copyright 2022 Phoenix Systems
+#
+
+
+ifneq (, $(findstring nrf, $(TARGET_SUBFAMILY)))
+    include hal/armv8m/nrf/Makefile
+    CFLAGS += -Ihal/armv8m/nrf -Ihal/armv8m
```
agkaminski, line 10: `-Ihal/armv8m` imho should be common, not only for `nrf`

hal/armv8m/nrf/91/nrf91.c
```diff
@@ -0,0 +1,218 @@
...
+{
+	return -ENOSYS;
+}
+
+
+void _hal_platformInit(void)
+{
+	hal_spinlockCreate(&nrf91_common.pltctlSp, "pltctl");
+}
+/* SysTick */
```
agkaminski, line 64:
```suggestion


/* SysTick */
```

hal/armv8m/nrf/91/nrf91.c
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	return 0;
+}
+
+
+int _nrf91_gpioSet(u8 pin, u8 val)
+{
+	if (pin > 31)
+		return -1;
```
agkaminski, line 115:
```suggestion
	if (pin > 31) {
		return -1;
	}
```

agkaminski (commented):
Small nitpicks, otherwise LGTM

damianloew:
> if (pin > 31)

Thanks, applied.

PR #360, opened 2022-12-21
klog: Separate log_write from readers wakeup

agkaminski:
This allows log to be written to in any place, as there are no longer need to take threads_common.spinlock on write. Readers wakeup operation has been moved to:
- after write to klog from user space,
- just before exit from syscall,
- idle thread.

JIRA: RTOS-309

## Motivation and Context

To fix a deadlock on threads_common.spinlock when attempting to use lib_printf from e.g. scheduler.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt-1064).

## Checklist:

## Special treatment

### DIFF
--- a/log/log.c
+++ b/log/log.c
@@ -55,6 +55,7 @@ static struct {
 	offs_t tail;
 	lock_t lock;
 	log_reader_t *readers;
+	volatile int updated;
 } log_common;
 
 static int _log_empty(void)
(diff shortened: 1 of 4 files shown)


agkaminski:
BTW, this does not fix stack canary - we wanted to stop the system, because stack overflow + stack hungry lib_printf causes general instability, but it's not possible right now - as we need to allow userspace thread to print this info to the console.

agkaminski:
Now stack canary should be fixed too

Maxez (approved):
LGTM

PR #362, opened 2022-12-21
schedule: Allow stack canary enable independent of NDEBUG

agkaminski:
DONE: RTOS-310

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

  - https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/360

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,9 @@ CFLAGS += $(BOARD_CONFIG)
 CFLAGS += -I. -I$(PROJECT_PATH)/
 CFLAGS += -DVERSION=\"$(VERSION)\"
 
+# uncomment to enable stack canary checking
+# CFLAGS += -DSTACK_CANARY
+
 EXTERNAL_HEADERS_DIR := ./include
 EXTERNAL_HEADERS := $(shell find $(EXTERNAL_HEADERS_DIR) -name \*.h)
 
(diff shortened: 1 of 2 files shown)


nalajcie:
not sure how should I enable the feature in a kosher way. I can't override CLFAGS from command-line, so I need to dig into Makefiles to do so..., if so - please at least add in kernel Makefile:
```make
# uncomment to enable stack canary checking
# CFLAGS += -DSTACK_CANARY
```

IMHO, it would be better to do it automatically based on some ENV variable, so no need to manually edit makes, eg:
```make
## compilation options (specify them in project file / env)
ifneq ($(KERNEL_STACK_CANARY),)
  CFLAGS += -DSTACK_CANARY
endif
```

I would appreciate specific documentation page about debugging tips for our OS (not right away, but IMHO much needed :)).

agkaminski:
@nalajcie imho best way would be to add setting -DSTACK_CANARY to phoenix-rtos-build based on ENV

nalajcie:
> @nalajcie imho best way would be to add setting -DSTACK_CANARY to phoenix-rtos-build based on ENV

The practice is to:
 - do so on "module level" not to pollute build environment of other unrelated components
 - use unique prefixes for all possible DEFINE's (probably module prefixes) to avoid name clashing
 - generating global `config.h` (nice, because incremental build works on env changes then)

Our preference would probably be the first for now.

agkaminski:
Ok

nalajcie (approved):
would prefer enabling by env but still better than nothing :)

agkaminski:
You said you prefer option 1 xP

nalajcie:
image

Precision of language :P.

Not a problem, I would change it when it would start to bother me too much ;).

PR #367, opened 2023-02-02
zynq7000: Increase system timer resolution

agkaminski:
JIRA: RTOS-347

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/zynq7000/timer.c
+++ b/hal/armv7a/zynq7000/timer.c
@@ -5,8 +5,8 @@
  *
  * System timer driver
  *
- * Copyright 2021 Phoenix Systems
- * Author: Hubert Buczynski
+ * Copyright 2021, 2023 Phoenix Systems
+ * Author: Hubert Buczynski, Aleksander Kaminski
  *
  * This file is part of Phoenix-RTOS.
  *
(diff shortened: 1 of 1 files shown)


agkaminski:
@kemonats Good catch, fixed, thanks

kemonats:
The conversion of ticks to microseconds could be improved a bit, but that's probably for another PR.

niewim19 (approved):
Not going into implementation details: works on zturn board, no drawbacks strange behavior  was observed.

hal/armv7a/zynq7000/timer.c
```diff
@@ -63,20 +61,32 @@ static int _timer_irqHandler(unsigned int n, cpu_context_t *ctx, void *arg)
...
+
+	/* Check if there's pending jiffies increment */
+	if ((*(timer_common.ttc + isr) & 1) != 0) {
+		/* ISR register is clear on read, we have to update jiffies now */
+		timer_common.jiffies += timer_common.ticksPerFreq;
+
+		/* Timer might've just wrapped-around,
+		 * take counter value again */
+		jiffies = timer_common.jiffies;
+		cnt = (time_t)(*(timer_common.ttc + cnt_value));
```
Maxez, line 85: Why duplicate `cnt` and `jiffies` readout? Read their values after pending interrupt check (remove lines 82-85 and move lines 74-75 after the check).

agkaminski: Because timer might've overflowed exactly between cnt readout and isr check (or might've not, we don't know). I re-read jiffies instead of incrementing mostly because of personal taste.

Maxez: Ack

Maxez (approved):

PR #370, opened 2023-02-17
Create port for Leon3 GR716

lukileczo:
## Description

Add port for Leon 3 GR716 uC
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


hal/sparcv8leon3/interrupts.c
```diff
@@ -0,0 +1,221 @@
...
+	int i;
+
+	for (i = 0; i < SIZE_INTERRUPTS; ++i) {
+		hal_spinlockCreate(&interrupts_common.spinlocks[i], "interrupts_common");
+		interrupts_common.handlers[i] = NULL;
+		interrupts_common.counters[i] = 0;
+	}
+
+	interrupts_common.int_ctrl = (void *)INT_CTRL_BASE;
+	// hal_cpuEnableInterrupts();
```
agkaminski, line 220: I don't think we want to this here anyway 
```suggestion
```

hal/sparcv8leon3/gr716/gr716.c
```diff
@@ -0,0 +1,325 @@
...
+
+void _gr716_cguClkDisable(u32 cgu, u32 device)
+{
+	volatile u32 *cgu_base = (cgu == cgu_primary) ? gr716_common.cgu_base0 : gr716_common.cgu_base1;
+	u32 msk = 1 << device;
+
+	*(cgu_base + cgu_unlock) |= msk;
+	*(cgu_base + cgu_clk_en) &= ~msk;
+	*(cgu_base + cgu_unlock) &= ~msk;
+}
```
agkaminski, line 263: Shouldn't we use `hal_cpuDataStoreBarrier` here?

hal/sparcv8leon3/spinlock.c
```diff
@@ -0,0 +1,128 @@
...
+
+struct {
+	spinlock_t spinlock;
+	spinlock_t *first;
+} spinlock_common;
+
+
+void hal_spinlockSet(spinlock_t *spinlock, spinlock_ctx_t *sc)
+{
+	__asm__ volatile(" \
```
agkaminski, line 31: Best to use `/* clang-format off */` and `on` on inline asms

proc/process.c
```diff
@@ -634,13 +634,37 @@ static int process_relocate(struct _reloc *reloc, size_t relocsz, char **addr)
 }
 
 
+static int process_relocateAddend(struct _reloc *reloc, size_t relocsz, char **addr, int addend)
```
agkaminski, line 637: Add addend argument to existing `process_relocate()` and provide zero on armv7m target.
Relocation method differs somewhat, but I think result should be the same.

proc/process.c
```diff
@@ -757,6 +781,51 @@ int process_load(process_t *process, vm_object_t *o, offs_t base, size_t size, v
 	if (process_relocate(reloc, relocsz, (char **)entry) < 0)
 		return -ENOEXEC;
 
+#ifdef __sparc__
```
agkaminski, line 784: Something for later, but we need to aspire to remove this `ifdef`. We could make a distinction by presence of `.rela` header

proc/process.c
```diff
@@ -930,6 +1000,8 @@ static void process_exec(thread_t *current, process_spawn_t *spawn)
 	if (!err) {
 		stack = process_putargs(stack, &spawn->envp, &count);
 		stack = process_putargs(stack, &spawn->argv, &count);
+		/* align stack down */
+		stack = (void *)(((unsigned int)stack - 8) & ~7);
```
agkaminski, line 1004:
```suggestion
		stack = (void *)(((ptr_t)stack - 8) & ~7);
```

agkaminski: Won't this alignment backfire? We'll lose the position of our arguments, if we really move the stack here.

Something for later: We should preallocate the space on stack (we need to find out size of arguments) and put arguments from aligned place (bottom -> up) instead of putting them top -> down and hoping for aligned result (not is scope of this PR ofc)

agkaminski:
I forgot, let's mark stack sizes as FIXME - we need to shrink them when we add full kernel/user stack support

proc/process.c
lukileczo: Refactored `hal_jmp` a little bit so that we won't end up with unaligned stack (and deleted this change)

hal/sparcv8leon3/gr716/gr716.c
agkaminski: I've meant after unlocking and before relocking

agkaminski (dismissed):
LGTM

ld/sparcv8leon3.ld
```diff
@@ -0,0 +1,209 @@
...
+  .debug_varnames  0 : { *(.debug_varnames) }
+  /* DWARF 3 */
+  .debug_pubtypes 0 : { *(.debug_pubtypes) }
+  .debug_ranges   0 : { *(.debug_ranges) }
+  /* DWARF Extension.  */
+  .debug_macro    0 : { *(.debug_macro) }
+  .debug_addr     0 : { *(.debug_addr) }
+  .gnu.attributes 0 : { KEEP (*(.gnu.attributes)) }
+  /DISCARD/ : { *(.note.GNU-stack) *(.gnu_debuglink) *(.gnu.lto_*) }
+}
```
agkaminski, line 209: No new line at the end of the file

agkaminski (commented):
Tiny nitpick, otherwise LGTM

agkaminski (dismissed):

agkaminski (approved):

PR #374, opened 2023-03-02
posix/unix: support socket receive buffer reallocation

ziemleszcz:
Allows setting SO_RCVBUF on sockets created by socketpair(). Data from old buffer is not copied to new buffer.

DONE: BES-239

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt-1064).

## Checklist:

## Special treatment

### DIFF
--- a/posix/unix.c
+++ b/posix/unix.c
@@ -866,16 +866,35 @@ int unix_shutdown(unsigned socket, int how)
 }
 
 
-/* TODO: support buffer reallocation */
+/* TODO: copy data from old buffer */
 static int unix_bufferSetSize(unixsock_t *s, int sz)
 {
-	if (sz < US_MIN_BUFFER_SIZE || sz > US_MAX_BUFFER_SIZE)
-		return -EINVAL;
+	void *v[2] = { NULL, NULL };
 
-	if (s->buffer.data != NULL)
+	if (sz < US_MIN_BUFFER_SIZE || sz > US_MAX_BUFFER_SIZE) {
 		return -EINVAL;
+	}
+
+	proc_lockSet(&s->lock);
+
+	if (s->buffer.data != NULL) {
+		v[0] = vm_kmalloc(sz);
+		if (v[0] == NULL) {
+			proc_lockClear(&s->lock);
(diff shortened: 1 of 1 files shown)


posix/unix.c
```diff
@@ -866,14 +866,23 @@ int unix_shutdown(unsigned socket, int how)
...
 		return -EINVAL;
 
-	if (s->buffer.data != NULL)
-		return -EINVAL;
+	if (s->buffer.data != NULL) {
+		if ((v = vm_kmalloc(sz)) == NULL)
+			return -ENOMEM;
+
+		proc_lockSet(&s->lock);
+		vm_kfree(s->buffer.data);
```
agkaminski, line 882: I would be nicer to do `vm_kfree()` outside of the lock (e.g. after new buffer init).

posix/unix.c
```diff
@@ -866,17 +866,39 @@ int unix_shutdown(unsigned socket, int how)
...
+	if (s->buffer.data != NULL) {
+		proc_lockClear(&s->lock);
+
+		v[0] = vm_kmalloc(sz);
+		if (v[0] == NULL) {
+			return -ENOMEM;
+		}
+
+		proc_lockSet(&s->lock);
+
```
agkaminski, line 889: I think we should either check here again if s->buffer.data is still != NULL, or don't let go of the lock :( Why we don't allocate a buffer if s->buffer data is NULL anyway? This would've make things simpler

ziemleszcz: I can remove `proc_lockClear` if you are OK with `vm_kmalloc` under lock. Access to socket cbuffer (including checking s->buffer.data) should always be done under lock.

agkaminski: Ok, it's a simpler solution, I'm ok with that

ziemleszcz: The current one or the one without `proc_lockClear`?

agkaminski: Current one is potentialy dangerous, so it's better to remove unlock/lock

agkaminski (approved):
LGTM

PR #379, opened 2023-03-20
klog: bugfixes, log_write and fifo rewrite

agkaminski:
JIRA: RTOS-404

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (ia32-generic, stm32l4).

## Checklist:

## Special treatment

### DIFF
--- a/log/log.c
+++ b/log/log.c
@@ -46,6 +46,7 @@ typedef struct _log_reader_t {
 	unsigned nonblocking;
 	log_rmsg_t *msgs;
 	struct _log_reader_t *prev, *next;
+	int refs;
 } log_reader_t;
 
 
@@ -59,15 +60,16 @@ static struct {
 	volatile int enabled;
 } log_common;
 
+
 static int _log_empty(void)
 {
-	return log_common.tail == log_common.head;
+	return (log_common.tail == log_common.head) ? 1 : 0;
 }
 
 
 static int _log_full(void)
 {
-	return log_common.tail - log_common.head == SIZE_LOG;
+	return ((log_common.tail - log_common.head) == SIZE_LOG) ? 1 : 0;
 }
 
 
@@ -79,10 +81,7 @@ static char _log_pop(void)
 
 static void _log_push(char c)
 {
-	log_common.buf[log_common.tail % SIZE_LOG] = c;
-	log_common.tail++;
-	if (_log_full())
-		_log_pop();
+	log_common.buf[log_common.tail++ % SIZE_LOG] = c;
 }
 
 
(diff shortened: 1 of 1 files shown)


log/log.c
```diff
@@ -337,24 +380,19 @@ void log_msgHandler(msg_t *msg, oid_t oid, unsigned long int rid)
 int log_write(const char *data, size_t len)
 {
 	size_t i = 0;
-	int overwrite = 0;
 	char c;
 
 	if (log_common.enabled != 0) {
 		proc_lockSet(&log_common.lock);
```
badochov, line 387: Shoudn't `if (log_common.enabled != 0)` be checked once again after lock?

log/log.c
```diff
@@ -158,25 +157,70 @@ static log_reader_t *log_readerFind(pid_t pid)
...
+}
+
+
 static int log_readerAdd(pid_t pid, unsigned nonblocking)
 {
 	log_reader_t *r;
 
-	if (log_readerFind(pid) != NULL)
+	r = log_readerFind(pid);
+
```
badochov, line 208: I'd delete this empty line

badochov (commented):
I think there is still one race condition in the code

log/log.c
```diff
@@ -59,15 +60,16 @@ static struct {
...
 {
-	return log_common.tail == log_common.head;
+	return (log_common.tail == log_common.head) ? 1 : 0;
 }
 
 
 static int _log_full(void)
 {
-	return log_common.tail - log_common.head == SIZE_LOG;
+	return (((log_common.tail + 1) % SIZE_LOG) == log_common.head) ? 1 : 0;
```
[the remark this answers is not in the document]
agkaminski: tail == head means empty, we need to differ fifo empty vs fifo full, so if next tail causes tail == head, we cannot increment tail to not cause sudden fifo empty

log/log.c
agkaminski: I've analysed the code and I decided against the change. `log_disable` is only needed if we predict that emptying klog to the console by uart server or via dmesg command will not be possible anymore - pretty much kernel panic only. It is assumed that before printing the dying message `log_disable()` would be invoked and log safely printed to the terminal. We don't really care about other threads in this case and in case of hazard no dangerous condition arise - it's still safe to write to the log.

log/log.c
```diff
@@ -310,6 +351,7 @@ void log_msgHandler(msg_t *msg, oid_t oid, unsigned long int rid)
 				else if (msg->o.io.err == 0 && r->nonblocking) {
```
lukileczo, line 351: Maybe change the `r->nonblocking` checks to be MISRA compliant too?

log/log.c
badochov: do you think it worth to leave a comment in the code about it?

agkaminski: I've added a brief comment

badochov (approved):
LGTM

PR #387, opened 2023-04-12
log: postpone log scrub to respond to write earlier

xvuko:
This can improve relatively slow klog write times. It is still important to know that logging from threads with higher priority than usrv can result in priority inversion.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/log/log.c
+++ b/log/log.c
@@ -367,6 +367,10 @@ void log_msgHandler(msg_t *msg, oid_t oid, unsigned long int rid)
 			break;
 		case mtWrite:
 			msg->o.io.err = log_write(msg->i.data, msg->i.size);
+			/* Writing thread is blocked on this write. Response should be sent as soon as possible */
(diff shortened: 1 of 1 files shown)


agkaminski (commented):
How is it blocked here? `log_scrub` only does `proc_respond`, it doesn't block in any way (except on a potential lock acquisition, but we need to get it anyway)

xvuko:
> How is it blocked here? `log_scrub` only does `proc_respond`, it doesn't block in any way (except on a potential lock acquisition, but we need to get it anyway)

It locks this lock second time and wakes up reader threads. If readers threads have same or higher priority than message handler than they could be scheduled before response is sent.

agkaminski:
If read has higher priority it should be scheduled instead. I agree on the lock though, it should could be taken once for both operations instead.

How about moving scrub after common `proc_respond` instead? I don't like ad hoc look of the current version.

PR #389, opened 2023-04-13
vm/object: Fix deadlock in vm_objectGet

agkaminski:
JIRA: RTOS-424

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (DTR).

## Checklist:

## Special treatment

### DIFF
--- a/vm/object.c
+++ b/vm/object.c
@@ -53,9 +53,10 @@ static int object_cmp(rbnode_t *n1, rbnode_t *n2)
 
 int vm_objectGet(vm_object_t **o, oid_t oid)
 {
-	vm_object_t t;
-	int i, n;
-	size_t sz;
+	vm_object_t t, *no = NULL;
+	size_t i, n;
+	offs_t sz;
+	int err = -ENOMEM;
 
 	t.oid.port = oid.port;
 	t.oid.id = oid.id;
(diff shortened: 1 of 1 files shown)


vm/object.c
```diff
@@ -64,27 +64,41 @@ int vm_objectGet(vm_object_t **o, oid_t oid)
...
 	if (*o == NULL) {
+		/* Take off the lock to avoid a deadlock in vm_kmalloc */
+		proc_lockClear(&object_common.lock);
+
 		sz = (size_t)proc_size(oid);
 		n = round_page(sz) / SIZE_PAGE;
 
-		if ((*o = (vm_object_t *)vm_kmalloc(sizeof(vm_object_t) + n * sizeof(page_t *))) == NULL) {
-			proc_lockClear(&object_common.lock);
+		if ((no = (vm_object_t *)vm_kmalloc(sizeof(vm_object_t) + n * sizeof(page_t *))) == NULL) {
```
kemonats, line 73:
```suggestion
		no = (vm_object_t *)vm_kmalloc(sizeof(vm_object_t) + n * sizeof(page_t *));
		if (no == NULL) {
```
What happens if `proc_size()` returns a negative value e.g. `-ENOMEM`.

vm/object.c
```diff
@@ -64,27 +64,41 @@ int vm_objectGet(vm_object_t **o, oid_t oid)
...
 			return -ENOMEM;
```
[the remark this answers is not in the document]
agkaminski: We can, why not. Code readability suffers a bit, but it's not that bad imho

vm/object.c
```diff
@@ -64,27 +64,49 @@ int vm_objectGet(vm_object_t **o, oid_t oid)
...
-		sz = (size_t)proc_size(oid);
-		n = round_page(sz) / SIZE_PAGE;
+		/* Take off the lock to avoid a deadlock in vm_kmalloc */
+		proc_lockClear(&object_common.lock);
 
-		if ((*o = (vm_object_t *)vm_kmalloc(sizeof(vm_object_t) + n * sizeof(page_t *))) == NULL) {
-			proc_lockClear(&object_common.lock);
+		sz = proc_size(oid);
+		if (sz < 0) {
+			return (int)sz;
```
ziemleszcz, line 72: Hmm, should we check again for object after `proc_size` fail (it calls `vm_kmalloc`)? This should probably apply to all error paths.

agkaminski: I don't know, this really shouldn't happen

ziemleszcz: It's very unlikely but it could happen. You could add `else` to `if` at lines 71 and 74 and nest them so that in case of failure we'll always proceed to second object check. Error code could be saved in a new variable and returned at line 87.

vm/object.c
```diff
@@ -64,27 +65,47 @@ int vm_objectGet(vm_object_t **o, oid_t oid)
...
+		proc_lockClear(&object_common.lock);
 
-		if ((*o = (vm_object_t *)vm_kmalloc(sizeof(vm_object_t) + n * sizeof(page_t *))) == NULL) {
-			proc_lockClear(&object_common.lock);
-			return -ENOMEM;
+		sz = proc_size(oid);
+		if (sz < 0) {
+			err = (int)sz;
+		}
+		else if ((sizeof(offs_t) <= sizeof(size_t)) || (sz < (offs_t)((size_t)-1))) {
```
ziemleszcz, line 75: sz <= (offs_t)((size_t)-1)?

vm/object.c
```diff
@@ -64,27 +65,47 @@ int vm_objectGet(vm_object_t **o, oid_t oid)
...
-		n = round_page(sz) / SIZE_PAGE;
+		/* Take off the lock to avoid a deadlock in vm_kmalloc */
+		proc_lockClear(&object_common.lock);
 
-		if ((*o = (vm_object_t *)vm_kmalloc(sizeof(vm_object_t) + n * sizeof(page_t *))) == NULL) {
-			proc_lockClear(&object_common.lock);
-			return -ENOMEM;
+		sz = proc_size(oid);
+		if (sz < 0) {
+			err = (int)sz;
```
badochov, line 73: Shouldn't we exit the function at this point?

vm/object.c
```diff
@@ -64,27 +65,47 @@ int vm_objectGet(vm_object_t **o, oid_t oid)
...
 	}
 
 	(*o)->refs++;
 	proc_lockClear(&object_common.lock);
 
+	/* Did we allocate an object we didn't need in the end? */
+	if (no != NULL) {
+		vm_kfree(no);
+	}
+
```
badochov, line 108: Else to if in L83 seems to be a better place

vm/object.c
agkaminski: We'll just try to find the object in the tree again and exit in line 86 if we fail again. It's a feature @ziemleszcz requested and it doesn't really hurt anything

badochov (approved):

PR #391, opened 2023-04-14
syscalls: Remove log_scrub from syscalls_dispatch

agkaminski:
There is a problem in msg implementation on MMU targets - it is assumed that the same process does proc_respond and it got the msg from proc_recv. In this case kernel thread did proc_recv and some random user thread does proc_respond, which causes unability to unmap msg related memory in proc_respond. log_scrub is not really needed here, so I chose to fix the problem this way.

DONE: RTOS-434

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (DTR, SKAL).

## Checklist:

## Special treatment

### DIFF
--- a/syscalls.c
+++ b/syscalls.c
@@ -25,7 +25,6 @@
 #include "vm/object.h"
 #include "posix/posix.h"
 #include "syspage.h"
-#include "log/log.h"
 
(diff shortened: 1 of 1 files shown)


xvuko (approved):
LGTM

PR #394, opened 2023-04-20
Kernel asserts implementation

agkaminski:
## Description

Implementation of asserts + some use cases of these

## Motivation and Context

To aid in code quality control and ease debugging

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (DTR).

## Checklist:

## Special treatment

### DIFF
--- a/lib/assert.h
+++ b/lib/assert.h
@@ -1,0 +1,38 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * Debug assert
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Aleksander Kaminski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#ifndef LIB_ASSERT_H_
+#define LIB_ASSERT_H_
+
+
+void lib_assertPanic(const char *func, int line, const char *fmt, ...);
+
+
+#define LIB_ASSERT_ALWAYS(condition, fmt, ...) \
+	if (!(condition)) { \
+		lib_assertPanic(__FUNCTION__, __LINE__, fmt, ##__VA_ARGS__); \
+	}
+
+#ifndef NDEBUG
+
+#define LIB_ASSERT(condition, fmt, ...) LIB_ASSERT_ALWAYS(condition, fmt, ##__VA_ARGS__)
+
+#else
+
+#define LIB_ASSERT(condition, fmt, ...)
+
+#endif
+
+#endif
--- a/lib/list.c
+++ b/lib/list.c
@@ -51,3 +51,25 @@ void lib_listRemove(void **list, void *t, size_t noff, size_t poff)
 	*((addr_t *)(t + noff)) = NULL;
 	*((addr_t *)(t + poff)) = NULL;
 }
+
+
+int lib_listBelongs(void **list, void *t, size_t noff, size_t poff)
+{
+	void *iter;
+
+	(void)poff;
+
+	if ((t == NULL) || ((*list) == NULL)) {
+		return 0;
+	}
+
+	iter = *list;
+	do {
+		if (iter == t) {
+			return 1;
+		}
+		iter = (void *)*((addr_t *)(iter + noff));
+	} while (iter != *list);
+
+	return 0;
+}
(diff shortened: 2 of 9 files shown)


lib/list.h
```diff
@@ -39,4 +42,10 @@ extern void lib_listRemove(void **list, void *t, size_t noff, size_t poff);
 #define LIST_REMOVE(list, t) LIST_REMOVE_EX(list, t, next, prev)
 
 
+#define LIST_BELONGS_EX(list, t, next, prev) \
+	lib_listBelongs((void **)(list), (void *)t, (size_t)&(((typeof(t))0)->next), (size_t)&(((typeof(t))0)->prev))
```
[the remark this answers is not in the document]
agkaminski: lol, no

astalke (approved):
Looks good to me

lib/list.c
```diff
@@ -51,3 +51,25 @@ void lib_listRemove(void **list, void *t, size_t noff, size_t poff)
 	*((addr_t *)(t + noff)) = NULL;
 	*((addr_t *)(t + poff)) = NULL;
 }
+
+
+int lib_listBelongs(void **list, void *t, size_t noff, size_t poff)
```
badochov, line 56: why does this function take poffs?

lib/assert.h
```diff
@@ -0,0 +1,37 @@
...
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#ifndef LIB_ASSERT_H_
+#define LIB_ASSERT_H_
+
+
+void __lib_assert(int condition, const char *func, int line, const char *fmt, ...);
```
badochov, line 20: why does this function's name begin with double underscore?

proc/threads.c
```diff
@@ -1649,7 +1668,15 @@ static void proc_lockUnlock(lock_t *lock)
...
-	if (lock->owner == NULL) {
+	thread_t *current = proc_current();
+
+	LIB_ASSERT(lock->owner != NULL, "lock: %s, pid: %d, tid: %d, unlock on not locked lock",
+		lock->name, (current->process != NULL) ? current->process->id : 0, current->id);
+
+	LIB_ASSERT(lock->owner == current, "lock: %s, pid: %d, tid: %d, owner: %d, unlocking someone's else lock",
+		lock->name, (current->process != NULL) ? current->process->id : 0, current->id, lock->owner->id);
+
+	if (lock->owner != current) {
```
badochov, line 1679: why was the check for lock->owner == NULL removed?

proc/threads.c
```diff
@@ -1601,12 +1613,16 @@ int proc_lockSetInterruptible(lock_t *lock)
 
 static int _proc_lockUnlock(lock_t *lock)
 {
-	thread_t *owner = lock->owner;
+	thread_t *owner = lock->owner, *current = proc_current();
```
badochov, line 1616: taking current_proc requires taking a spinlock I think it would be better to avoid calling this function in a non debug mode, as I belive that compiler won't optimize away this call.

lib/list.c
agkaminski: To keep the API the same as the rest of these functions. I'd rather keep it as is

lib/assert.h
agkaminski: Fixed

proc/threads.c
agkaminski: If lock->owner == NULL then lock->owner != current (as current is never NULL). I guess we allowed to free mutex owned by a different thread, so I think I'll restore this possibility

proc/threads.c
agkaminski: Fixed

proc/threads.c
```diff
@@ -1601,12 +1613,17 @@ int proc_lockSetInterruptible(lock_t *lock)
...
 {
-	thread_t *owner = lock->owner;
+	thread_t *owner = lock->owner, *current;
 	spinlock_ctx_t sc;
 	int ret = 0, lockPriority;
 
 	hal_spinlockSet(&threads_common.spinlock, &sc);
 
+	current = _proc_current();
+	(void)current; /* Unused in non-debug build */
```
badochov, line 1623: I'd wrap L1622 in an if macro so that it's present only when NDEBUG is set

agkaminski: I'd rather avoid introducing `#if`s if possible

badochov (approved):
LGTM

PR #396, opened 2023-05-04
Never have assgined invalid map to the process

agkaminski:
## Description

Assign map to the process only after successful process_load, deassign before map destroy.

## Motivation and Context

Should fix a problem with signal push causing exception 

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


proc/process.c
```diff
@@ -90,17 +90,20 @@ process_t *proc_find(unsigned pid)
 static void process_destroy(process_t *p)
 {
 	thread_t *ghost;
+	vm_map_t *mapp = p->mapp, *imapp = p->imapp;
```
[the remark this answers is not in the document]
nalajcie: as this is a recurring problem, I would advise to add `.codespell_ignore` file - like in https://github.com/phoenix-rtos/phoenix-rtos-ports/pull/54/commits/b0d43c0fc58040eed213a92b8e6591b6edff4ca6

nalajcie (dismissed):
seems ok, I don't know kernel code enough to verify the fix is complete. Approving anyway as it's a step in right direction (manipulating process maps under lock).

lukileczo (approved):

proc/process.c
```diff
@@ -1365,7 +1367,7 @@ static int process_execve(thread_t *current)
 	else {
 		/* Reinitialize process */
 		map = current->process->mapp;
-		proc_changeMap(current->process, NULL, NULL);
+		proc_changeMap(current->process, NULL, NULL, NULL);
```
lukileczo, line 1370: setting process->imapp here changes behaviour for if in line 1375

agkaminski: Fixed, thanks

PR #407, opened 2023-05-31
STM32L4 low power timer fixes

agkaminski:
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4).

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/stm32/l4/timer.c
+++ b/hal/armv7m/stm32/l4/timer.c
@@ -43,7 +43,8 @@ static struct {
 
 	volatile u32 *lptim;
 	volatile time_t upper;
-	volatile int wakeup;
+
+	intr_handler_t timerh;
 } timer_common;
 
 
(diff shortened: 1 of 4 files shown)


proc/threads.c
```diff
@@ -454,7 +453,8 @@ int threads_timeintr(unsigned int n, cpu_context_t *context, void *arg)
 
 	hal_spinlockClear(&threads_common.spinlock, &sc);
 
-	return EOK;
+	/* Invoke scheduler */
+	return 1;
```
xvuko, line 457: This invokes sheduler only on cpu 0 due to early return earlier.

agkaminski: Missed that, fixed, thanks!

hal/armv7m/stm32/l4/timer.c
```diff
@@ -163,19 +165,45 @@ void timer_setAlarm(time_t us)
 		ticks = (ARR_VAL * 2) / 3;
```
xvuko, line 160: This is no longer needed as it implies that (setval > ARR_VAL)

agkaminski: Overflow check still needed, sightly reformatted the condition

astalke (approved):

anglov (approved):

PR #419, opened 2023-07-20
posix: Reduce max fd count on all NOMMU targets

agkaminski:
- increase max fd count on STM32 from 8 to 32,
- free ~3840 bytes per process on imxrt* and leon3 targets

DONE: RTOS-522

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -23,8 +23,8 @@
 #include "posix_private.h"
 #include "../lib/cbuffer.h"
 
-#ifdef CPU_STM32
-#define MAX_FD_COUNT 8
+#ifdef NOMMU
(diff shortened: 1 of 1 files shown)


anglov:
Please be only sure if this won't create any issues with network communication on imxrt (not sure if creating socket doesn't create  additional fd)

If so, maybe it would be reasonable to set this value to large enough but much smaller than 512 (32? 64?). 

I'm ok with it on stm32 target

gerard5 (dismissed):
It works for me, verified on N.

anglov (approved):
LGTM

lukileczo (approved):

PR #425, opened 2023-07-31
posix: Code cleanup and dynamic fd table size

agkaminski:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


kemonats:
1. Some suggestions for formating the code:
- line 29 - change `//` to `/* */`
- line 69, 83 - add empty line after declaration variable
- line 1058 - missing curly brackets
- line 1626 - add space before `*`
- line 1679, 2178 - format `switch`

2. Two suggestions not directly related to this PR:
- moving the `splitname()` function to `lib/` directory, it is also used in `unix.c` file.
- create `strdup()` function, can be used in functions:
```c
    posix_link()
    posix_unlink()
    posix_create()
    unix_bind()
    proc_start()
    process_copy()
    proc_execve()
```

agkaminski:
@kemonats thank you for your suggestions, I've applied all of them

posix/posix.c
```diff
@@ -826,12 +904,14 @@ int posix_pipe(int fildes[2])
 	fildes[0] = 0;
 
 	proc_lockSet(&p->lock);
-	while (p->fds[fildes[0]].file != NULL && fildes[0]++ < p->maxfd);
+	while (p->fds[fildes[0]].file != NULL && fildes[0]++ < p->maxfd) {
```
badochov: brackets missing same in while below

posix/posix.c
```diff
@@ -826,12 +904,14 @@ int posix_pipe(int fildes[2])
...
+	}
 
 	fildes[1] = fildes[0] + 1;
-	while (p->fds[fildes[1]].file != NULL && fildes[1]++ < p->maxfd);
+	while (p->fds[fildes[1]].file != NULL && fildes[1]++ < p->maxfd) {
```
badochov: i think that using increment in while condition is not compliant to misra rule 13.3, however I am not sure if I interpret that rule right

while loop may instead be changed to a for loop this way it will be compliant for sure and clearer imo

posix/posix.c
```diff
@@ -1207,20 +1374,25 @@ static int posix_fcntlDup(int fd, int fd2, int cloexec)
...
 	proc_lockSet(&p->lock);
-	if (fd < 0 || fd > p->maxfd || fd2 < 0 || fd > p->maxfd) {
+	if ((fd < 0) || (fd > p->maxfd) || (fd2 < 0) || (fd2 > p->maxfd)) {
 		proc_lockClear(&p->lock);
 		pinfo_put(p);
 		return -EBADF;
 	}
 
-	while (p->fds[fd2].file != NULL && fd2++ < p->maxfd);
+	while ((p->fds[fd2].file != NULL) && (fd2++ < p->maxfd)) {
```
badochov: i think that using increment in while condition is not compliant to misra rule 13.3, however I am not sure if I interpret that rule right

while loop may instead be changed to a for loop this way it will be compliant for sure and clearer imo

posix/posix.c
```diff
@@ -2367,97 +2350,6 @@ int posix_poll(struct pollfd *fds, nfds_t nfds, int timeout_ms)
 	return ready;
 }
 
-#else
-
-int posix_poll(struct pollfd *fds, nfds_t nfds, int timeout_ms)
```
badochov, line 2128: what was the purpose of this code?

posix/posix.c
```diff
@@ -887,14 +899,21 @@ int posix_pipe(int fildes[2])
 	fildes[0] = 0;
```
badochov: can be moved to the first cluase of the for loop

posix/posix.c
```diff
@@ -887,14 +899,21 @@ int posix_pipe(int fildes[2])
...
 
 	proc_lockSet(&p->lock);
-	while (p->fds[fildes[0]].file != NULL && fildes[0]++ < p->maxfd) {
+	for (; fildes[0] < p->maxfd; ++fildes[0]) {
+		if (p->fds[fildes[0]].file == NULL) {
+			break;
+		}
 	}
 
 	fildes[1] = fildes[0] + 1;
```
badochov: can be moved to the first cluase of the for loop

posix/posix.c
```diff
@@ -196,7 +196,7 @@ int posix_getOpenFile(int fd, open_file_t **f)
 	}
 
 	proc_lockSet(&p->lock);
-	if ((fd < 0) || (fd > p->maxfd) || (p->fds[fd].file == NULL)) {
+	if ((fd < 0) || (fd >= p->maxfd) || (p->fds[fd].file == NULL)) {
```
badochov: maxbe change the name from `maxfd` to `fdsize`?

lib/strutil.c
```diff
@@ -0,0 +1,66 @@
...
+
+static char *lib_strrchr(const char *s, int c)
+{
+	const char *p = NULL;
+
+	do {
+		if (*s == c) {
+			p = s;
+		}
+	} while (*(s++) != '\0');
```
badochov, line 28: is incremetn in while condition compliant with misra 13.3?

badochov: I'd rewrite it using for loop for clarity

posix/posix.c
badochov: ow it's already done in the next commit

posix/posix.c
```diff
@@ -214,19 +211,49 @@ int posix_getOpenFile(int fd, open_file_t **f)
...
 {
-	open_file_t *f;
-
-	proc_lockSet(&p->lock);
+	fildes_t *nfds;
+	int nfdsz;
 
 	for (; fd < p->maxfd; ++fd) {
+		if (fd >= p->fdsz) {
+			nfdsz = 2 * p->fdsz;
```
badochov, line 185: when `p->fdsz == p->maxfd`, whenever user tries to alocate new file we reallocate memory, I'd add a check to prevent that

badochov (changes requested):

posix/posix.c
```diff
@@ -201,24 +175,61 @@ int posix_getOpenFile(int fd, open_file_t **f)
...
+
+			vm_kfree(p->fds);
+
+			p->fds = nfds;
+			p->fdsz = nfdsz;
+		}
+
+		if (p->fds[fd].file == NULL) {
+			return fd;
+		}
```
kemonats, line 206: I see a problem in this function.
Such a case:
`fd == 128`, `fdsz == 32`, `maxfd == 1024`
`fd < maxfd` and `fd > fdsz`
you are realocating and then `fdsz == 64`
and in the condition
`p->fds[fd].file == NULL`
array index is out of range.

Can such case occur?
if so, then solution is to change calculation of the value `nfdsz`.

posix/posix.c
kemonats: For what value of `fd` will this occur?

posix/posix.c
agkaminski: Alternative approach to poll, never tested, without support in device nor fs drivers. I guess I got carried away with removing it, thought, as I should remove bunch of additional code too. I'll restore it for now.

posix/posix.c
agkaminski: if `p->fdsz == p->maxfd` and `fd < p->maxfd` we'll never go into `fd >= p->fdsz` case, so no relocation will occur.
if `p->fdsz == p->maxfd` and `fd >= p->maxfd` we won't enter `for` loop and we'll return `-1`

Yes, we allow `fd == p->fdsz`, but please note that we deny `fd == p->maxfd` using `for` condition.

Or am I missing something here?

posix/posix.c
agkaminski: Ooops, big bug, thanks for spotting that

agkaminski:
@badochov I'm skipping remarks regarding intermediate commits, as these lines are replaced altogether along the way

lib/strutil.c
agkaminski: Done

badochov (approved):
LGTM

lukileczo:
What about the commented-out `posix_exit` declaration (`posix.h`, line 75)?

lib/strutil.c
```diff
@@ -0,0 +1,67 @@
...
+	const char *p = NULL;
+	const char *i;
+
+	for (i = s; *i != '\0'; ++i) {
+		if (*i == c) {
+			p = i;
+		}
+	}
+
+	return (char *)p;
```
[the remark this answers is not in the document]
agkaminski: It's a retarded `strrchr` API that causes this. I can change it thought, as it is not a libc implementation

lukileczo (approved):
LGTM

Issue #171, opened 2021-01-25
armv7m: Signals don't work.


MaciejPurski:
Function `hal_cpuPushSignal` is not implemented on this architecture. I tried to use signals on imxrt1064 CPU.
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/f84dea0c56a9743e9b9825911aff836a8b4f3c81/hal/armv7m/cpu.h#L333

PR #428, opened 2023-08-04
rework signal handling

lukileczo:
## Description

Rework signal handling:
- Signal context is now prepared on user stack
- Return from signal handler is done through new sigreturn syscall
## Motivation and Context

Fixes: https://github.com/phoenix-rtos/phoenix-rtos-kernel/issues/171
JIRA: RTOS-539
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `riscv64-generic-qemu`, `ia32-generic-qemu`, `armv7a9-zynq7000-qemu`, `armv7m4-stm32l4x6-nucleo`, `armv7a7-imx6ull-dataro`, `sparcv8leon3-gr716-mini`, `armv7m7-imxrt106x-evk`

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work: https://github.com/phoenix-rtos/libphoenix/pull/278

### DIFF
--- a/hal/armv7a/arch/cpu.h
+++ b/hal/armv7a/arch/cpu.h
@@ -161,9 +161,9 @@ static inline void hal_cpuRestore(cpu_context_t *curr, cpu_context_t *next)
 }
 
 
-static inline void hal_cpuSetReturnValue(cpu_context_t *ctx, int retval)
+static inline void hal_cpuSetReturnValue(cpu_context_t *ctx, void *retval)
 {
-	ctx->r0 = retval;
+	ctx->r0 = (u32)retval;
 }
 
 
--- a/hal/armv7a/arch/exceptions.h
+++ b/hal/armv7a/arch/exceptions.h
@@ -17,6 +17,7 @@
 #define _HAL_ARMV7A_EXCEPTIONS_H_
 
 #include "types.h"
+#include "cpu.h"
 
 #define EXC_DEFAULT 128
 
@@ -27,33 +28,12 @@
 
 
 typedef struct _exc_context_t {
-	u32 savesp;
-
 	u32 dfsr;
 	u32 dfar;
 	u32 ifsr;
 	u32 ifar;
 
-	u32 psr;
-
-	u32 r0;
-	u32 r1;
-	u32 r2;
-	u32 r3;
-	u32 r4;
-	u32 r5;
-	u32 r6;
-	u32 r7;
-	u32 r8;
-	u32 r9;
-	u32 r10;
-
-	u32 fp;
-	u32 ip;
-	u32 sp;
-	u32 lr;
-
-	u32 pc;
+	cpu_context_t cpuCtx;
 } exc_context_t;
 
 #endif
--- a/hal/armv7a/exceptions.c
+++ b/hal/armv7a/exceptions.c
@@ -64,27 +64,27 @@ void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
 	hal_strcpy(buff += hal_strlen(buff), "\n");
 	buff += hal_strlen(buff);
 
-	i += hal_i2s(" r0=", &buff[i], ctx->r0, 16, 1);
-	i += hal_i2s("  r1=", &buff[i], ctx->r1, 16, 1);
-	i += hal_i2s("  r2=", &buff[i], ctx->r2, 16, 1);
-	i += hal_i2s("  r3=", &buff[i], ctx->r3, 16, 1);
+	i += hal_i2s(" r0=", &buff[i], ctx->cpuCtx.r0, 16, 1);
+	i += hal_i2s("  r1=", &buff[i], ctx->cpuCtx.r1, 16, 1);
+	i += hal_i2s("  r2=", &buff[i], ctx->cpuCtx.r2, 16, 1);
+	i += hal_i2s("  r3=", &buff[i], ctx->cpuCtx.r3, 16, 1);
 
-	i += hal_i2s("\n r4=", &buff[i], ctx->r4, 16, 1);
-	i += hal_i2s("  r5=", &buff[i], ctx->r5, 16, 1);
-	i += hal_i2s("  r6=", &buff[i], ctx->r6, 16, 1);
-	i += hal_i2s("  r7=", &buff[i], ctx->r7, 16, 1);
+	i += hal_i2s("\n r4=", &buff[i], ctx->cpuCtx.r4, 16, 1);
+	i += hal_i2s("  r5=", &buff[i], ctx->cpuCtx.r5, 16, 1);
+	i += hal_i2s("  r6=", &buff[i], ctx->cpuCtx.r6, 16, 1);
+	i += hal_i2s("  r7=", &buff[i], ctx->cpuCtx.r7, 16, 1);
 
-	i += hal_i2s("\n r8=", &buff[i], ctx->r8, 16, 1);
-	i += hal_i2s("  r9=", &buff[i], ctx->r9, 16, 1);
-	i += hal_i2s(" r10=", &buff[i], ctx->r10, 16, 1);
-	i += hal_i2s("  fp=", &buff[i], ctx->fp, 16, 1);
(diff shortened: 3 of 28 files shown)


hal/armv7a/exceptions.c
```diff
@@ -19,6 +19,7 @@
 #include "../console.h"
 #include "../string.h"
 #include "../../include/mman.h"
+#include "../../proc/threads.h"
```
agkaminski, line 22: This should be avoided, sadly we indeed need one function from `proc` module, this can be solved like here: https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/4cdff06e1309956cb835aabc4ddcde42d7587441/hal/armv7m/stm32/interrupts.c#L76

proc/threads.c
```diff
@@ -1462,14 +1468,8 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
 
-	if (thread != NULL && hal_cpuPushSignal(thread->kstack + thread->kstacksz, thread->process->sighandler, sig) != EOK) {
+	if (thread != NULL) {
 		thread->sigpend |= sigbit;
-
-		if (sig == signal_segv || sig == signal_illegal) {
-			/* If they can't handle those right away, kill */
-			kill = 1;
-			_proc_threadExit(thread);
-		}
```
agkaminski, line 1472: Hmm, we need to handle insta-kill procedure somewhere. Did you checked that PF is able to kill guilty process (it had a problems on the current code too, because we determined that the process is able to handle the signal when in fact it was not)

proc/threads.c
```diff
@@ -1497,13 +1499,56 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+	thread_t *thread;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	thread = _proc_current();
+
+	kstackTop = thread->kstack + thread->kstacksz;
+	ctx = kstackTop - sizeof(*ctx);
+	signalCtx = (void *)((char *)hal_cpuGetUserSP(ctx) - sizeof(*signalCtx));
+
+	if (_threads_checkSignal(thread, thread->process, signalCtx) == 0) {
```
agkaminski, line 1540: Shouldn't we check for `thread->process == NULL`?

proc/threads.c
```diff
@@ -1497,13 +1499,56 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+	void *f;
+	void *kstackTop;
+	thread_t *thread;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	thread = _proc_current();
+
+	kstackTop = thread->kstack + thread->kstacksz;
+	ctx = kstackTop - sizeof(*ctx);
+	signalCtx = (void *)((char *)hal_cpuGetUserSP(ctx) - sizeof(*signalCtx));
```
agkaminski, line 1538: Perhaps do this stuff under if too, otherwise it's not needed

syscalls.c
```diff
@@ -914,6 +914,30 @@ int syscalls_signalSuspend(void *ustack)
 }
 
 
+int syscalls_sigreturn(void *ustack)
```
agkaminski, line 917: Always zero, we should kill the process imho in case of some shenanigans on it's part
```suggestion
void syscalls_sigreturn(void *ustack)
```

hal/armv7a/arch/cpu.h
```diff
@@ -161,7 +161,7 @@ static inline void hal_cpuRestore(cpu_context_t *curr, cpu_context_t *next)
 }
 
 
-static inline void hal_cpuSetReturnValue(cpu_context_t *ctx, int retval)
+static inline void hal_cpuSetReturnValue(cpu_context_t *ctx, ptr_t retval)
```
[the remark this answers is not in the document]
lukileczo: In `syscalls_dispatch`, the return value is set as `void *` so on 64-bit it's cast to int of different size. And this function sets value in context, which may be either `u32` or `u64`

hal/armv7m/imxrt/_init.S
```diff
@@ -109,32 +109,55 @@ _start:
...
+	/* r0-r3, r12, lr, pc, psr */
+	stmia r10!, {r1-r8}
+
+	/* s0-s7 */
+	ldmia r0!, {r1-r8}
+	stmia r10!, {r1-r8}
+
+	/* s8-s15, fpscr */
+	ldmia r0!, {r1-r8, r11}
+	stmia r10!, {r1-r8, r11}
```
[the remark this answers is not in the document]
lukileczo: These are the FPU registers, s0-s15, why this change?

proc/threads.c
lukileczo: I need `signalCtx` as arg to `_threads_checkSignal`

hal/armv7m/imxrt/_init.S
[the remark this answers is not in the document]
agkaminski: Misunderstood the code, sorry

hal/armv7a/arch/exceptions.h
```diff
@@ -27,13 +27,18 @@
...
 
 typedef struct _exc_context_t {
-	u32 savesp;
-
 	u32 dfsr;
 	u32 dfar;
 	u32 ifsr;
 	u32 ifar;
 
+	u32 savesp;
```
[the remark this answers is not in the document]
lukileczo: Yeah, I thought about it, will do.

proc/threads.c
lukileczo: Can we have userspace thread without a process? I'll add check if needed.

hal/armv7m/stm32/_init.S
```diff
@@ -130,20 +140,25 @@ _syscall_dispatch:
...
 	/* Switch stacks */
-	mov r0, #3
-	msr control, r0
+	mov r1, #3
+	msr control, r1
 	isb
 
-	ldmia sp!, {r0-r3, r12, lr}
+	add sp, sp, #(5 * 4)
+	pop {lr}
```
agkaminski, line 161: A tiny window where the stack is not aligned, but interrupts are enabled - tiny, but still might cause problems

agkaminski: Hmm, from what I see stm32 and imxrt implementation diverged, why? There shouldn't be any difference apart from FPU stuff

lukileczo: The implementations were a little bit different originally, and I couldn't find eval kit to test this further

agkaminski:
@anglov This might affect DEND project (sudden onset of previously unexpected EINTRs + greater user stack usage)

anglov:
> @anglov This might affect DEND project (sudden onset of previously unexpected EINTRs + greater user stack usage)

OK, I will check. Please let me know what is the worst case scenario:

Is it extend user stack? (eg. by interrupt handler). Or only kernel stack? 
If so - how many memory on user stack need to be preserved for interrupt handling? 

Is there any system mechanism for sending signals which can cause EINTR apart of proc ending?

agkaminski:
@anglov Signal will allocate full cpu_context_t on the user stack + stack needed for signal handler and it's variables. I don't think we should expect any signals other than SIGCHLD right now. Switch in board_config.h will be added to allow disabling the signals for compatibility purposes (hopefully temporally)

agkaminski (approved):
LGTM

gerard5:
Verified on imxrt117x (^C) works, thanks. I'm soooo happpyyyy!

code used  to verify:
```C
#include <stdio.h>
#include <stdlib.h>
#include <signal.h>
#include <unistd.h>

static int gQuit = 0; 
 
static void handler(int no)
{
    if (no == SIGINT) {
        gQuit = 1;
    }
}

int main(void)
{
    printf("press Ctrl+C to quit\n");

    signal(SIGINT, handler);

    while (gQuit == 0) {
        printf("idle\n");
        sleep(1);
    }

    printf("exit normally\n");

    return 0;
}

```

result:

2023-09-29-104600_882x478_scrot

PR #429, opened 2023-08-10
NOMMU/vm_mapDestroy: Remove only entries associated with the map

agkaminski:
DONE: RTOS-551

## Description

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/721

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/vm/map.c
+++ b/vm/map.c
@@ -752,6 +752,22 @@ int vm_mapCreate(vm_map_t *map, void *start, void *stop)
 }
 
 
+static void _map_free(map_entry_t *entry)
+{
+	map_common.nfree++;
+	entry->next = map_common.free;
+	map_common.free = entry;
+}
+
+
+void map_free(map_entry_t *entry)
+{
+	proc_lockSet(&map_common.lock);
+	_map_free(entry);
+	proc_lockClear(&map_common.lock);
+}
+
+
 void vm_mapDestroy(process_t *p, vm_map_t *map)
 {
 	map_entry_t *e;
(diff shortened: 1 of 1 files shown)


vm/map.c
```diff
@@ -775,11 +775,29 @@ void vm_mapDestroy(process_t *p, vm_map_t *map)
...
+			/* Remove only entries associated with the map */
+			if (e->map == map) {
+				LIST_REMOVE(&p->entries, e);
+				if (next == e) {
+					next = NULL;
+				}
+				e->process = NULL;
+				lib_rbRemove(&map->tree, &e->linkage);
+				e->map = NULL;
+				map_free(e);
```
badochov, line 794: `map_common.lock` is locked and unlocked here every time, maybe number of locking may be reduced?

vm/map.c
```diff
@@ -775,11 +775,29 @@ void vm_mapDestroy(process_t *p, vm_map_t *map)
...
+			}
+
+			e = next;
+		} while ((e != NULL) && (e != p->entries));
 	}
+	proc_lockClear(&p->lock);
 	proc_lockClear(&map->lock);
```
badochov, line 801: isn't `proc_lockDone(&map->lock);` missing?

vm/map.c
```diff
@@ -775,11 +775,29 @@ void vm_mapDestroy(process_t *p, vm_map_t *map)
...
-	proc_lockSet(&map->lock);
-	while ((e = p->entries) != NULL) {
-		_map_remove(map, e);
-		map_free(e);
+	map_entry_t *next;
+
+	proc_lockSet2(&map->lock, &p->lock);
+	if (p->entries != NULL) {
+		e = p->entries;
+		do {
```
badochov, line 783: `if` maybe removed

```c
e = p->entries;
while ((e != NULL) && (e != p->entries))
```

badochov (changes requested):

vm/map.c
agkaminski: Nope, if e == p-> entries we won't enter such while

badochov: Ow, right

vm/map.c
agkaminski: Done

vm/map.c
agkaminski: It's not, we never destroy a map on NOMMU for real, we only "detach" a process from it

vm/map.c
```diff
@@ -775,11 +791,31 @@ void vm_mapDestroy(process_t *p, vm_map_t *map)
...
 #else
-	proc_lockSet(&map->lock);
-	while ((e = p->entries) != NULL) {
-		_map_remove(map, e);
-		map_free(e);
+	map_entry_t *next;
+
+	proc_lockSet2(&map->lock, &p->lock);
+	proc_lockSet(&map_common.lock);
+	if (p->entries != NULL) {
```
lukileczo, line 798: Small optimization - locking `map_common.lock` may be moved inside the `if`

PR #450, opened 2023-09-13
msg: Return a handle instead of kernel pointer in msgRecv

agkaminski:
Breaking API change!

Add idtree from libphoenix, some MISRA improvements

JIRA: RTOS-606

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (a whole lot).

### DIFF
--- a/lib/idtree.c
+++ b/lib/idtree.c
@@ -1,0 +1,168 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * id allocating tree
+ *
+ * Copyright 2018, 2023 Phoenix Systems
+ * Author: Jan Sikorski, Aleksander Kaminski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "lib.h"
+
+
+static int lib_idtreeCmp(rbnode_t *n1, rbnode_t *n2)
+{
+	idnode_t *i1 = lib_treeof(idnode_t, linkage, n1);
+	idnode_t *i2 = lib_treeof(idnode_t, linkage, n2);
+
+	if (i1->id > i2->id) {
+		return 1;
+	}
+	else if (i2->id > i1->id) {
+		return -1;
+	}
+	else {
+		return 0;
+	}
+}
+
+
+static int lib_idtreeGapcmp(rbnode_t *n1, rbnode_t *n2)
+{
+	idnode_t *r1 = lib_treeof(idnode_t, linkage, n1);
+	idnode_t *r2 = lib_treeof(idnode_t, linkage, n2);
+	rbnode_t *child = NULL;
+	int ret;
+
+	if ((r1->lmaxgap > 0) && (r1->rmaxgap > 0)) {
+		if (r2->id > r1->id) {
+			child = n1->right;
+			ret = -1;
+		}
+		else {
+			child = n1->left;
+			ret = 1;
+		}
+	}
+	else if (r1->lmaxgap > 0) {
+		child = n1->left;
+		ret = 1;
+	}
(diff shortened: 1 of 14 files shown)


lib/idtree.c
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+
+static int lib_idtreeGapcmp(rbnode_t *n1, rbnode_t *n2)
+{
+	idnode_t *r1 = lib_treeof(idnode_t, linkage, n1);
+	idnode_t *r2 = lib_treeof(idnode_t, linkage, n2);
+	rbnode_t *child = NULL;
+	int ret = 1;
```
badochov, line 41: IMHO `ret` should be set to zero as that's its de facto default value, or not set at all as whenevere child != NULL then ret is set

lib/idtree.c
```diff
@@ -0,0 +1,169 @@
...
+
+	if (r1->lmaxgap > 0 && r1->rmaxgap > 0) {
+		if (r2->id > r1->id) {
```
badochov, line 44: r2-> id is always zero so there's no need for this branch thus this whole if else can be simplified to 
```c
if (r1->lmaxgap > 0) {
		child = n1->left;
		ret = 1;
	}
	else if (r1->rmaxgap > 0) {
		child = n1->right;
		ret = -1;
	}
```

agkaminski: Why do you think it's always zero? It's not obligatory to have id == 0 element in the tree:
- allocate id = 0,
- allocate id = 1,
- remove id = 0
and we have "id = 0"-less tree

lib/idtree.h
```diff
@@ -0,0 +1,53 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * id allocating tree
+ *
+ * Copyright 2018 Phoenix Systems
+ * Author: Jan Sikorski
```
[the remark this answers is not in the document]
agkaminski: It is, I've imported it from libphoenix. I modified it quite substantially now though, so I've added myself as a coauthor

proc/msg-nommu.c
```diff
@@ -55,11 +57,20 @@ int proc_send(u32 port, msg_t *msg)
...
+
+			if ((err != EOK) && (kmsg.state == msg_waiting)) {
+				LIST_REMOVE(&p->kmessages, &kmsg);
+				break;
+			}
+		}
+
+		if (kmsg.state == msg_responded) {
+			err = EOK; /* Don't report EINTR if we got the response already */
+		}
```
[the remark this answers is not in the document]
agkaminski: I guess it's better to do this in one place, so I've modified it as suggested

lib/rb.h
```diff
@@ -19,10 +19,10 @@
 //#include "../hal/hal.h"
```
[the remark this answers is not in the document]
agkaminski: Not really in scope of this PR, but ok

lib/idtree.c
agkaminski: As it turns out we don't need a child at all

agkaminski: Retreat, it's need afterall

lib/idtree.c
badochov: this function is only used in `lib_rbFindEx` where n2 is set to `n` whose id is set to 0. During the execution of `lib_rbFindEx` the second argument is constant thus r2->id is always 0

agkaminski: I'd rather have this code more generic than it's necessary right now, so I won't get bitten in the future

badochov: Ok

badochov (approved):

anglov (approved):
Tested by hand within DEND project and works fine

PR #454, opened 2023-09-22
msg-nommu: Fix rid being allocaded in the wrong place

agkaminski:
JIRA: RTOS-606

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/proc/msg-nommu.c
+++ b/proc/msg-nommu.c
@@ -107,18 +107,25 @@ int proc_recv(u32 port, msg_t *msg, msg_rid_t *rid)
 	}
 
 	kmsg = p->kmessages;
+	if (kmsg != NULL) {
+		kmsg->state = msg_received;
+		LIST_REMOVE(&p->kmessages, kmsg);
+	}
 
 	if (p->closed) {
 		/* Port is being removed */
 		if (kmsg != NULL) {
 			kmsg->state = msg_rejected;
-			LIST_REMOVE(&p->kmessages, kmsg);
 			proc_threadWakeup(&kmsg->threads);
 		}
 
-		hal_spinlockClear(&p->spinlock, &sc);
+		err = -EINVAL;
+	}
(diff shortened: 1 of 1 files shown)


proc/msg-nommu.c
```diff
@@ -107,19 +107,23 @@ int proc_recv(u32 port, msg_t *msg, msg_rid_t *rid)
...
 
 	if (p->closed) {
 		/* Port is being removed */
 		if (kmsg != NULL) {
 			kmsg->state = msg_rejected;
-			LIST_REMOVE(&p->kmessages, kmsg);
 			proc_threadWakeup(&kmsg->threads);
 		}
 
 		hal_spinlockClear(&p->spinlock, &sc);
```
lukileczo, line 122: I think spinlock clear may be safely moved earlier - `kmsg` is no longer accessible through `p`:
```suggestion
	if (p->closed) {
		/* Port is being removed */
		hal_spinlockClear(&p->spinlock, &sc);
		if (kmsg != NULL) {
			kmsg->state = msg_rejected;
			proc_threadWakeup(&kmsg->threads);
		}
```

agkaminski: I kinda agree, but sender still can access `kmsg` if it spuriously wakes up, same reason in lines 129 and 132

lukileczo:
Shouldn't there be an early return path if we get `EINTR` from `proc_threadWaitInterruptible` but the message has not arrived (i.e. is `NULL`)? Now we can get to `proc_portRidAlloc` with `NULL`

agkaminski:
> Shouldn't there be an early return path if we get `EINTR` from `proc_threadWaitInterruptible` but the message has not arrived (i.e. is `NULL`) and port is not closed? Now we can get to `proc_portRidAlloc` with `NULL` Should've looked at the first review 💀

You're right, should be fixed now

lukileczo (approved):
LGTM

PR #457, opened 2023-09-27
posix: posix_died rewrite

agkaminski:
- reap all init zombies (fix memory leak),
- consume SIGCHLD when parent is waiting in waitpid,
- reap orphan's zombies (fix memory leak).

DONE:  RTOS-624

## Description

Example code to showcase a memory leak that has been fixed:
```
#include <stdio.h>
#include <fcntl.h>
#include <stdlib.h>
#include <unistd.h>

void process(int level)
{
	while (fork()) {
		printf("level: %d\n", level);
		if (!--level) {
			break;
		}
	}
}


int main(int argc, char *argv[])
{
	process(5000);

	return 0;
}
```

## Motivation and Context

Fixes a memory leak.
Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/188 (hopefully)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: zynq7000-qemu.

## Checklist:

## Special treatment

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -2686,43 +2686,70 @@ int posix_waitpid(pid_t child, int *status, int options)
 
 void posix_died(pid_t pid, int exit)
 {
-	process_info_t *pinfo, *ppinfo, *init, *cinfo;
+	process_info_t *pinfo, *ppinfo, *init, *cinfo, *zinfo, *zombies;
+	int waited, adopted = 1;
 
 	pinfo = pinfo_find(pid);
-	if (pinfo == NULL) {
-		return;
-	}
+	LIB_ASSERT_ALWAYS(pinfo != NULL, "pinfo not found, pid: %d", pid);
+
+	init = pinfo_find(1);
+	LIB_ASSERT_ALWAYS(init != NULL, "init not found");
+
+	ppinfo = pinfo_find(pinfo->parent);
 
 	posix_exit(pinfo, exit);
 
-	ppinfo = pinfo_find(pinfo->parent);
+	/* We might not find a parent if it died just now */
 	if (ppinfo != NULL) {
+		/* Make a zombie, wakeup waitpid */
 		proc_lockSet(&ppinfo->lock);
-		LIST_REMOVE(&ppinfo->children, pinfo);
-		LIST_ADD(&ppinfo->zombies, pinfo);
-		proc_threadBroadcast(&ppinfo->wait);
-		proc_lockClear(&ppinfo->lock);
-
-		init = pinfo_find(1);
-		if (init != NULL) {
-			proc_lockSet2(&pinfo->lock, &init->lock);
-			while (pinfo->children != NULL) {
-				cinfo = pinfo->children;
-
-				/* Treat cinfo->parent as atomic, no need for lock */
-				cinfo->parent = 1;
-
-				LIST_REMOVE(&pinfo->children, cinfo);
-				LIST_ADD(&init->children, cinfo);
-			}
-			proc_lockClear(&pinfo->lock);
-			proc_lockClear(&init->lock);
-			pinfo_put(init);
+		/* Check if we didn't get adopted by the init in the meantime */
+		if ((ppinfo != init) && (LIST_BELONGS(&ppinfo->children, pinfo) != 0)) {
+			LIST_REMOVE(&ppinfo->children, pinfo);
+			LIST_ADD(&ppinfo->zombies, pinfo);
+			waited = proc_threadBroadcast(&ppinfo->wait);
+			adopted = 0;
 		}
-
-		posix_sigchild(pinfo->parent);
-
+		proc_lockClear(&ppinfo->lock);
 		pinfo_put(ppinfo);
+	}
+
+	proc_lockSet2(&pinfo->lock, &init->lock);
+	/* Collect all zombies */
+	zombies = pinfo->zombies;
+	pinfo->zombies = NULL;
+
+	/* Adopt children */
+	while (pinfo->children != NULL) {
+		cinfo = pinfo->children;
+		LIST_REMOVE(&pinfo->children, cinfo);
+		/* Treat as atomic */
+		cinfo->parent = 1;
+		LIST_ADD(&init->children, cinfo);
+	}
+
+	if (adopted != 0) {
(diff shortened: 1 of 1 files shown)


posix/posix.c
```diff
@@ -2686,45 +2686,74 @@ int posix_waitpid(pid_t child, int *status, int options)
...
-		posix_sigchild(pinfo->parent);
+	/* Reap all orphaned zombies */
+	while (zombies != NULL) {
+		zinfo = zombies;
+		LIST_REMOVE(&zombies, zinfo);
+		pinfo_put(zinfo);
+	}
 
-		pinfo_put(ppinfo);
+	/* Signal parent if noone was waiting in waitpid() */
```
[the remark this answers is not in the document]
agkaminski: !1

posix/posix.c
```diff
@@ -2686,45 +2686,68 @@ int posix_waitpid(pid_t child, int *status, int options)
...
-				cinfo->parent = 1;
+	/* Make a zombie, wakeup waitpid */
+	proc_lockSet(&ppinfo->lock);
+	LIST_REMOVE(&ppinfo->children, pinfo);
+	if (ppinfo != init) {
+		LIST_ADD(&ppinfo->zombies, pinfo);
+		wait = proc_threadBroadcast(&ppinfo->wait);
+	}
+	else {
+		LIST_ADD(&zombies, pinfo);
```
nalajcie, line 2720: I would add a comment here as I've missed at first that it handles the direct children of `init` to become zombies and leak.

Maybe we should also add comment in `init` process that the zombie reaping is being done in `posix_died` as I probably wouldn't think to look here?

posix/posix.c
```diff
@@ -2686,45 +2686,68 @@ int posix_waitpid(pid_t child, int *status, int options)
...
+	}
+	proc_lockClear(&ppinfo->lock);
 
-				LIST_REMOVE(&pinfo->children, cinfo);
-				LIST_ADD(&init->children, cinfo);
-			}
-			proc_lockClear(&pinfo->lock);
-			proc_lockClear(&init->lock);
-			pinfo_put(init);
+	/* Replant children */
```
nalajcie, line 2725: for orphaned children to find a new parent is called an **adoption** :P

posix/posix.c
```diff
@@ -2686,45 +2686,68 @@ int posix_waitpid(pid_t child, int *status, int options)
...
+	if (children != NULL) {
+		proc_lockSet(&init->lock);
+		while (children != NULL) {
+			cinfo = children;
+			LIST_REMOVE(&children, cinfo);
+			cinfo->parent = 1;
```
nalajcie, line 2731: we've lost the comment about `pid_t` atomicity - might be important as `posix_died` is reentrant (current child might die in our hands) - if pid_t assignment would not be atomic, we might get invalid/mssing `ppinfo` for them.

If we're not sure about pid_t atomicity than maybe we need lock here?

EDIT:
Finally I think there might be some race conditions in reaping/reattaching zombies inside this function:
Assume we have process hierarchy INIT -> PARENT -> CHILD  (no grandchilds)

Scenario 1:
 - PARENT dies, is in `posix_died` in previous line (cinfo points to old parent)
 - CHILD dies, is recognized as zombie, is attached to `PARENT->zombies` list -> memory leak (parent pinfo would be reaped without re-checking zombies list)

Scenario 2:
 - PARENT dies, is in `posix_died` in next line (cinfo points to INIT as a parent)
 - CHILD dies, is recognized as zombie, is reaped at the end of the function -> we have entry in `INIT->children` pointing to deallocated memory

I think we might mitigate the race conditions with additional clever pinfo locking (eg. `pinfo->lock` should probably be held for the full duration of this function to ensure no new entries would be added to `pinfo->zombies`) and/or moving actual reaping to the init process or even making `posix_died` non-reentrant (drastic).

nalajcie (commented):
Haven't tested it but from the reading I see some potential race conditions (probably not significant on non-SMP) but still should be taken care of as they might result in leaked memory or use-after-free.

Your changes make code more readable, but still I had problems with grasping at first why we don't need to reap children in `init` process now - that's why I suggested adding some explicit comment regarding that ;).

posix/posix.c
```diff
@@ -2686,45 +2686,68 @@ int posix_waitpid(pid_t child, int *status, int options)
...
+			LIST_ADD(&init->children, cinfo);
 		}
+		proc_lockClear(&init->lock);
+	}
 
-		posix_sigchild(pinfo->parent);
+	/* Reap all orphaned zombies */
```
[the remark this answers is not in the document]
agkaminski: Init now never has any zombies, we kill children on sight

posix/posix.c
agkaminski: Ok, I'll add comment here

posix/posix.c
agkaminski: You're right, thanks for spotting this.

Now I **replant** children and zombies in one operation, so race condition is no longer possible. Sadly, it means that proc_lockSet2 is needed again, but it doesn't hurt anything, I just wanted to avoid it, if possible

agkaminski: Ok, one race is still possible, we might be adopted mid-death, will think about it

agkaminski: Well, that was difficult, but I think I managed

posix/posix.c
```diff
@@ -2686,43 +2686,72 @@ int posix_waitpid(pid_t child, int *status, int options)
...
+	/* We might not find a parent if it died just now */
 	if (ppinfo != NULL) {
+		/* Make a zombie, wakeup waitpid */
 		proc_lockSet(&ppinfo->lock);
-		LIST_REMOVE(&ppinfo->children, pinfo);
-		LIST_ADD(&ppinfo->zombies, pinfo);
-		proc_threadBroadcast(&ppinfo->wait);
+		/* Check if we didn't get adopted by the init in the meantime */
+		if ((ppinfo != init) && (ppinfo->children != NULL)) {
+			if (LIST_BELONGS(&ppinfo->children, pinfo) != 0) {
```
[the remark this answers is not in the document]
agkaminski: Right, thanks, changed

badochov (approved):
LGTM

lukileczo (approved):

PR #463, opened 2023-10-06
Use idtree

agkaminski:
## Description

Replace multiple implementations of interval tree for ID allocations with one universal one.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/proc/cond.c
+++ b/proc/cond.c
@@ -27,7 +27,7 @@ cond_t *cond_get(int c)
 	thread_t *t = proc_current();
 	resource_t *r = resource_get(t->process, c);
 	LIB_ASSERT((r == NULL) || (r->type == rtCond), "process: %s, pid: %d, tid: %d, handle: %d, resource type mismatch",
-		t->process->path, t->process->id, t->id, c);
+		t->process->path, process_getPid(t->process), t->id, c);
 	return ((r != NULL) && (r->type == rtCond)) ? r->payload.cond : NULL;
 }
 
@@ -38,11 +38,11 @@ void cond_put(cond_t *cond)
 	int rem;
 
 	LIB_ASSERT(cond != NULL, "process: %s, pid: %d, tid: %d, cond == NULL",
-		t->process->path, t->process->id, t->id);
+		t->process->path, process_getPid(t->process), t->id);
 
 	rem = resource_put(t->process, &cond->resource);
 	LIB_ASSERT(rem >= 0, "process: %s, pid: %d, tid: %d, refcnt below zero",
-		t->process->path, t->process->id, t->id);
+		t->process->path, process_getPid(t->process), t->id);
 	if (rem <= 0) {
 		proc_threadBroadcastYield(&cond->queue);
 		vm_kfree(cond);
--- a/proc/cond.c
+++ b/proc/cond.c
@@ -27,7 +27,7 @@ cond_t *cond_get(int c)
 	thread_t *t = proc_current();
 	resource_t *r = resource_get(t->process, c);
 	LIB_ASSERT((r == NULL) || (r->type == rtCond), "process: %s, pid: %d, tid: %d, handle: %d, resource type mismatch",
-		t->process->path, process_getPid(t->process), t->id, c);
+		t->process->path, process_getPid(t->process), proc_getTid(t), c);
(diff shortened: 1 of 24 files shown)


proc/threads.c
```diff
@@ -855,6 +749,12 @@ int proc_threadCreate(process_t *process, void (*start)(void *), unsigned int *i
...
 	t->lastTime = t->startTime;
 
+	if (thread_alloc(t) < 0) {
+		vm_kfree(t->kstack);
+		vm_kfree(t);
+		return -ENOMEM;
+	}
+
 	if (process != NULL && (process->tls.tdata_sz != 0 || process->tls.tbss_sz != 0)) {
 		err = process_tlsInit(&t->tls, &process->tls, process->mapp);
```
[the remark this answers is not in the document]
agkaminski: Oops, we should indeed, thx

lukileczo:
One more thing: why the exclamation mark in commit message?

lukileczo (dismissed):
LGTM

proc/cond.c
```diff
@@ -5,115 +5,129 @@
...
 
 
-cond_t *cond_get(unsigned int c)
+void cond_put(cond_t *cond)
 {
-	return resourceof(cond_t, resource, resource_get(proc_current()->process, rtCond, c));
+	thread_t *t = proc_current();
+	int rem = resource_put(t->process, &cond->resource);
+	LIB_ASSERT(rem >= 0, "process: %s, pid: %d, tid: %d, refcnt below zero",
+		t->process->path, process_getPid(t->process), proc_getTid(t));
```
kemonats, line 40: What do you think about adding a check if the pointer `cond == NULL` at the begining of this function and removing check before calling this function.
Missing empty line after the declaration of local variables.
```suggestion
	thread_t *t;
	int rem;

	if (cond == NULL) {
		return;
	}
	t = proc_current();
	rem = resource_put(t->process, &cond->resource);
	LIB_ASSERT(rem >= 0, "process: %s, pid: %d, tid: %d, refcnt below zero",
		t->process->path, process_getPid(t->process), proc_getTid(t));
```

kemonats: But you can remove the check in the `userintr.c` file ( for times ) such as this:
```c
		if (cond != NULL) {
			cond_put(cond);
		}
```

agkaminski: I'm aware, but it's a matter of - do we allow to `put` a resource you never had? IMHO it's a bad practice to do so and it makes a code using it somewhat less clear

proc/threads.c
```diff
@@ -85,37 +85,14 @@ static int threads_sleepcmp(rbnode_t *n1, rbnode_t *n2)
...
-		return -1;
-
-	else if (t1->id > t2->id)
-		return 1;
-
-	return 0;
+	if (t1->wakeup != t2->wakeup) {
+		return (t1->wakeup > t2->wakeup) ? 1 : -1;
+	}
+	else {
```
lukileczo, line 91: Unnecessary `else`

proc/cond.c
```diff
@@ -5,115 +5,134 @@
...
+	if (rem <= 0) {
+		proc_threadBroadcastYield(&cond->queue);
+		vm_kfree(cond);
+	}
 }
 
 
 int proc_condCreate()
```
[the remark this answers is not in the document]
agkaminski: Fixed

proc/threads.c
agkaminski: Thanks for finding this, fixed

proc/threads.c
```diff
@@ -1128,6 +1023,19 @@ static void _proc_threadDequeue(thread_t *t)
 }
 
 
+static void _proc_threadInsertSleeping(thread_t *t)
```
lukileczo, line 1026: Hmm, I'm not convinced that solves the problem: `lib_rbInsert` can only return not `EOK` if `tree->compare` (in this case `threads_sleepcmp`) returns 0 - which it doesn't. So we'd still insert thread with the same `wakeup`

agkaminski: Right, I've reverted the change

lukileczo (approved):
LGTM

jmaksymowicz (approved):
LGTM

PR #466, opened 2023-10-21
vm/map: replace dead ends with LIB_ASSERT_ALWAYS

gerard5:
## Description

The change replaces dead ends like `lib_printf(...); for (;;);` by using `LIB_ASSERT_ALWAYS` in places where there is no chance for klog reader to print an appropriate panic message to the debug console.  This PR additionally checks all uses of `for (;;)` and `while (1)` loops to detect potential missing log-useful information, and normalizes its usage to the MISRA 2012 rules.

Changes are split into separate commits.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `imxrt1176-nil`, `ia32-generic`

## Checklist:

## Special treatment

### DIFF
--- a/vm/map.c
+++ b/vm/map.c
@@ -1145,6 +1145,7 @@ static int _map_mapsInit(vm_map_t *kmap, vm_object_t *kernel, void **bss, void *
 {
 #ifdef NOMMU
 	size_t mapsCnt, id = 0;
+	int result;
 
 	map_entry_t *entry;
 	const mapent_t *sysEntry;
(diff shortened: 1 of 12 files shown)


vm/map.c
```diff
@@ -1156,10 +1156,8 @@ static int _map_mapsInit(vm_map_t *kmap, vm_object_t *kernel, void **bss, void *
 	map_common.maps = (vm_map_t **)(*bss);
 
 	while ((*top) - (*bss) < sizeof(vm_map_t *) * mapsCnt) {
-		if (_page_sbrk(&map_common.kmap->pmap, bss, top) < 0) {
-			lib_printf("vm: Problem with extending kernel heap for vm_map_t pool (vaddr=%p)\n", *bss);
-			for (;;);
-		}
+		LIB_ASSERT_ALWAYS(_page_sbrk(&map_common.kmap->pmap, bss, top) == EOK,
```
agkaminski, line 1159: Assert condition shouldn't have any side effects

agkaminski (changes requested):

vm/map.c
gerard5: replaced with `lib_assertPanic()`

lib/assert.h
```diff
@@ -20,6 +20,9 @@
 void lib_assertPanic(const char *func, int line, const char *fmt, ...);
 
 
+#define LIB_ASSERT_PANIC(fmt, ...) \
```
agkaminski, line 23: I have two problems with this macro:
- it's not really needed. You can do
```
ret = func();
assert(ret >= 0, "error");
```
instead of
```
if (func() < 0)
 assert("error");
```
- it breaks assert meaning - it's an assert without a condition

gerard5: Done.

agkaminski (approved):

PR #468, opened 2023-10-21
Apply NDEBUG logic in lib_assertPanic()

gerard5:
## Description

Applies `NDEBUG` logic [consistent with that in exceptions handling function](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/master/hal/armv7m/exceptions.c#L100-L107) to `lib_assertPanic()`, if it occurs, depending on `NDEBUG` is defined or not the system will reset or halt. This change affects the operation of the `LIB_ASSERT_ALWAYS()` macro, but is neutral for regular `LIB_ASSERT()` which is `no-op` when `NDEBUG` is defined.

JIRA: RTOS-658

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `imxrt1176-nil`, `stm32l4a6-nucleo`

## Checklist:

## Special treatment

### DIFF
--- a/lib/assert.c
+++ b/lib/assert.c
@@ -1,4 +1,3 @@
-
 /*
  * Phoenix-RTOS
  *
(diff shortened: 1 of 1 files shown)


agkaminski (approved):
I was sure this was already the case, thanks for the fix

PR #471, opened 2023-11-02
mmap: make interface posix compliant

badochov:
JIRA: RTOS-665

## Description
Adjust mmap to POSIX standard.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/libphoenix/pull/306

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -729,6 +729,24 @@ ssize_t posix_write(int fildes, void *buf, size_t nbyte)
 }
 
 
+int posix_getOid(int fildes, oid_t *oid)
+{
+	open_file_t *f;
+	int err;
+
+	err = posix_getOpenFile(fildes, &f);
+	if (err < 0) {
+		return err;
+	}
+
+	hal_memcpy(oid, &f->oid, sizeof(oid_t));
+
+	posix_fileDeref(f);
+
+	return EOK;
+}
+
+
 int posix_dup(int fildes)
 {
 	TRACE("dup(%d)", fildes);
--- a/proc/msg.c
+++ b/proc/msg.c
@@ -22,7 +22,10 @@
 #define CEIL(x)  (((x) + SIZE_PAGE - 1) & ~(SIZE_PAGE - 1))
 
 
+/* clang-format off */
 enum { msg_rejected = -1, msg_waiting = 0, msg_received, msg_responded };
+/* clang-format on */
+
 
 
 struct {
@@ -115,7 +118,7 @@ static void *msg_map(int dir, kmsg_t *kmsg, void *data, size_t size, process_t *
 			return NULL;
 		}
 
-		vaddr = vm_mmap(msg_common.kmap, NULL, NULL, SIZE_PAGE, PROT_READ | PROT_WRITE, OID_PHYSMEM, bpa, flags);
+		vaddr = vm_mmap(msg_common.kmap, NULL, NULL, SIZE_PAGE, PROT_READ | PROT_WRITE, (void *)-1, bpa, flags);
 		ml->bvaddr = vaddr;
 		if (vaddr == NULL) {
 			return NULL;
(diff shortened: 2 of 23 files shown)


badochov:
Project branch with all changes to confirm it works:
https://github.com/phoenix-rtos/phoenix-rtos-project/pull/909

proc/msg.c
```diff
@@ -115,7 +118,7 @@ static void *msg_map(int dir, kmsg_t *kmsg, void *data, size_t size, process_t *
 			return NULL;
 		}
 
-		vaddr = vm_mmap(msg_common.kmap, NULL, NULL, SIZE_PAGE, PROT_READ | PROT_WRITE, OID_PHYSMEM, bpa, flags);
+		vaddr = vm_mmap(msg_common.kmap, NULL, NULL, SIZE_PAGE, PROT_READ | PROT_WRITE, (void *)-1, bpa, flags);
```
agkaminski, line 121: Perhaps move `OID_PHYSMEM` to vm/map.h instead of removing it? Internal vm_mmap API hasn't changed

vm/map.c
```diff
@@ -441,16 +441,17 @@ int _vm_munmap(vm_map_t *map, void *vaddr, size_t size)
...
+void *_vm_mmap(vm_map_t *map, void *vaddr, page_t *p, size_t size, u8 prot, vm_object_t *o, off_t offs, u8 flags)
 {
 	int attr = PROT_NONE;
 	void *w;
 	process_t *process = NULL;
 	thread_t *current;
 	map_entry_t *e;
 
-	if (!size || (size & (SIZE_PAGE - 1)))
+	if ((size & (SIZE_PAGE - 1)) != 0) {
```
agkaminski, line 452: Logic has changed - previously we return NULL if size == 0, now we don't

syscalls.c
```diff
@@ -52,39 +52,49 @@ void *syscalls_mmap(void *ustack)
...
-		o = NULL;
-	}
-	else if (oid == OID_PHYSMEM) {
-		o = (void *)-1;
-	}
-	else if (oid == OID_CONTIGUOUS) {
-		o = vm_objectContiguous(size);
-		if (o == NULL) {
-			return (void *)-1;
+	if (fildes == -1) {
```
agkaminski, line 67: After some thought: perhaps it would be safer to use MAP_ANONYMOUS instead of fd as an indicator?

That is if we decide to always combine MAP_PHYSMEM and MAP_CONTIGUOUS with MAP_ANONYMOUS

proc/msg.c
badochov: But this argument isn't oid but vm_object_t

syscalls.c
agkaminski: fd = -1 is not always needed, some external app might not adhere to this rule

proc/msg.c
agkaminski: Yes, `OID_` prefix needs to go, e.g. `VM_OBJ_PHYSMEM`?

syscalls.c
badochov: Ok

posix/posix.c
```diff
@@ -729,6 +729,24 @@ ssize_t posix_write(int fildes, void *buf, size_t nbyte)
 }
 
 
+int posix_get_oid(int fildes, oid_t *oid)
```
agkaminski, line 732:
```suggestion
int posix_getOid(int fildes, oid_t *oid)
```

vm/map.c
```diff
@@ -673,7 +676,7 @@ static void map_pageFault(unsigned int n, exc_context_t *ctx)
 
 #ifdef PAGEFAULTSTOP
 	process_dumpException(n, ctx);
-	__asm__ volatile ("1: b 1b");
+	__asm__ volatile("1: b 1b");
```
agkaminski, line 679: clang-format is mistaken

badochov: About that we are not concitend in the code as sometimes there is space and sometimes there's not that's why i decided to stay with clang-format but I'll change it back

agkaminski (approved):

PR #473, opened 2023-11-08
Enable SMP on GR712RC

lukileczo:
## Description

This PR:
- extracts common TLB interface in HAL and adapts `ia32` port to work with the changes
- enables SMP on GR712RC
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/sparcv8leon3/gaisler/gr712rc/_init.S
+++ b/hal/sparcv8leon3/gaisler/gr712rc/_init.S
@@ -67,6 +67,13 @@ _init:
 
 	wr %g0, PSR_S, %psr
 
+	/* Get CPU ID */
+	rd %asr17, %g1
+	srl %g1, 28, %g1
+	cmp %g1, %g0
+	bnz _init_core
+	nop
+
 	/* %g2 = syspage pa (from plo) */
 	set PHY_ADDR(syspage), %g3
 	set_page_align VADDR_SYSPAGE, %g1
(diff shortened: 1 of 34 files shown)


hal/sparcv8leon3/gaisler/gr712rc/_init.S
```diff
@@ -220,3 +227,71 @@ set_pdir3:
...
+	/* Enable interrupts and wait to be scheduled */
+	mov %psr, %g1
+	andn %g1, (PSR_PIL), %g1
+	wr %g1, %psr
+	nop
+	nop
+	nop
+
+	/* GR712RC errata 1.7.8 */
+	set 0xfffffff0, %g1
```
agkaminski, line 291: Maybe we should do this before enabling interrupts? I guess there's a chance we might never execute this instruction

agkaminski (commented):
One small doubt on my part, otherwise LGTM. I like the TLB code refactor very much

hal/sparcv8leon3/gaisler/gr712rc/_init.S
lukileczo: It doesn't really matter - this is only needed if we execute the `wr %g0, %asr19` (which basically is a `wfi` instruction).
From manual:
> the first instruction or data (whichever is first) access to the bus after leaving power-down might be performed incorrectly
> the load instruction after the write to %asr19 provides immunity against a data cache failure due to
the error described by this errata

Basically the first access after power-down must not be cached.

agkaminski: Ok, now I understand, thanks

agkaminski (approved):
LGTM

PR #479, opened 2023-11-22
hal/ia32: Refactor timers

astalke:
## Description

Make timers more modular and add displaying description string about them.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## How Has This Been Tested?

- [ ] Already covered by automatic testing.
- [ ] New test added: (add PR link here).
- [x] Tested by hand on: ia32-generic-qemu, ia32-generic-pc

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing linter checks and tests passed.
- [ ] My changes generate no new compilation warnings for any of the targets.

## Special treatment

- [ ] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
- [ ] I will merge this PR by myself when appropriate.

PR #485, opened 2023-12-01
hal/ia32: Fix potential buffer overflow in hal_timerFeatures()

astalke:
## Description

Fixes bugs reported by @kemonats  here https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/479#issuecomment-1830501494

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic-qemu

## Checklist:

## Special treatment

### DIFF
--- a/hal/ia32/timer.c
+++ b/hal/ia32/timer.c
@@ -92,28 +92,32 @@ struct {
 } timer_common;
 
 
-/* Programmable Interval Timer (Intel 8253/8254) */
-
-static unsigned int _hal_pitName(char *s, unsigned int *len)
+static unsigned int _hal_timersName(char *s, const char *prefix, unsigned long val, const char *suffix, unsigned int *len)
 {
-	static const char suffix[] = "kHz)";
-	unsigned int off = 0, n;
-
-	n = hal_i2s("Programmable Interval Timer (", s + off, PIT_FREQUENCY, 10, 0);
+	unsigned int off = 0, n = hal_strlen(prefix) + sizeof(val) * 10 / 4;
 	if (*len < n) {
-		n = *len;
+		return off;
 	}
+	n = hal_i2s(prefix, s + off, val, 10, 0);
 	off += n;
 	*len -= n;
 
 	hal_strncpy(s + off, suffix, *len);
-	n = sizeof(suffix) - 1;
+	n = hal_strlen(suffix);
 	if (*len < n) {
 		n = *len;
 	}
 	off += n;
 	*len -= n;
 	return off;
+}
+
+
+/* Programmable Interval Timer (Intel 8253/8254) */
+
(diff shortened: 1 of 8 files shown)


agkaminski (approved):

hal/ia32/timer.c
```diff
@@ -310,22 +300,7 @@ static int _hal_lapicTimerIrqHandler(unsigned int n, cpu_context_t *ctx, void *a
...
-	off += n;
-	*len -= n;
-
-	hal_strncpy(s + off, suffix, *len);
-	n = sizeof(suffix) - 1;
-	off += n;
-	*len -= n;
-
-	return off;
+	return _hal_timersName(s, "High Precision Timer (", (u32)(1000000000000LLU / ((u64)timer_common.hpetData.period)), "kHz)", len);
```
gerard5, line 303: ULL, instead of LLU ?

agkaminski (approved):

hal/ia32/timer.c
```diff
@@ -92,22 +92,21 @@ struct {
...
-
-	n = hal_i2s("Programmable Interval Timer (", s + off, PIT_FREQUENCY, 10, 0);
+	unsigned int off = 0, n = hal_strlen(prefix) + 10; /* 10 should be enough space for val */
+	if (*len < n) {
+		return off;
+	}
+	n = hal_i2s(prefix, s + off, val, 10, 0);
 	if (*len < n) {
 		n = *len;
 	}
```
kemonats, line 104:
```suggestion
```
This is unnecessary, you check it beforehand.

gerard5 (approved):
Looks good to me.

agkaminski (approved):

kemonats:
This PR has been merged but the problem remains.
I will write again.
If after calling:
```c
        off += timer_common.timestampTimer->name(features + off, &len);
```
variable **`len == 0`** ( that is, there is not enough space in **`features`** ) and then **`off`** is equal to the value **`len`** argument in the function call and to add a **`nul`** character at the end of **`features`** the value of **`off`**  must be decreased by **1**. 
```c
	if (len == 0) {
		off--;
	}
	features[off] = '\0';
```

Another solution is to leave only pre-existing line:
```c
	(features + off)[len - 1] = '\0';
```
without:
```c
	features{off] = '\0';
```
and then with **`len == 0`** it would be like this:
```c
	(features + off)[(unsigned int)-1] = '\0';
```

Probably a better solution ( more redable ) would be to increase only the **`off`** variable and call the function with the argument **`len - off`** instead **`&len`**, and then there would be a line:
```c
        features[len - 1] = '\0';
```


Adding at the begining of the function:
```c
        if (len == 0) {
                return features;
        }
```
only protects against calling:
```c
hal_timerFeatures(features, 0)
```

PR #497, opened 2023-12-15
process: align stack at process start

badochov:
Change semantics of last argument. And fix stack alignment

JIRA: RTOS-720

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/libphoenix/pull/327

### DIFF
--- a/hal/armv7a/arch/cpu.h
+++ b/hal/armv7a/arch/cpu.h
@@ -49,18 +49,14 @@
 #define SYSTICK_INTERVAL 1000
 
 
-#define PUTONSTACK(kstack, t, v) \
-	do { \
-		(kstack) -= (sizeof(t) + 3) & ~0x3; \
-		*((t *)kstack) = (v); \
-	} while (0)
+#define SIZE_STACK_ARG(sz) (((sz) + 3u) & ~0x3u)
 
 
 #define GETFROMSTACK(ustack, t, v, n) \
 	do { \
 		ustack = (void *)(((addr_t)ustack + sizeof(t) - 1) & ~(sizeof(t) - 1)); \
 		(v) = *(t *)ustack; \
-		ustack += (sizeof(t) + 3) & ~0x3; \
+		ustack += SIZE_STACK_ARG(sizeof(t)); \
 	} while (0)
 
 typedef struct _cpu_context_t {
--- a/hal/armv7m/stm32/_init.S
+++ b/hal/armv7m/stm32/_init.S
@@ -320,31 +320,32 @@ hal_cpuReschedule0:
 .ltorg
 
 
-.globl hal_jmp /* void hal_jmp(void *f, void *kstack, void *stack, int argc) */
+.globl hal_jmp /* void hal_jmp(void *f, void *kstack, void *ustack, int kargc, const arg_t *kargs) */
 .type hal_jmp, %function
 hal_jmp:
-	push {r0-r3}
-	pop {r4-r7}
-
 	cpsid if
 	isb
 
-	cmp r2, #NULL
+	push {r0-r3}
+	pop {r4-r8} /* kargs has been passed on the stack */
+
+	cmp r6, #NULL
 	bne hal_jmp_user
 
 	mov sp, r5
+	isb
 	subs r7, #1
 	bmi 1f
-	pop {r0}
+	ldr r0, [r8]
 	subs r7, #1
 	bmi 1f
-	pop {r1}
+	ldr r1, [r8, #4]
 	subs r7, #1
 	bmi 1f
-	pop {r2}
+	ldr r2, [r8, #8]
 	subs r7, #1
 	bmi 1f
-	pop {r3}
+	ldr r3, [r8, #12]
 1:
 	cpsie if
 	isb
@@ -353,24 +354,13 @@ hal_jmp:
 
 hal_jmp_user:
 	cpsid if
+	isb
 	msr msp, r5
-	subs r7, #1
-	bmi 2f
-	pop {r0}
-	subs r7, #1
-	bmi 2f
-	pop {r1}
-	subs r7, #1
-	bmi 2f
-	pop {r2}
-	subs r7, #1
-	bmi 2f
-	pop {r3}
-2:
+	isb
 	msr psp, r6
 	cpsie if
 	isb
-	mov r5, #0x3
+	mov r5, USERCONTROL
 	msr control, r5
 	isb
 	dmb
--- a/hal/armv8m/nrf/_init.S
+++ b/hal/armv8m/nrf/_init.S
@@ -316,31 +316,32 @@ hal_cpuReschedule0:
 .ltorg
 
 
-.globl hal_jmp /* void hal_jmp(void *f, void *kstack, void *stack, int argc) */
+.globl hal_jmp /* void hal_jmp(void *f, void *kstack, void *ustack, int kargc, const arg_t *kargs) */
 .type hal_jmp, %function
 hal_jmp:
-	push {r0-r3}
-	pop {r4-r7}
-
 	cpsid if
 	isb
 
-	cmp r2, #NULL
+	push {r0-r3}
+	pop {r4-r8} /* kargs has been passed on the stack */
+
+	cmp r6, #NULL
 	bne hal_jmp_user
 
 	mov sp, r5
+	isb
 	subs r7, #1
 	bmi 1f
-	pop {r0}
+	ldr r0, [r8]
 	subs r7, #1
 	bmi 1f
-	pop {r1}
+	ldr r1, [r8, #4]
 	subs r7, #1
 	bmi 1f
-	pop {r2}
+	ldr r2, [r8, #8]
 	subs r7, #1
 	bmi 1f
-	pop {r3}
+	ldr r3, [r8, #12]
 1:
 	cpsie if
 	isb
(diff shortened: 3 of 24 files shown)


proc/process.c
```diff
@@ -1068,10 +1068,15 @@ static void *process_putargs(void *stack, char ***argsp, int *count)
 }
 
 
+#define STACK_ARGSZ (STACK_SIZE(char **) + STACK_SIZE(char **) + STACK_SIZE(int) + STACK_SIZE(void *))
```
agkaminski, line 1071: I rather remove this macro, because:
- it's used exactly once, we have a doctrine of avoiding macros is feasible,
- its value is detached from it's meaning - it's usage defines why there are values that there are.

proc/process.c
```diff
@@ -1068,10 +1068,15 @@ static void *process_putargs(void *stack, char ***argsp, int *count)
...
+
+
 static void process_exec(thread_t *current, process_spawn_t *spawn)
 {
 	void *stack, *entry;
+	ptr_t stackp;
```
agkaminski, line 1077:
```suggestion
```

hal/armv7a/arch/cpu.h
```diff
@@ -49,9 +49,15 @@
 #define SYSTICK_INTERVAL 1000
 
 
+#define STACK_ALIGN 8
```
agkaminski, line 52: This macro can be avoided by just assuming highest for all, i.e. 16. No harm in this.

hal/armv7a/arch/cpu.h
```diff
@@ -49,9 +49,15 @@
...
+
+
+#define STACK_SIZE(t) ((sizeof(t) + 3u) & ~0x3u)
```
agkaminski, line 55: Name is kinda misleading (suggest size of the whole stack, not it's element), in kernel we have a convention that attribute is first in the name (see SIZE_KSTACK, SIZE_PAGE etc)
```suggestion
#define SIZE_STACK_ARG(t) ((sizeof(t) + 3u) & ~0x3u)
```

hal/armv7a/arch/cpu.h
```diff
@@ -49,9 +49,15 @@
...
+
+
 #define PUTONSTACK(kstack, t, v) \
 	do { \
-		(kstack) -= (sizeof(t) + 3) & ~0x3; \
+		(kstack) = (char *)(kstack) - STACK_SIZE(t); \
```
agkaminski, line 60: To avoid the warning - perhaps we should always pass `kstack` casted to `char *` to avoid UB on the other hand? Or leave this line as is, but fix usage where `kstack` is `u32`. I guess the second option would be better
```suggestion
		(kstack) -= STACK_SIZE(t); \
```

agkaminski (commented):
Take a look on warnings

hal/armv7a/arch/cpu.h
badochov: At first I thought about doing it like this, however, using the highest value still requires a macro or an extensive comment for readability and I think this kind of fine grained control doesn't introduce too much complication.

badochov:
@agkaminski it works on all architecuters but SPARC @lukileczo will have a look at in on Friday

hal/armv7a/arch/cpu.h
badochov: I;ve decided to assume highest of all for coherency of ABI sake

hal/common.c
```diff
@@ -0,0 +1,44 @@
...
+
+
+void hal_stackPutArgs(void **stackp, size_t argc, const struct stackArg *argv)
+{
+	size_t i, misalign, argsz = 0;
+	ptr_t stack = (ptr_t)*stackp;
+
+	for (i = 0; i < argc; i++) {
+		argsz += SIZE_STACK_ARG(argv[i].sz);
+	}
```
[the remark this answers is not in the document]
badochov: Thanks for spotting SIZE_STACK_ARG was wrong

badochov:
@agkaminski to work on sparc update also in libphoenix is needed. However after both changes are applied @lukileczo confirmed that it works

hal/common.c
```diff
@@ -0,0 +1,44 @@
...
+
+	misalign = (STACK_ALIGN + stack - argsz) & (STACK_ALIGN - 1);
```
kemonats, line 35: adding `STACK_ALIGN` is unnecessary, the compiler ignores it.
```suggestion
	misalign = (stack - argsz) & (STACK_ALIGN - 1);
```

hal/armv7m/stm32/_init.S
```diff
@@ -320,31 +320,32 @@ hal_cpuReschedule0:
 .ltorg
 
 
-.globl hal_jmp /* void hal_jmp(void *f, void *kstack, void *stack, int argc) */
+.globl hal_jmp /* void hal_jmp(void *f, void *kstack, void *ustack, int kargc, const arg_t *kargs) */
 .type hal_jmp, %function
 hal_jmp:
 	push {r0-r3}
-	pop {r4-r7}
+	pop {r4-r8}
```
agkaminski, line 327: As in imxrt, imho move interrupt disable to the start

hal/armv8m/nrf/_init.S
```diff
@@ -316,31 +316,32 @@ hal_cpuReschedule0:
 .ltorg
 
 
-.globl hal_jmp /* void hal_jmp(void *f, void *kstack, void *stack, int argc) */
+.globl hal_jmp /* void hal_jmp(void *f, void *kstack, void *ustack, int kargc, const arg_t *kargs) */
 .type hal_jmp, %function
 hal_jmp:
 	push {r0-r3}
-	pop {r4-r7}
+	pop {r4-r8}
```
agkaminski, line 323: As in imxrt, imho move interrupt disable to the start

hal/ia32/cpu.c
```diff
@@ -255,10 +258,25 @@ void hal_longjmp(cpu_context_t *ctx)
...
 
 
-void hal_jmp(void *f, void *kstack, void *stack, int argc)
+void hal_jmp(void *f, void *kstack, void *ustack, size_t kargc, const arg_t *kargv)
 {
+	/* We support passing at most 4 args on every architecture. */
+	struct stackArg args[4];
+	size_t i;
+
+	if (kargc > 4) {
```
agkaminski, line 267: Perhaps an assert too, so it will be easier to find out when inevitably someone pass more than 4?

proc/process.c
```diff
@@ -1092,12 +1093,14 @@ static void process_exec(thread_t *current, process_spawn_t *spawn)
...
 		stack = process_putargs(stack, &spawn->envp, &count);
 		stack = process_putargs(stack, &spawn->argv, &count);
-
-		/* temporary? put arguments to main on stack */
-		PUTONSTACK(stack, char **, spawn->envp);
-		PUTONSTACK(stack, char **, spawn->argv);
-		PUTONSTACK(stack, int, count);
-		PUTONSTACK(stack, void *, NULL); /* return address */
+		/* clang-format off */
+		hal_stackPutArgs(&stack, 4, (const struct stackArg[]) {
```
agkaminski, line 1097: Let's keep kernel somewhat conservative and use intermediate const var for this struct ;)

proc/process.c
```diff
@@ -1295,10 +1298,11 @@ void proc_exit(int code)
 		kstack = current->kstack = current->execkstack;
 		kstack += current->kstacksz;
 
-		PUTONSTACK(kstack, int, FORKED);
-		PUTONSTACK(kstack, process_spawn_t *, spawn);
-		PUTONSTACK(kstack, thread_t *, current);
-		hal_jmp(proc_vforkedExit, kstack, NULL, 3);
+		hal_jmp(proc_vforkedExit, kstack, NULL, 3, (const arg_t[]) {
```
agkaminski, line 1301: As in process_exex, let's use intermediate var.
Also, this formatting doesn't work that good

hal/ia32/cpu.c
badochov: I wanted to but iirc lib should not be used in hal

hal/sparcv8leon3/_traps.S
```diff
@@ -1127,14 +1127,14 @@ hal_jmp:
...
-	ld [%fp + 4], %o1
+	ld [%i4 + 4], %o1
 	subcc %i3, 1, %i3
 	bneg 2f
-	ld [%fp + 8], %o2
-	ld [%fp + 12], %o3
+	ld [%i4 + 8], %o2
+	ld [%i4 + 12], %o3
 2:
 	andn %fp, 0x7, %fp
```
lukileczo, line 1139: Now that the stacks are aligned, this can be also removed

proc/process.c
```diff
@@ -1092,12 +1099,9 @@ static void process_exec(thread_t *current, process_spawn_t *spawn)
...
 		stack = process_putargs(stack, &spawn->argv, &count);
-
-		/* temporary? put arguments to main on stack */
-		PUTONSTACK(stack, char **, spawn->envp);
-		PUTONSTACK(stack, char **, spawn->argv);
-		PUTONSTACK(stack, int, count);
-		PUTONSTACK(stack, void *, NULL); /* return address */
+		/* clang-format off */
+		hal_stackPutArgs(&stack, sizeof(args) / sizeof(args[0]), args);
+		/* clang-format on */
```
lukileczo, line 1104: Unnecessary `clang-format-off/on` here

lukileczo (approved):
LGTM

agkaminski (approved):

PR #502, opened 2024-01-03
riscv64: add FPU context stacking

lukileczo:
## Description

Add FPU handling on riscv64.
## Motivation and Context

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/988
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `riscv64-generic-qemu`, `riscv64-generic-spike`

## Checklist:

## Special treatment

### DIFF
--- a/hal/riscv64/_interrupts.S
+++ b/hal/riscv64/_interrupts.S
@@ -35,9 +35,9 @@
 	beqz a0, 1f
 
 	/* Kernel mode */
-	sd sp, -8(sp)
+	sd sp, -272(sp)
 
-	addi sp, sp, -296
+	addi sp, sp, -CPU_CTX_SIZE
 	j 2f
 
 1:
@@ -47,10 +47,10 @@
 	ld a0, hal_cpuKernelStack
 
 	/* Save task's stack pointer */
-	sd sp, -8(a0)
+	sd sp, -272(a0)
 
 	/* Swap to kernel stack */
-	addi sp, a0, -296
+	addi sp, a0, -CPU_CTX_SIZE
 
 2:
 	/* restore a0 */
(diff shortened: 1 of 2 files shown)


hal/riscv64/_interrupts.S
```diff
@@ -118,6 +168,50 @@
 	andi a0, a0, ~SSTATUS_SIE
 	csrw sstatus, a0
 
+	/* Check FPU status */
+	srl t0, a0, 13
+	andi t0, t0, 3
+	li t1, 2       /* FPU clean */
+	bne t0, t1, 1f
```
agkaminski, line 175: I think we should restore FPU context based  on the fact it has been stored for thread at hand. Consider:

1) thread A comes into the kernel with dirty FPU - ctx is saved,
2) thread B is scheduled, FPU is clean, so no ctx restore,
3) thread B comes back into the kernel with dirty FPU - ctx is saved,
4) thread A is scheduled, FPU state is clean (kernel doesn't use it), so no context restore.

This way now thread A has FPU context of thread B.

Or am I missing something?

lukileczo: It's the other way: we restore FP ctx only if status is clean. But that is also currently incorrect. Code was written according to RISC-V spec https://riscv.org/wp-content/uploads/2017/05/riscv-privileged-v1.10.pdf#page=35, but now I think it must assume we have some dedicated space for storing context, as we'd restore state that wasn't saved. I'll rewrite it a little.

hal/riscv64/_interrupts.S
```diff
@@ -196,6 +288,112 @@ _interrupts_notIrq:
...
+	 * 0000111 - LOAD-FP
+	 * 0100111 - STORE-FP
+	 * 1000011 - MADD
+	 * 1000111 - MSUB
+	 * 1001011 - NMSUB
+	 * 1001111 - NMADD
+	 * 1010011 - OP-FP
+	 */
+	
+	xori t1, t0, 0b1010011 /* OP-FP */
```
agkaminski, line 321: `0b` is a GNU extension, better to use plain hex

agkaminski (approved):

PR #514, opened 2024-01-31
posix: signal SIGPIPE where necessary

### LINKED ISSUE #733: Lack of sending SIGPIPE signal
From Open Group `write()` website:

image

`errno` is set correctly but signal is not delivered.

Reproduction:
```c
#include <stdio.h>
#include <unistd.h>
#include <signal.h>
#include <errno.h>


void handler(int signum) {
	printf("SIGPIPE recieved\n");
}


int main(void)
{
	int pfd[2];

	pipe(pfd);
	close(pfd[0]);

	signal(SIGPIPE, handler);

	write(pfd[1], "x", 1);
	printf("errno: %d\n", errno);

	return 0;
}
```

lukileczo:
## Description

Send `SIGPIPE` where POSIX requires it.

Depends-On: phoenix-rtos-tests:adamdebek/CI-694
## Motivation and Context

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/733
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -783,6 +783,14 @@ ssize_t posix_read(int fildes, void *buf, size_t nbyte, off_t offset)
 }
 
 
+static void posix_sigpipe(void)
+{
+	thread_t *curr = proc_current();
+
+	(void)threads_sigpost(curr->process, curr, SIGPIPE);
+}
+
+
 ssize_t posix_write(int fildes, void *buf, size_t nbyte, off_t offset)
 {
 	TRACE("write(%d, %p, %zu, %jd)", fildes, buf, nbyte, (intmax_t)offset);
@@ -827,6 +835,17 @@ ssize_t posix_write(int fildes, void *buf, size_t nbyte, off_t offset)
 			f->offset += rcnt;
 			(void)proc_lockClear(&f->lock);
 		}
+	}
+
+	if ((rcnt == -EPIPE) && ((f->type == ftUnixSocket) || (f->type == ftPipe) || (f->type == ftFifo) || (f->type == ftInetSocket))) {
+		/* NOTE: for a UNIX socket, SIGPIPE shall be sent if the socket has been shut down
+		 * for writing or is no longer connected. The latter case applies only to SOCK_STREAM
+		 * sockets. Currently, shutdown() closes the socket altogether, so unix_sendto()
+		 * cannot return EPIPE for a socket shut down for writing. unix_sendto() must also
+		 * not return EPIPE for no longer connected SOCK_DGRAM sockets, because SIGPIPE is not
+		 * required for them.
+		 */
+		posix_sigpipe();
 	}
 
 	(void)posix_fileDeref(f);
@@ -2202,6 +2221,10 @@ ssize_t posix_sendto(int socket, const void *message, size_t length, int flags, 
 		(void)posix_fileDeref(f);
 	}
 
+	if ((err == -EPIPE) && (((unsigned int)flags & MSG_NOSIGNAL) == 0U)) {
+		posix_sigpipe();
+	}
+
 	return err;
 }
 
@@ -2256,6 +2279,10 @@ ssize_t posix_sendmsg(int socket, const struct msghdr *msg, int flags)
 		}
 
 		(void)posix_fileDeref(f);
+	}
+
+	if ((err == -EPIPE) && (((unsigned int)flags & MSG_NOSIGNAL) == 0U)) {
+		posix_sigpipe();
 	}
 
 	return err;
--- a/posix/unix.c
+++ b/posix/unix.c
@@ -984,7 +984,6 @@ static ssize_t send(unsigned int socket, const void *buf, size_t len, unsigned i
 			}
 
 			if ((s->state & US_PEER_CLOSED) != 0U) {
-				(void)posix_tkill(process_getPid(proc_current()->process), 0, SIGPIPE);
 				err = -EPIPE;
 				break;
 			}


lukileczo:
I guess this has to wait for https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/310/files

julianuziemblo:
@lukileczo could you rebase this PR?

julianuziemblo (dismissed):
LGTM

adamdebek:
Exit() test need to be augmented, since one of test_case which close pipe doesn't handle `SIGPIPE`, which result in process termination.

julianuziemblo:
> Exit() test need to be augmented, since one of test_case which close pipe doesn't handle SIGPIPE, which result in process termination.

@adamdebek is https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/310 still in progress then, should it be done separately?

adamdebek:
There are two separate things in [PR #310](https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/310). I’ll create a separate PR for `SIGPIPE` in a moment.

posix/inet.c
```diff
@@ -204,14 +204,20 @@ ssize_t inet_sendto(unsigned int socket, const void *message, size_t length, uns
...
 
 	hal_memset(&msg, 0, sizeof(msg));
 	msg.type = sockmSend;
 	smi->send.flags = flags;
 	msg.i.data = message;
 	msg.i.size = length;
 
-	return sockdestcall(socket, &msg, dest_addr, dest_len);
+	err = sockdestcall(socket, &msg, dest_addr, dest_len);
+	if ((err == -EPIPE) && ((flags & MSG_NOSIGNAL) == 0U)) {
```
ziemleszcz: lwIP doesn't return `EPIPE` (see `err_to_errno_table`).

posix/posix.c
```diff
@@ -818,9 +826,13 @@ ssize_t posix_write(int fildes, void *buf, size_t nbyte, off_t offset)
 
 	if (f->type == ftUnixSocket) {
 		rcnt = unix_sendto((unsigned int)f->oid.id, buf, nbyte, 0, NULL, 0);
+		/* SIGPIPE handled by `send` */
 	}
 	else {
 		rcnt = proc_write(f->oid, offs, buf, nbyte, status);
+		if (rcnt == -EPIPE) {
+			posix_sigpipe();
```
ziemleszcz: `SIGPIPE` is raised for all servers. POSIX mandates this for pipes, FIFOs and connection-oriented sockets. For example  we return `EPIPE` in `libtty_write()`.

posix/posix.c
```diff
@@ -783,6 +783,14 @@ ssize_t posix_read(int fildes, void *buf, size_t nbyte, off_t offset)
 }
 
 
+void posix_sigpipe(void)
+{
+	thread_t *curr = proc_current();
+
+	(void)posix_tkill(process_getPid(curr->process), proc_getTid(curr), SIGPIPE);
```
ziemleszcz: `(void)threads_sigpost(curr->process, curr, SIGPIPE)`?

ziemleszcz:
This commit should be marked as breaking. Apps that don't handle `SIGPIPE` are now terminated.

posix/inet.c
lukileczo: I'm aware of that, until yesterday it wasn't signaled. However I decided to add it, makes this path POSIX compliant and LWIP not returning EPIPE is an implementation detail.

posix/posix.c
```diff
@@ -818,9 +826,13 @@ ssize_t posix_write(int fildes, void *buf, size_t nbyte, off_t offset)
 
 	if (f->type == ftUnixSocket) {
 		rcnt = unix_sendto((unsigned int)f->oid.id, buf, nbyte, 0, NULL, 0);
+		/* SIGPIPE handled by `send` */
```
ziemleszcz: Why comment on it only here? This applies to `posix_sendto()` and `posix_sendmsg()` as well. I'd remove this comment altogether.

posix/inet.c
ziemleszcz: Please add a FIXME/TODO comment then.

posix/unix.c
```diff
@@ -984,7 +984,9 @@ static ssize_t send(unsigned int socket, const void *buf, size_t len, unsigned i
 			}
 
 			if ((s->state & US_PEER_CLOSED) != 0U) {
-				(void)posix_tkill(process_getPid(proc_current()->process), 0, SIGPIPE);
+				if ((flags & MSG_NOSIGNAL) == 0U) {
```
ziemleszcz: Consider moving this code, together with the corresponding code in `inet_sendto()`, to `posix_write()`, `posix_send()`, and `posix_sendto()`. This way, all `EPIPE` cases would be handled in one file, and `posix_sigpipe()` could be made static.

lukileczo: I can do that, only thing is that according to POSIX in case of a no longer connected socket, `SIGPIPE` should be only sent if the socket type is `SOCK_STREAM`.
> [EPIPE]
A write was attempted on a socket that is shut down for writing, or is no longer connected. In the latter case, if the socket is of type SOCK_STREAM, a SIGPIPE signal shall also be sent to the thread.

We now don't return `EPIPE` in the `DGRAM` case, but in theory POSIX allows it, I can leave a comment in `posix_write`.

posix/inet.c
lukileczo: Do we want a comment in kernel about LWIP? In theory I could port another IP stack, which does return EPIPE and it wouldn't be applicable.

posix/posix.c
lukileczo: Removed

posix/inet.c
julianuziemblo: Indeed I wouldn't add the comment in the kernel. Maybe you could add an issue in YT that LWIP doesn't return EPIPE and so we can't deliver SIGPIPE to the user on connection hangup, but commenting here seems outside of the scope.

julianuziemblo (dismissed):
LGTM

posix/posix.c
```diff
@@ -829,6 +837,14 @@ ssize_t posix_write(int fildes, void *buf, size_t nbyte, off_t offset)
 		}
 	}
 
+	if ((rcnt == -EPIPE) && ((f->type == ftUnixSocket) || (f->type == ftPipe) || (f->type == ftFifo) || (f->type == ftInetSocket))) {
+		/* NOTE: in case of a UNIX socket, if the socket is no longer connected,
+		 * SIGPIPE shall be sent only if the socket is of type SOCK_STREAM.
+		 * Currently that's true because of the `unix_sendto` return value not returned in other cases.
```
ziemleszcz: That's not exactly true. According to POSIX, `unix_sendto()` can also return `EPIPE` when the socket has been shut down for writing (by calling `shutdown(s, SHUT_WR)`). Currently, this is not possible because our `shutdown()` implementation simply closes the socket. Please add this information to the note, as we may add this functionality in the future.

Also, please make the note clearer. The last sentence is difficult to understand.

lukileczo: Updated the note

ziemleszcz (approved):
LGTM.

However, I would welcome a phoenix-rtos-tests PR that adds tests for catching SIGPIPE on Unix sockets, pipes, and Internet sockets (marking that it doesn't work yet).

This PR adds SIGPIPE signal delivery to pipes, so it may be worth notifying the other teams about this change, since the default action for SIGPIPE is to terminate the process.

PR #521, opened 2024-02-13
posix: fix send() blocking

adamdebek:
Sending message larger than curricular buffer size to DGRAM or SQEPACKET sockets caused infinite block

JIRA: RTOS-778

## Types of changes

## How Has This Been Tested?

- [x] New test added: (add PR link here).

## Checklist:

## Special treatment

### DIFF
--- a/include/errno.h
+++ b/include/errno.h
@@ -59,6 +59,7 @@
 #define EWOULDBLOCK  EAGAIN  /* Operation would block */
 
 #define ENOTSOCK        88
+#define EMSGSIZE        90
 #define EPROTOTYPE      91
 #define ENOPROTOOPT     92
 #define EPROTONOSUPPORT 93
--- a/posix/unix.c
+++ b/posix/unix.c
@@ -750,6 +750,12 @@ static ssize_t send(unsigned socket, const void *buf, size_t len, int flags, con
 					_cbuffer_write(&conn->buffer, &len, sizeof(len));
 					_cbuffer_write(&conn->buffer, buf, err = len);
 				}
+				else if (conn->buffsz < len + sizeof(len)) { /* SOCK_DGRAM or SOCK_SEQPACKET */
+					err = -EMSGSIZE;
+					proc_lockClear(&conn->lock);
+					break;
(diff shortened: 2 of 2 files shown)


posix/unix.c
```diff
@@ -746,9 +746,16 @@ static ssize_t send(unsigned socket, const void *buf, size_t len, int flags, con
...
-				else if (_cbuffer_free(&conn->buffer) >= len + sizeof(len)) { /* SOCK_DGRAM or SOCK_SEQPACKET */
-					_cbuffer_write(&conn->buffer, &len, sizeof(len));
-					_cbuffer_write(&conn->buffer, buf, err = len);
+				else if (s->type == SOCK_DGRAM || s->type == SOCK_SEQPACKET) {
+					if (_cbuffer_free(&conn->buffer) >= len + sizeof(len)) {
+						_cbuffer_write(&conn->buffer, &len, sizeof(len));
+						_cbuffer_write(&conn->buffer, buf, err = len);
+					}
+					else {
+						err = -EMSGSIZE;
```
ziemleszcz, line 755: I think your should return EMSGSIZE only if packet cannot fit to empty buffer. Currently this change breaks nonblocking send (it can return EMSGSIZE instead of EAGAIN). Something like this:
if (s->type == SOCK_STREAM) {
...
else if (_cbuffer_free(&conn->buffer) >= len + sizeof(len)) {
...
else if (conn->buffsz < len + sizeof(len)) {
  err = -EMSGSIZE;
}

include/errno.h
```diff
@@ -59,6 +59,7 @@
 #define EWOULDBLOCK  EAGAIN  /* Operation would block */
 
 #define ENOTSOCK        88
+#define EMSGSIZE        90
```
nalajcie, line 62: please note that this errno is defined in libphoenix, so additional PR to remove it from there is needed

nalajcie:
maybe indicate in commit message title that it's related to unix sockets send() only?

ziemleszcz (approved):
LGTM

PR #526, opened 2024-03-08
arm: add optimized memcpy and memset implementations

lukileczo:
JIRA: RTOS-789

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


agkaminski (approved):
Sharing code between armv7m, armv7a and armv8 is planned, but not in the scope of this PR. Will be done later

hal/armv7a/_memcpy.S
```diff
@@ -0,0 +1,299 @@
...
+ * Author: Lukasz Leczkowski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#define DST r0
+#define SRC r1
+#define LEN r12
```
kemonats, line 18: If you use instead of the `r12 register some one in the `r0-r7` range you can reduce the size of the code.
Thus, for example, using `r5`
```
#define LEN r5
...
push {r0, r5}
...
pop {r0, r5}
bx lr
```
Result:
```
$ size _memcpy.o

LEN=r12 560
LEN=r5 488
```

kemonats:
A bit off this PR but what is further the point of using `-mno-unaligned-access` option for `ARM` ( there will be less code for structures with `packed attribute` ).

agkaminski:
@kemonats yes, removal of `-mno-unaligned-access` is planned, but it's impact on our project needs to be examined thoroughly first

agkaminski (approved):

lukileczo:
Oops, I've just realized that I've changed the register usage only on armv7-a, will do a follow up PR (as well as in libphoenix).

PR #528, opened 2024-03-15
msg API changes

lukileczo:
## Description

**Breaking changes!**

#### This PR introduces the following changes in msg API:
- `oid` is moved from `i.<type>` to common message part
- `const` is added to input data pointer - **this introduces warnings in some functions!**
- `err` is moved from `o.<type>` to `o` struct
- `mtGetAttrAll` msg type is added, allowing the caller to get all file attributes through one message. In this case `struct _attrs` should be allocated and pointer to it should be written into `msg.o.data` (and sizeof to `msg.o.size`). https://github.com/phoenix-rtos/libphoenix/pull/351 provides `_phoenix_initAttrsStruct` that initializes errors in `struct _attrs`.
Convention for the return value is following:
    - If the server doesn't at all provide support for `mtGetAttrAll` message, it will return negative value in `msg.o.err` field. Other fields remain uninitialized.
    - Otherwise, `msg.o.err` will be set to `0` and all per-attribute errors will be set accordingly. 

Additionally, `id` and `pid` fields from are removed from `ioctl_in_t` struct and `ioctl_out_t` struct is removed completely (but this structs were only used indirectly through `libphoenix`).
## Motivation and Context

Simplify message API.

Speed up `ls` command:
- on `armv7a9-zynq7000-qemu`
    - before changes - `psh_ls` call on directory with 100 files takes 1.8-2.3 s
    - after changes - 0.5-0.9 s
- on `ia32-generic-qemu`
    - before changes - 0.9-1.2 s
    - after changes - 0.5-0.7 s
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/include/msg.h
+++ b/include/msg.h
@@ -23,13 +23,14 @@ typedef int msg_rid_t;
  * Message types
  */
 
+/* clang-format off */
 
 enum {
 	/* File operations */
 	mtOpen = 0, mtClose, mtRead, mtWrite, mtTruncate, mtDevCtl,
 
 	/* Object operations */
-	mtCreate, mtDestroy, mtSetAttr, mtGetAttr,
+	mtCreate, mtDestroy, mtSetAttr, mtGetAttr, mtGetAttrAll,
 
 	/* Directory operations */
 	mtLookup, mtLink, mtUnlink, mtReaddir,
@@ -37,26 +38,52 @@ enum {
 	mtCount
 };
 
+/* clang-format on */
+
 
 #pragma pack(push, 8)
+
+
+struct _attr {
+	long long val;
+	int err;
+};
+
+
+struct _attrAll {
+	struct _attr mode;
+	struct _attr uid;
+	struct _attr gid;
+	struct _attr size;
+	struct _attr blocks;
+	struct _attr ioblock;
+	struct _attr type;
+	struct _attr port;
+	struct _attr pollStatus;
+	struct _attr eventMask;
+	struct _attr cTime;
+	struct _attr mTime;
+	struct _attr aTime;
+	struct _attr links;
+	struct _attr dev;
+};
 
 
 typedef struct _msg_t {
 	int type;
 	unsigned int pid;
 	unsigned int priority;
+	oid_t oid;
 
 	struct {
 		union {
 			/* OPEN/CLOSE */
 			struct {
-				oid_t oid;
 				int flags;
 			} openclose;
 
 			/* READ/WRITE/TRUNCATE */
 			struct {
-				oid_t oid;
 				off_t offs;
 				size_t len;
 				unsigned mode;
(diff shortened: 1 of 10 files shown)


include/msg.h
```diff
@@ -23,117 +23,124 @@ typedef int msg_rid_t;
...
 #pragma pack(push, 8)
 
 
+struct _attr {
+	long long val;
+	int err;
+};
+
+
+struct _attrs {
```
agkaminski, line 53: Just one char difference between names, easy to make a mistake or misread the code. Perhaps `_attrArray`, `_attrAll` or `_attrTab`?

posix/posix.c
```diff
@@ -1576,22 +1553,21 @@ int posix_fcntl(int fd, unsigned int cmd, char *ustack)
...
 
-void ioctl_pack(msg_t *msg, unsigned long request, void *data, id_t id)
+static void ioctl_pack(msg_t *msg, unsigned long request, void *data, oid_t oid)
 {
 	size_t size = IOCPARM_LEN(request);
 	ioctl_in_t *ioctl = (ioctl_in_t *)msg->i.raw;
 	struct ifconf *ifc;
 	struct rtentry *rt;
 
+	msg->oid = oid;
```
agkaminski, line 1563: Potential memcpy being emitted
```suggestion
	hal_memcpy(&msg->oid, &oid, sizeof(oid));
```

posix/posix.c
```diff
@@ -1576,22 +1553,21 @@ int posix_fcntl(int fd, unsigned int cmd, char *ustack)
 #define SIOCDELRT   _IOC(IOC_IN, 'S', 0x45, sizeof(struct rtentry))
 
 
-void ioctl_pack(msg_t *msg, unsigned long request, void *data, id_t id)
+static void ioctl_pack(msg_t *msg, unsigned long request, void *data, oid_t oid)
```
agkaminski, line 1556: Might be better to pass via pointer, some performance penalty in passing struct by value

proc/name.c
```diff
@@ -219,7 +219,7 @@ int proc_portLookup(const char *name, oid_t *file, oid_t *dev)
 
 	/* Query servers */
 	do {
-		msg->i.lookup.dir = srv;
+		msg->oid = srv;
```
agkaminski, line 222:
```suggestion
		hal_memcpy(&msg->oid, &srv, sizeof(srv));
```

proc/name.c
```diff
@@ -325,14 +327,15 @@ int proc_create(int port, int type, int mode, oid_t dev, oid_t dir, char *name,
 	msg->i.create.type = type;
 	msg->i.create.mode = mode;
 	msg->i.create.dev = dev;
-	msg->i.create.dir = dir;
+	msg->oid = dir;
```
agkaminski, line 330: Safer to use memcpy. Seem both previous version and line above had this problem before too

agkaminski (approved):

PR #537, opened 2024-04-12
Cortex-M: Fix pendsv induced crashes

agkaminski:
## Description

Hopefully fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1017 and https://github.com/phoenix-rtos/phoenix-rtos-project/issues/885 too

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


hal/armv7m/cpu.c
```diff
@@ -162,6 +159,7 @@ int hal_cpuPushSignal(void *kstack, void (*handler)(void), cpu_context_t *signal
 
 	signalCtx->psp -= sizeof(cpu_context_t);
 	signalCtx->hwctx.pc = (u32)handler;
+	signalCtx->hwctx.psr = 0x61000000;
```
lukileczo, line 162: Why do we need this value here? I think a comment also would be nice

lukileczo (commented):
One remark only, otherwise LGTM

lukileczo:
And I'm not sure if it's worth including the commits changing the pendsv handler which gets removed altogether in the next commit

agkaminski:
> And I'm not sure if it's worth including the commits changing the pendsv handler which gets removed altogether in the next commit

I wanted to keep bug fix in the history, just in case

lukileczo (approved):

PR #548, opened 2024-05-10
zynq7000: SMP kernel

jmaksymowicz:
Allows tasks to be scheduled on both CPU cores on zynq7000 platform.

## Description
NOTE: the corresponding PR in PLO must be merged first - otherwise on zynq7000 kernel will hang waiting for other CPU cores which will never start.

Code is capable of detecting number of available CPUs at runtime, the `NUM_CPUS` constant is used as maximum number of CPUs to prepare for.

One change was made not related to SMP: size of page directories was changed from `0x4000` bytes to `0x2000` bytes, as the upper half was unnecessary

Memory attributes were changed in common armv7a code, which affects imx6ull (but should not cause significant change).

Using SMP on armv7a9-zynq7000-qemu causes drastic slowdown (about 4x-5x slower) when executing single-core workloads. This is because under QEMU the idle thread creates a lot of load on the host CPU, as well as contention on the timer spinlock. This slowdown does not occur on real hardware. For this reason I suggest limiting QEMU to emulating only one core, as we can test SMP configurations on real hardware.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a9-zynq7000-zturn, armv7a9-zynq7000-qemu

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work: https://github.com/phoenix-rtos/plo/pull/329 https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/504


hal/armv7a/_interrupts.S
```diff
@@ -19,14 +19,32 @@
...
+	vpush {d0-d15}
+	vmrs \reg_tmp, fpscr
+	push {\reg_tmp}
+.endm
+
+.macro unlock_scheduler
+	ldr r0, =schedulerLocked
+	mov r1, #0
+	dmb
+	str r1, [r0]
```
agkaminski, line 35: `strex` perhaps?

jmaksymowicz: Exclusive monitors check memory locations for any write operation - so a normal `str` is enough, we don't need to do a `ldrex`, `strex` sequence.

hal/armv7a/arch/cpu.h
```diff
@@ -183,16 +184,46 @@ static inline int hal_cpuSupervisorMode(cpu_context_t *ctx)
...
+{
+	/* clang-format off */
+	__asm__ volatile (
+	"1:\n"
+		"ldrex r2, [%0]\n"
+		"add r2, r2, #1\n"
+		"strex r1, r2, [%0]\n"
+		"cmp r1, #0\n"
+		"bne 1b\n"
+		"dmb\n"
```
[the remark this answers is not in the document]
jmaksymowicz: From what I read, memory barriers in atomic operations are added before and after the operation. I added a barrier before the operation.

agkaminski (dismissed):

jmaksymowicz:
Reverted change that activated L2 cache - it caused stability problems in some tests. I haven't found the reason for it yet. Because it's not necessary for SMP I think we should do it in a different PR.

Darchiv (approved):

Darchiv (approved):

PR #549, opened 2024-05-14
proc_fork: copy parent context to child's kstack

lukileczo:
JIRA: RTOS-835

## Description

If a signal was to be handled on return from fork syscall and was delivered to the child process, the signal context would be constructed from junk.
This change fixes this issue so that the child can safely handle the signal.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `riscv64-generic-qemu`, `ia32-generic-qemu`

## Checklist:

## Special treatment

### DIFF
--- a/proc/process.c
+++ b/proc/process.c
@@ -1513,7 +1513,7 @@ int proc_fork(void)
 {
 	int err = -ENOSYS;
 #ifndef NOMMU
-	thread_t *current;
+	thread_t *current, *parent;
 	unsigned sigmask;
 	arg_t args[3];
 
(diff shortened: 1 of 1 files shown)


proc/process.c
```diff
@@ -1532,6 +1532,10 @@ int proc_fork(void)
 		current->kstack = current->execkstack;
 		_hal_cpuSetKernelStack(current->kstack + current->kstacksz);
 
+		parent = ((process_spawn_t *)current->execdata)->parent;
```
agkaminski, line 1535: Could you please add a comment why we do this? imho it's not obvious

PR #576, opened 2024-08-09
Cortex-M: NVIC and SCB code duplication cleanup

agkaminski:
Clenup standard part of Cortex-M component (i.e. SCB and NVIC) mass copy-paste across targets. If this approach passes, then it will be done over plo too

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (mcxn94x).

## Checklist:

## Special treatment


agkaminski:
@anglov just to verify if it does not break your stuff

anglov:
I really like this change as it remove a lot of duplicated code. I check it on DEND (checked low power) and NIL projects and at the first glance it looks ok!

Some general remarks (or ideas for future):
- I don't like using "SCB" and "System Control Block" name 
According to Armv7-M Architecture Reference Manual (DDI0403E.e) and Armv8-M Architecture Reference Manual (DDI0553B.y) SCB is small part of System Control Space which also contain NVIC, Systick, MPU and other control, debug  and id registers.
It's confusing for the first time. 
Maybe more appropriate name should be "SCS", as for "System Control Space"?
Maybe NVIC registers should also be used from one big SCS structure?
- I see that interrupts.c on armv7m architectures differs mainly with includes and priority of interrupts. It probably should be revised and also use common implementation in the future.
- It's worth considering if _init.S couldn't be shared between cores somehow. 
armv8m implementations differs mainly with interrupt size, hal entry point and cosmetic things and approaches if I didn't miss anything.
In armv7m implementation it differs  mainly with interrupt size FPU hard/soft implementation and .cpu  directive. In cortex-m4 was also errata requiring  additional dsb at intertupt dispatch, which was propagated to cortex-m7 too anyways.

Current refactoring is great step in good direction anyways.

Darchiv (approved):
Huge improvement, nice! Just two remarks (for the future, PR can be merged as-is):
- Check cache availability in `_hal_scbCacheIsSupported()` via Cache Control Identification Registers instead of relying solely on `cpuid->PARTNO`, because cache is a silicon option.
- A few functions are missing, e.g. `_imxrt_nvicGetPendingIRQ()`. They were not used, but maybe they should not be scrapped if code already exists.

PR #578, opened 2024-08-20
vm/mman: handle properly munmap of chunks larger than SIZE_PAGE

badochov:
JIRA: RTOS-893

## Description
Munmap handled badly segments bigger that SIZE_PAGE, after it the rb tree was in corrupted state.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia-32

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/581

### DIFF
--- a/vm/map.c
+++ b/vm/map.c
@@ -370,12 +370,47 @@ void *vm_mapFind(vm_map_t *map, void *vaddr, size_t size, u8 flags, u8 prot)
 }
 
 
+static void vm_mapEntryCopy(map_entry_t *dst, map_entry_t *src, int refAnons)
+{
+	hal_memcpy(dst, src, sizeof(map_entry_t));
+	src->amap = amap_ref(dst->amap);
+	/* In case of splitting the entry the anons shouldn't be reffed as they just change the owner. */
+	if (refAnons != 0) {
+		amap_getanons(dst->amap, dst->aoffs, dst->size);
+	}
+	src->object = vm_objectRef(dst->object);
+}
+
+
+static void vm_mapEntrySplit(process_t *p, vm_map_t *m, map_entry_t *e, map_entry_t *new, size_t len)
+{
+	vm_mapEntryCopy(new, e, 0);
+
+	new->vaddr += len;
+	new->size -= len;
+	new->aoffs += len;
+	new->offs = (new->offs == -1) ? -1 : (new->offs + len);
+	new->lmaxgap = 0;
+
+	e->size = len;
+	e->rmaxgap = 0;
+	map_augment(&e->linkage);
+
+	_map_add(p, m, new);
+}
+
+
 int _vm_munmap(vm_map_t *map, void *vaddr, size_t size)
 {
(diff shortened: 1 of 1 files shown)


badochov:
> Perhaps we could use `vm_mapEntrySplit` in case of the "in the middle" `munmap`

Only thing that comes to my mind is changing this part to

```c
			s = map_alloc();
			/* This case if only possible if an unmapped region is in the middle of a single entry,
			 * so there is no possibility of partially unmapping. */
			if (s == NULL) {
				return -ENOMEM;
			}
			vm_mapEntrySplit(proc, map, e, s, overlapStart);
			continue; /* Process in next iteration. */
```

What do you think about it?

agkaminski:
> > Perhaps we could use `vm_mapEntrySplit` in case of the "in the middle" `munmap`
> 
> Only thing that comes to my mind is changing this part to
> 
> ```c
> 			s = map_alloc();
> 			/* This case if only possible if an unmapped region is in the middle of a single entry,
> 			 * so there is no possibility of partially unmapping. */
> 			if (s == NULL) {
> 				return -ENOMEM;
> 			}
> 			vm_mapEntrySplit(proc, map, e, s, overlapStart);
> 			continue; /* Process in next iteration. */
> ```
> 
> What do you think about it?

Looks nice!

agkaminski (approved):
LGTM

badochov:
@agkaminski @damianloew
`unix-socket` test failed on CI on armv7a-qemu due to timeout 358ms instead of 350ms, should I be concerned?

https://github.com/phoenix-rtos/phoenix-rtos-kernel/actions/runs/11050286361/job/30697841405?pr=578

agkaminski:
> @agkaminski @damianloew `unix-socket` test failed on CI on armv7a-qemu due to timeout 358ms instead of 350ms, should I be concerned? Munmap changes should change this test, as in poll and gettime theres no munmap
> 
> https://github.com/phoenix-rtos/phoenix-rtos-kernel/actions/runs/11050286361/job/30697841405?pr=578

I guess it did pass on the rerun. @damianloew perhaps this timeout is somewhat on the edge?

damianloew:
> > @agkaminski @damianloew `unix-socket` test failed on CI on armv7a-qemu due to timeout 358ms instead of 350ms, should I be concerned? Munmap changes should change this test, as in poll and gettime theres no munmap
> > https://github.com/phoenix-rtos/phoenix-rtos-kernel/actions/runs/11050286361/job/30697841405?pr=578
> 
> I guess it did pass on the rerun. @damianloew perhaps this timeout is somewhat on the edge?

Yeah, we had a similar problem some time ago and changed the value to 350 from 310:
https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/350

We will consider increasing the value more - probably we will check 400. However I am not sure whether higher values are caused by our system or by the gh machine on which the tests are executed

PR #586, opened 2024-09-09
build: remove program headers from first loadable segment

badochov:
Program headers should not be loaded into kernel as there is no use for them. Previously starting kernel at start of memory region required explicit linker script.

JIRA: RTOS-908

## Description
On all platform but ARM, program headers are placed in loadable segment before _init section. Resulting from this _init section start address isn't start address of first loadable segment. Due to that in PLO we may not be able to find proper segment for kernel if an address of the section that the kernel is to be loaded to = VADDR_KERNEL_INIT. Because of this additional linker script was required for sparc in kernel.

On ARM we headers weren't loaded most probably because of a bug(or very fortunated accident) in our binutils patch.
THe [change](https://github.com/bminor/binutils-gdb/commit/a2bea3245ecf4ff67f923c85adf57aa78423eb45) introduced new file armelf_phoenix.sh but did't use it (ld/configure.tgt), instead it used armelf by default which defines EMBEDDED to true resulting in headers not being included.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


Makefile
```diff
@@ -53,14 +53,22 @@ include log/Makefile
...
+
+# By default ld adds program headers to first LOAD segment.
+# However, in kernel headers are not used and it is expected that kernel address space starts at _start.
+# This target creates linker script that do not add program headers into the first loadable segment.
+# 1) Get internal linker script with prefix and suffix containing noise.
+# 2) Match only internal linker script which is between two lines of =
+# 3) Remove those lines
+# 4) Remove "+ SIZEOF_HEADERS", which causes inclusion of program headers.
+$(PREFIX_O)$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt:
+	$(SIL)$(LD) $(LDFLAGS_PREFIX)--verbose 2>/dev/null | sed -n '/^=*$$/,/^=*$$/p' | sed '1,1d; $$d' | sed s/"\s*+\s*SIZEOF_HEADERS"// > "$@"
```
lukileczo, line 65: `^=*$` will also match empty line as `*` means "zero or more characters". I'd change it to `^==*$` as there's no "one or more" in basic `sed` regexp and we also want to be compatible with MacOS.

Also you can grab the necessary lines in single command: `sed -n '/^==*$$/,/^==*$$/{/^=*$$/d;p;}'` - deletion can strip empty lines I guess.

badochov: Thanks for suggestions! I've added `==*` however I feel that breaking the sed into two executions makes the script much more readable, thus I'd prefer to keep it as it is

lukileczo (approved):
LGTM

nalajcie:
please don't use semantic commit messages, this is not in our coding standard (they don't help creating changelog in case of superproject)

badochov:
> please don't use semantic commit messages, this is not in our coding standard (they don't help creating changelog in case of superproject)

FIxed

nalajcie:
> > please don't use semantic commit messages, this is not in our coding standard (they don't help creating changelog in case of superproject)
> 
> FIxed

Thanks. 
Just for reference: we might think about semantic commit messages if/when we will switch to monorepo, right now we do generate changelog from the PR descriptions (Types of changes section) - which is not ideal (as one PR might fix/add new feature simultaneously)) - for example: https://github.com/phoenix-rtos/phoenix-rtos-project/releases/tag/v3.2.0

PR #600, opened 2024-09-26
hal/imxrt: Support RTT as a HAL console

Darchiv:
UART is enabled by default (current behavior), but can be turned off by defining UART_CONSOLE_KERNEL with an empty value. RTT is enabled by defining RTT_ENABLED with a truthy value and (optionally) selecting a channel to use by RTT_CONSOLE_KERNEL (0-based). UART and RTT may be used at the same time.

RTT driver reuses the control block configured by plo, which must also create a map named via RTT_SYSPAGE_MAP_NAME (default: "rtt"). Channel configuration is not altered at all, so RTT will work even when user-mode driver takes control - exception dumps will still be printed.

JIRA: RTOS-754

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt1170-evk.

## Checklist:

## Special treatment

### DIFF
--- a/hal/arm/rtt.c
+++ b/hal/arm/rtt.c
@@ -1,0 +1,158 @@
+/*
+ * Phoenix-RTOS
+ *
+ * SEGGER's Real Time Transfer - simplified driver
+ *
+ * Copyright 2023-2024 Phoenix Systems
+ * Author: Gerard Swiderski, Daniel Sawka
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "include/errno.h"
+#include "syspage.h"
+#include "hal/arm/barriers.h"
+
+#include <board_config.h>
+#include "rtt.h"
+
+#ifndef RTT_SYSPAGE_MAP_NAME
+#define RTT_SYSPAGE_MAP_NAME "rtt"
+#endif
+
+#ifndef RTT_CB_SIZE
+#define RTT_CB_SIZE 256
+#endif
+
+
+struct rtt_pipe {
+	const char *name;
+	volatile unsigned char *ptr;
+	unsigned int sz;
+	volatile unsigned int wr;
+	volatile unsigned int rd;
+	unsigned int flags;
+};
+
+
+struct rtt_desc {
+	char tag[16];
+	unsigned int txChannels;
+	unsigned int rxChannels;
+	struct rtt_pipe channels[];
+};
+
+
+static struct {
+	volatile struct rtt_desc *rtt;
+} common;
+
+
+static int rtt_check(unsigned int chan, rtt_dir_t dir)
+{
+	if ((dir == rtt_dir_up) && (chan >= common.rtt->txChannels)) {
+		return -ENODEV;
+	}
+
+	if ((dir == rtt_dir_down) && (chan >= common.rtt->rxChannels)) {
+		return -ENODEV;
+	}
+
+	return 0;
+}
+
+
+int _hal_rttWrite(unsigned int chan, const void *buf, unsigned int count)
+{
+	unsigned int sz;
+	unsigned int rd;
+	unsigned int wr;
+	unsigned int todo;
+	volatile unsigned char *dstBuf;
+
+	if (rtt_check(chan, rtt_dir_up) < 0) {
+		return -ENODEV;
+	}
+
+	hal_cpuDataMemoryBarrier();
+	dstBuf = common.rtt->channels[chan].ptr;
+	sz = common.rtt->channels[chan].sz - 1;
+	rd = (common.rtt->channels[chan].rd + sz) & sz;
+	wr = common.rtt->channels[chan].wr & sz;
+	todo = count;
+
+	/* TODO: Support all buffer modes (currently only trim is used, regardless of flags) */
+	while ((todo != 0) && (rd != wr)) {
+		dstBuf[wr] = *(const unsigned char *)buf++;
+		wr = (wr + 1) & sz;
+		todo--;
+	}
+
+	hal_cpuDataMemoryBarrier();
+	common.rtt->channels[chan].wr = wr;
+
+	return count - todo;
+}
+
+
+int _hal_rttTxAvail(unsigned int chan)
+{
+	unsigned int sz;
+	unsigned int rd;
+	unsigned int wr;
+
+	if (rtt_check(chan, rtt_dir_up) < 0) {
+		return -ENODEV;
+	}
+
+	hal_cpuDataMemoryBarrier();
+	sz = common.rtt->channels[chan].sz - 1;
+	rd = (common.rtt->channels[chan].rd + sz) & sz;
+	wr = common.rtt->channels[chan].wr & sz;
+
+	if (wr > rd) {
+		return sz + 1 - (wr - rd);
+	}
+	else {
+		return rd - wr;
+	}
+}
+
+
+int _hal_rttReset(unsigned int chan, rtt_dir_t dir)
+{
+	if (rtt_check(chan, dir) < 0) {
+		return -ENODEV;
+	}
+
+	hal_cpuDataMemoryBarrier();
+	if (dir == rtt_dir_up) {
+		common.rtt->channels[chan].wr = common.rtt->channels[chan].rd;
+	}
+	else {
+		chan = common.rtt->txChannels + chan;
+		common.rtt->channels[chan].rd = common.rtt->channels[chan].wr;
+	}
+	hal_cpuDataMemoryBarrier();
+	return 0;
+}
+
+
+int _hal_rttInit(void)
+{
+	const syspage_map_t *map = syspage_mapNameResolve(RTT_SYSPAGE_MAP_NAME);
+
+	if (map == NULL) {
+		return -ENOENT;
+	}
+
+	if (map->start + RTT_CB_SIZE > map->end) {
+		return -EINVAL;
+	}
+
+	/* TODO: Place CB always at the start of the map? */
+	common.rtt = (void *)(map->end - RTT_CB_SIZE);
+	return 0;
+}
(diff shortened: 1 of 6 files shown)


hal/arm/rtt.c
```diff
@@ -0,0 +1,193 @@
...
+	else {
+		return rd - wr;
+	}
+}
+
+
+void _hal_rttReset(int chan, rtt_dir_t dir)
+{
+	if (rtt_check(chan, dir) < 0) {
+		return -ENODEV;
```
jmaksymowicz, line 160: Return value in a function returning void

hal/arm/rtt.c
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	volatile struct rtt_desc *rtt = common.rtt;
+
+	hal_cpuDataMemoryBarrier();
+	if (dir == rtt_dir_up) {
+		common.rtt->channels[chan].wr = common.rtt->channels[chan].rd;
```
jmaksymowicz, line 167: You can use pointer `rtt` instead of `common.rtt`.

jmaksymowicz (commented):
Looks good, just a few suggestions.

hal/arm/rtt.c
Darchiv: Done

hal/arm/rtt.c
Darchiv: Done

hal/arm/rtt.c
```diff
@@ -0,0 +1,193 @@
...
+}
+
+
+int _hal_rttRead(int chan, void *buf, unsigned int count)
+{
+	if (rtt_check(chan, rtt_dir_down) < 0) {
+		return -ENODEV;
+	}
+
+	volatile struct rtt_desc *rtt = common.rtt;
```
agkaminski, line 81: Why this intermediate variable?

hal/arm/rtt.c
```diff
@@ -0,0 +1,193 @@
...
+
+	hal_cpuDataMemoryBarrier();
+	chan = rtt->txChannels + chan;
+	volatile unsigned char *srcBuf = rtt->channels[chan].ptr;
+	unsigned char *dstBuf = (unsigned char *)buf;
+	unsigned int sz = rtt->channels[chan].sz - 1;
+	unsigned int rd = rtt->channels[chan].rd & sz;
+	unsigned int wr = rtt->channels[chan].wr & sz;
+	unsigned int todo = count;
```
agkaminski, line 90: We keep variables declarations on the beginning in plo and kernel

hal/arm/rtt.c
```diff
@@ -0,0 +1,193 @@
...
+
+	volatile struct rtt_desc *rtt = common.rtt;
+
+	hal_cpuDataMemoryBarrier();
+	chan = rtt->txChannels + chan;
+	volatile unsigned char *srcBuf = rtt->channels[chan].ptr;
+	unsigned char *dstBuf = (unsigned char *)buf;
+	unsigned int sz = rtt->channels[chan].sz - 1;
+	unsigned int rd = rtt->channels[chan].rd & sz;
+	unsigned int wr = rtt->channels[chan].wr & sz;
```
agkaminski, line 89: Do we have a strong assumption that `sz` is a power of 2?

hal/arm/rtt.c
```diff
@@ -0,0 +1,193 @@
...
+		return -ENODEV;
+	}
+
+	volatile struct rtt_desc *rtt = common.rtt;
+
+	hal_cpuDataMemoryBarrier();
+	const unsigned char *srcBuf = (const unsigned char *)buf;
+	volatile unsigned char *dstBuf = rtt->channels[chan].ptr;
+	unsigned int sz = rtt->channels[chan].sz - 1;
+	unsigned int rd = (rtt->channels[chan].rd + sz) & sz;
```
agkaminski, line 117: This code will fail silently and do nothing when `count == N * sz`

hal/arm/rtt.h
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Copyright 2023-2024 Phoenix Systems
+ * Author: Gerard Swiderski, Daniel Sawka
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#ifndef HAL_ARMV_RTT_H_
```
agkaminski, line 14:
```suggestion
#ifndef HAL_ARM_RTT_H_
```

lib/helpers.h
```diff
@@ -0,0 +1,40 @@
...
+These macros produce a logically correct result only if X is defined. You may use `#ifdef` or `defined()`
+for this purpose. Unfortunately, `defined()` cannot be conveniently put inside a macro as this is undefined
+behavior (see -Wexpansion-to-defined for details), so you have to use it directly on the spot, for example:
+`#if defined(PERIPH1) && !ISEMPTY(PERIPH1)
+// Use PERIPH1
+#endif`
+*/
+
+/* True if X is empty (has no value). The result in #if is valid only if defined(X) is true */
+#define ISEMPTY(X) ((0 - X - 1) == 1 && (X + 0) != -2)
```
agkaminski, line 36: This macro seems redundant, we have stardard `defined()`

hal/armv7m/imxrt/117x/console.c
```diff
@@ -14,21 +14,35 @@
...
+#define RTT_CONSOLE_KERNEL
+#endif
 #endif
 
 #define CONCAT3(a, b, c) a##b##c
 #define CONSOLE_BAUD(n)  (CONCAT3(UART, n, _BAUDRATE))
 
-#if CONSOLE_BAUD(UART_CONSOLE)
-#define CONSOLE_BAUDRATE CONSOLE_BAUD(UART_CONSOLE)
+#if !ISEMPTY(UART_CONSOLE_KERNEL) && CONSOLE_BAUD(UART_CONSOLE_KERNEL)
```
agkaminski, line 44:
```suggestion
#if defined(UART_CONSOLE_KERNEL) && CONSOLE_BAUD(UART_CONSOLE_KERNEL)
```

lib/helpers.h
```diff
@@ -0,0 +1,40 @@
...
+/*
+These macros should only be used in `#if` and `#elif` directives, because undefined identifiers expand to 0 there.
+Otherwise there will be "use of undefined identifier" errors (an exception: identifier is first checked for
+existence with e.g. `#ifdef`).
+
+Anything that expands to a numerical expression (or to an empty value) is fine. String literals don't work.
+*/
+
+/* True if X expands to a non-zero value. Also works with undefined and empty identifiers. */
+#define ISTRUTHY(X) ((X + 0) != 0)
```
agkaminski, line 24: Dead code

lib/helpers.h
```diff
@@ -0,0 +1,40 @@
...
+behavior (see -Wexpansion-to-defined for details), so you have to use it directly on the spot, for example:
+`#if defined(PERIPH1) && !ISEMPTY(PERIPH1)
+// Use PERIPH1
+#endif`
+*/
+
+/* True if X is empty (has no value). The result in #if is valid only if defined(X) is true */
+#define ISEMPTY(X) ((0 - X - 1) == 1 && (X + 0) != -2)
+/* True if X is either 0 or 1. The result in #if is valid only if defined(X) is true */
+#define ISBOOLEAN(X) (!ISEMPTY(X) && ((X + 0) == 0 || (X + 0) == 1))
```
agkaminski, line 38: Dead code. Why do we need this check anyway?

hal/armv7m/imxrt/117x/console.c
```diff
@@ -156,3 +177,20 @@ void _hal_consoleInit(void)
...
+
+void _hal_consoleInit(void)
+{
+	unsigned int rtt_enable = 0u;
+
+#if RTT_ENABLED && !ISEMPTY(RTT_CONSOLE_KERNEL)
+	rtt_enable = 1u;
+#endif
+
+	_hal_rttInit(rtt_enable);
```
agkaminski, line 191: Perhaps it would be cleaner to just move this define based check to _hal_rttInit instead? Then we do not need `enable` argument

lib/helpers.h
jmaksymowicz: It's not, if a macro is defined as empty, `defined()` will be true. The intention is that in `board_config.h` `RTT_CONSOLE_KERNEL` will be defined as empty if we want to disable RTT, or undefined if we want to use the default settings.

hal/arm/rtt.c
agkaminski: Let's assume that we start with rd == wr and count == sz. Then (rd + sz % sz) == rd and rd still == wr. We will not enter while below (&& (rd != wr)) and in consequence, do nothing at all

jmaksymowicz: You didn't notice that the variable `sz` is the buffer size minus 1. Therefore, the result of this line is equivalent to `(rtt->channels[chan].rd + rtt->channels[chan].sz - 1) % (rtt->channels[chan].sz)`

agkaminski: Ok, then count == sz-1, problem persists ;)

hal/arm/rtt.c
```diff
@@ -0,0 +1,154 @@
...
+	hal_cpuDataMemoryBarrier();
+	dstBuf = common.rtt->channels[chan].ptr;
+	sz = common.rtt->channels[chan].sz - 1;
+	rd = (common.rtt->channels[chan].rd + sz) & sz;
+	wr = common.rtt->channels[chan].wr & sz;
+	todo = count;
+
+	/* TODO: Support all buffer modes (currently only trim is used, regardless of flags) */
+	while ((todo != 0) && (rd != wr)) {
+		dstBuf[wr] = *(unsigned char *)buf++;
```
agkaminski, line 88: casting `const` away
```suggestion
		dstBuf[wr] = *(const unsigned char *)buf++;
```

Darchiv: Done

agkaminski (approved):

agkaminski (approved):

PR #606, opened 2024-10-04
syscalls: Sanitize all pointers

agkaminski:
DONE: RTOS-939

## Motivation and Context

Fixes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/976

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/syscalls.c
+++ b/syscalls.c
@@ -36,12 +36,14 @@
 
 void syscalls_debug(void *ustack)
 {
+	process_t *proc = proc_current()->process;
 	const char *s;
 
-	/* FIXME: pass strlen(s) from userspace */
-
 	GETFROMSTACK(ustack, const char *, s, 0);
-	hal_consolePrint(ATTR_USER, s);
+
+	if (vm_mapStringBelongs(proc, s) == 0) {
+		hal_consolePrint(ATTR_USER, s);
+	}
 }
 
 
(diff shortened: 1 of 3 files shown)


syscalls.c
```diff
@@ -173,52 +175,119 @@ int syscalls_release(void *ustack)
...
+	}
+
+	if (vm_mapBelongs(proc, v, sizeof(v)) < 0) {
+		return -1;
+	}
+
+	for (i = 0;; ++i) {
+		if (vm_mapBelongs(proc, &v[i], sizeof(v)) < 0) {
+			return -1;
+		}
```
badochov, line 193: When`i` = 0 `&v[i] == v`

vm/map.c
```diff
@@ -1120,6 +1122,32 @@ int vm_mapBelongs(const struct _process_t *proc, const void *ptr, size_t size)
 }
 
 
+int vm_mapStringBelongs(const struct _process_t *proc, const char *str)
```
badochov, line 1125: This check has one more vulnerability than `vm_mapBelongs` has. User can also alter the string post NUL byte check. We either have to copy the string or pass the string with length found by this function.

syscalls.c
```diff
@@ -173,52 +175,119 @@ int syscalls_release(void *ustack)
 }
 
 
+static int syscalls_sanitizeVector(process_t *proc, char **v)
```
badochov, line 178: On top of vulnerabilities of `vm_mapStringBelongs` and `vm_mapStringBelongs`, this function is also vulnerable to user changing the elements of vector. Unfortunatelly, I think that there's no way around copying the vector, but a FIXME note should be sufficient for now.

badochov:
Probably beyond the scope of this PR, but it would be nice to be resiliant to
```asm
mov sp,0
int 0x80
```

vm/map.c
Darchiv, line 1125: Shouldn't there be a fast return for `str == NULL` or at least handle that per-syscall? To avoid arithmetic on `NULL`s and unnecessary locks.

Darchiv (commented):
Several syscalls (e.g. `sys_munmap`, `mprotect`) still lack any checks - is this intentional?

badochov:
> Several syscalls (e.g. `sys_munmap`, `mprotect`) still lack any checks - is this intentional?

PR #611, opened 2024-10-22
process: unify load on NOMMU

badochov:
JIRA: RTOS-958

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: sparc-mimas, armv8r

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7m/arch/elf.h
+++ b/hal/armv7m/arch/elf.h
@@ -1,0 +1,29 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * Hardware Abstraction Layer ELF support
+ *
+ * Copyright 2024 Phoenix Systems
+ * Author: Hubert Badocha
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#ifndef _HAL_ARMV7M_ELF_H_
+#define _HAL_ARMV7M_ELF_H_
+
+
+#define R_ARM_ABS32   2
+#define R_ARM_TARGET1 38
+
+
+static inline int hal_isRelReloc(int relType)
+{
+	return ((relType == R_ARM_ABS32) || (relType == R_ARM_TARGET1)) ? 1 : 0;
+}
+
+#endif
--- a/proc/process.c
+++ b/proc/process.c
@@ -15,6 +15,7 @@
  */
 
 #include "hal/hal.h"
+#include "hal/elf.h"
 #include "include/errno.h"
 #include "include/signal.h"
 #include "vm/vm.h"
@@ -742,15 +743,12 @@ int process_load(process_t *process, vm_object_t *o, off_t base, size_t size, vo
 	Elf32_Ehdr *ehdr;
 	Elf32_Phdr *phdr;
 	Elf32_Shdr *shdr, *shstrshdr;
-#ifdef __sparc__
-	Elf32_Rela *rela;
-	Elf32_Sym *sym;
-	ptr_t symTab;
-#else
-	Elf32_Rel *rel;
-#endif
+	Elf32_Rela rela;
+	int isrela;
+	Elf32_Addr symval;
+	Elf32_Sym *symTab = NULL;
 	unsigned prot, flags, reloffs;
-	int i, j, relocsz = 0, reltype, badreloc = 0, err;
+	int i, j, relocsz = 0, badreloc = 0, err;
 	void *relptr;
 	char *snameTab;
 	ptr_t *got;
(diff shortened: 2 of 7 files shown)


proc/process.c
```diff
@@ -877,93 +874,80 @@ int process_load(process_t *process, vm_object_t *o, off_t base, size_t size, vo
...
+					/* Don't modify ELF file! */
+					if (((ptr_t)relptr >= (ptr_t)base) && ((ptr_t)relptr < ((ptr_t)base + size))) {
+						++badreloc;
+						continue;
+					}
+
+					sym = (Elf32_Sym *)(symTab + (ELF32_R_SYM(rela->r_info) * sizeof(Elf32_Sym)));
+
+					/* Write addend to the address */
+					*(char **)relptr = (char *)(sym->st_value + rela->r_addend);
```
agkaminski, line 917: This is a pretty much the only difference between SPARC and ARM stuff. Perhaps we could store strcmp result and use it here to do this operation conditionally?

badochov: I thought about it and it was a bunny hole.

Each architecture has a separate set of Relocation Types, however relocation types are just numbers and there maybe a clash eg. `R_SPARC_32` on SPARC  and `R_ARM_REL32`(unused now) on ARM have the same value of 3. Thus, we should compare relocation type separately on each architecture. This can be done either via ifdefs or via checking if in the ELF header the machine type(`e_machine` field). Now this clash will not happen, as SPARC uses RELA and ARM uses REL. If we unify the code there is high potential for a bug in the future if more relocation types would be needed for some reason.

In my opinion the code should be modified so that:
1) Getting `rela` section on an architecture using `rel` would result in `ENOEXEC` and vice versa
2) We should have separate relocations handler for each architecture

What do you think about it? Also, Should we check the architecture via ifdef or the ELF HEADER?

agkaminski: Well, first question that has to be asked is if we'll ever need to handle any more relocation types in the kernel? Aren't dynamic linker solely in the userspace?

Regarding collisions of values of relocation types between archs - this can be solved by e.g. defining our own, abstract, machine independent relocation types on per arch basis, and handle these instead of e.g. `R_SPARC_32` directly (that might be represented by e.g. `HAL_R_REL32` or something along these lines).

badochov: 1) To support static FDPIC program some 2 more relocations are needed, unless we want to move relocation process to the userspace. 

Additionally if we ever want to handle C++ on NOMMU then to have proper exceptions handling there may be a need to add few more relocations. However that's a harder topic as currently exception handling causes relocations in text segment :skull_and_crossbones: 

2) Nice idea!

Also my bad, both ARM and SPARC support in ABI both rela and rel, however GCC by default uses only one of those types if possible.

agkaminski: My perspective is as follows - better for this code to do the thing it has to do nice and simple and to not prepare for the future that might never come (kinda a part of the top-down philosophy).

From my side - the proposed version is way better and status quo as is, so I have no objection to merge if you wish to not make any further changes.

agkaminski (dismissed):

proc/process.c
badochov: I checked and compilation with `-static-pie` fixes issue of relocation in text segment in exceptions handling section

badochov: I have adapted code to you suggestions, and added feature to handle both Rel and Rela relocations on both targets

hal/armv7m/arch/elf.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+
+#define R_ARM_ABS32   2
+#define R_ARM_TARGET1 38
+
+
+static inline int hal_isRelReloc(int relType)
+{
+	return (relType == R_ARM_ABS32) || (relType == R_ARM_TARGET1);
```
agkaminski, line 23: MISRA

Side note, I'm not sure if this approach is scalable. Might be actually more useful to be able to just know relocation type in the process_load.
```suggestion
	return ((relType == R_ARM_ABS32) || (relType == R_ARM_TARGET1)) ? 1 : 0;
```

hal/armv7m/arch/elf.h
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright 2024 Phoenix Systems
+ * Author: Hubert Badocha
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+
+#define R_ARM_ABS32   2
```
agkaminski, line 17: Missing header guard (applies to other new headers too)

proc/process.c
```diff
@@ -877,93 +875,76 @@ int process_load(process_t *process, vm_object_t *o, off_t base, size_t size, vo
...
-			reltype = ELF32_R_TYPE(rel->r_info);
+		for (j = 0; j < (shdr->sh_size / shdr->sh_entsize); ++j) {
+			/* Valid for both Elf32_Rela and Elf32_Rel, due to correct size being stored in shdr->sh_entsize. */
+			/* For .rel. section make sure not to access addend field! */
+			hal_memcpy(&rela, (Elf32_Rela *)((ptr_t)shdr->sh_offset + (ptr_t)ehdr + (j * shdr->sh_entsize)), shdr->sh_entsize);
+			reltype = ELF32_R_TYPE(rela.r_info);
 
-			if (reltype == R_ARM_ABS32 || reltype == R_ARM_TARGET1) {
-				relptr = (void *)rel->r_offset;
+			if (hal_isRelReloc(reltype) == 0) {
```
agkaminski, line 909: `reltype` became redundant in the current approach
```suggestion
			if (hal_isRelReloc(ELF32_R_TYPE(rela.r_info)) == 0) {
```

proc/process.c
```diff
@@ -877,93 +875,76 @@ int process_load(process_t *process, vm_object_t *o, off_t base, size_t size, vo
...
 					return -ENOEXEC;
 				}
+				symval = 0;
+			}
+			else {
+				symval = symTab[ELF32_R_SYM(rela.r_info)].st_value;
+			}
+
+			if (isrela == 0) {
+				/* If Rel is used addend stored is in the location. */
```
agkaminski, line 936: Yoda :smile: 
```suggestion
				/* If Rel is used addend is stored in the location. */
```

proc/process.c
```diff
@@ -877,93 +875,76 @@ int process_load(process_t *process, vm_object_t *o, off_t base, size_t size, vo
...
+				*(char **)relptr += symval;
+			}
+			else {
+				*(char **)relptr = (char *)(symval + rela.r_addend);
```
lukileczo, line 940: I don't know if we expect any more relocations in kernel, but it's worth noting that different relocation types might need different calculations, so maybe it would be better to create some per-arch function like `hal_doReloc` which would do appropriate calculation for each relocation type (and then [this comment](https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/611/files#r1816169695) would be no issue)

hal/armv7m/arch/elf.h
badochov: When it comes to the scalability its a hard call without having more relocations. To support FDPIC format I will have to extend the list in this function on ARM, and add one other type checking function. That sound ok to me

proc/process.c
badochov: I prefer the approach in which we group relocations, independent of the architecture, based on how they are calculated.
Looking at the list of relocations on both architecture this should be doable.

agkaminski (approved):

PR #623, opened 2024-11-13
syscalls: make nsleep handle unslept time correctly

adamdebek:
JIRA: RTOS-977

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/syscalls.c
+++ b/syscalls.c
@@ -330,7 +330,7 @@ int syscalls_nsleep(void *ustack)
 	process_t *proc = proc_current()->process;
 	time_t *sec;
 	long int *nsec;
-	time_t start, us, stop, diff;
+	time_t start, us, stop, elapsed, unslept;
 	int ret;
 
 	GETFROMSTACK(ustack, time_t *, sec, 0);
@@ -360,10 +360,11 @@ int syscalls_nsleep(void *ustack)
 
 	if (ret == -EINTR) {
 		proc_gettime(&stop, NULL);
-		diff = stop - start;
-		if (diff < us) {
-			*sec = diff / (1000 * 1000);
-			*nsec = (diff % (1000 * 1000)) * 1000;
+		elapsed = stop - start;
+		if (us > elapsed) {
+			unslept = us - elapsed;
+			*sec = unslept / (1000 * 1000);
+			*nsec = (unslept % (1000 * 1000)) * 1000;
 		}
 	}
 


syscalls.c
```diff
@@ -360,10 +360,15 @@ int syscalls_nsleep(void *ustack)
 
 	if (ret == -EINTR) {
 		proc_gettime(&stop, NULL);
-		diff = stop - start;
-		if (diff < us) {
-			*sec = diff / (1000 * 1000);
-			*nsec = (diff % (1000 * 1000)) * 1000;
+		elapsed = stop - start;
+		unslept = us - elapsed;
+		if ((long long)unslept <= 0LL) {
```
agkaminski, line 365: This check is UB - cast might cause integer overflow, compiler has right to optimize this condition to false (as no integer overflow may occur per standard)

adamdebek: I changed it and also uncertained why time_t is unsigned. If it was signed all would have been much easier.

syscalls.c
```diff
@@ -360,10 +360,15 @@ int syscalls_nsleep(void *ustack)
...
 		proc_gettime(&stop, NULL);
-		diff = stop - start;
-		if (diff < us) {
-			*sec = diff / (1000 * 1000);
-			*nsec = (diff % (1000 * 1000)) * 1000;
+		elapsed = stop - start;
+		unslept = us - elapsed;
+		if (elapsed >= us) {
+			*sec = 0;
+			*nsec = 0;
```
kemonats, line 367: repeated
`*sec = 0` and `*nsec = 0` are set before 
`if (ret == -EINTR)`

syscalls.c
```diff
@@ -360,10 +360,17 @@ int syscalls_nsleep(void *ustack)
...
-			*nsec = (diff % (1000 * 1000)) * 1000;
+		elapsed = stop - start;
+		if (us > elapsed) {
+			unslept = us - elapsed;
+		}
+		else {
+			unslept = 0;
+		}
+
+		if (unslept > 0) {
```
agkaminski, line 371: `us == elapsed` is very unlikely, nevertheless, this will behave well under this edge case
```suggestion
		elapsed = stop - start;
		if (us > elapsed) {
			unslept = us - elapsed;
```

adamdebek: This code also handles cases where elapsed is greater than us, which, since time_t is unsigned, would cause unslept to become a very large number instead of 0.

agkaminski: Doesn't matter, in this case we don't use value of unslept

adamdebek: I see it now

agkaminski (approved):

PR #624, opened 2024-11-18
ia32/exceptions: fix FPU exceptions

badochov:
## Description
This PR aims to resolve situation with exceptions caused by FPU on Intel.

First commit resolves issue where recently instead of a triple fault on FPU error exception 16 got presented on the screen, due to wrong instruction setting flag.

Second, commit fixes saving context in exceptions.

Third commit changes `fsave` in context saving procedure to `fnsave` to prevent infinite recursion on FPU exception causing a triple fault. Due to this change exception 16 is reported correctly as exception 16 not as tripple fault. Thus, exception 16 is excepted to appear not Triple fault.

Fourth, commit fixes the root cause of the intermittent error. The `cr0Bits` were overwritten by eax during context restoration. We may restore twice from the same context in case of a signal pushed during `_threads_schedule`. Then we restore once from the context to go to the signal handler, after the signal is handled the context is modified to have original `eip` and `esp` and we longjump to it resulting in second restoration from the context with overwritten `cr0Bits`. This resulted in `TS` flag being removed without proper FPU initialization on next restore.  

The following commits are optimizations.

Fixes: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1012

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic-qemu

## Checklist:

## Special treatment

### DIFF
--- a/hal/ia32/_exceptions.S
+++ b/hal/ia32/_exceptions.S
@@ -156,8 +156,7 @@ sym:
 	call *%eax;\
 	addl $8, %esp;\
 	/* cpu_context_t */;\
-	movl %esp, %eax;\
-	addl $28, %eax;\
+	leal 28(%esp), %eax;\
 	pushl %eax;\
 	call hal_cpuSupervisorMode;\
 	addl $4, %esp;\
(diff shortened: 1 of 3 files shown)


agkaminski (approved):

mateusz-bloch (approved):
Tested the changes with the latest master of the project, and everything seems to be ok.

PR #627, opened 2024-11-28
fork: fix return value not being set in child

badochov:
Fixes a return value not being set in child during fork on ARM(7a,8r) and SPARC.

The problem was caused by threads_setupUserValue using current thread stack to obtain context, while in fork we return using parents stack.

JIRA: RTOS-979

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32, zynq-qemu

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/_interrupts.S
+++ b/hal/armv7a/_interrupts.S
@@ -259,31 +259,29 @@ _interrupts_dispatch:
 .type _syscalls_dispatch, %function
 _syscalls_dispatch:
 	stmfd sp, {r0-r4}^
-	sub r1, sp, #0x14
+	sub r2, sp, #0x14
 	mrs r3, spsr
-	mov r2, lr
+	mov r1, lr
 	tst r3, #THUMB_STATE
-	ldreq r0, [r2, #-4]
+	ldreq r0, [r1, #-4]
 	biceq r0, r0, #0xff000000
-	ldrneh r0, [r2, #-2]
+	ldrneh r0, [r1, #-2]
 	bicne r0, r0, #0xff00
 	mrc p15, 0, r4, c13, c0, 4 /* TPIDRPRW */
 
 	cpsie af, #SYS_MODE
 
 	/* Store CPU state onto kernel stack */
-	stmfd r4!, {r2}
+	stmfd r4!, {r1}
 	stmfd r4!, {r5-r14}
-	mov r2, sp
+	mov r1, sp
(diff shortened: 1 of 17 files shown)


syscalls.c
```diff
@@ -1868,8 +1868,6 @@ void *syscalls_dispatch(int n, char *ustack)
 		proc_threadEnd();
 	}
 
-	threads_setupUserReturn(retval);
```
lukileczo, line 1871: I think I'd prefer changing `syscalls_dispatch` to `void *syscalls_dispatch(int n, cpu_context_t *ctx, char *ustack)` and keeping this call here, that way it would not have to be added in each asm file.

badochov: Yeah, youre rigth I switched to this approach

hal/armv7a/_interrupts.S
```diff
@@ -259,31 +259,31 @@ _interrupts_dispatch:
...
 	push_fpu_state r4
-	sub r1, sp, #8
-	push {r1}
-	push {r1}
+	sub r2, sp, #8
+	push {r2}
+	push {r2}
 
-	mov r1, r2
+	mov r2, sp
```
lukileczo, line 286: This instruction shouldn't be necessary, this value is written into `r2` 2 instructions earlier

hal/armv8r/_interrupts.S
```diff
@@ -259,31 +259,31 @@ _interrupts_dispatch:
...
 	push_fpu_state r4
-	sub r1, sp, #8
-	push {r1}
-	push {r1}
+	sub r2, sp, #8
+	push {r2}
+	push {r2}
 
-	mov r1, r2
+	mov r2, sp
```
lukileczo, line 286: Same as in armv7a

hal/sparcv8leon/_traps.S
```diff
@@ -673,6 +673,7 @@ s_wovfl_done:
 	FPU_SAVE
 
 s_fpu_done:
+	mov %sp, %o2 /* cpu_context_t * */
```
lukileczo, line 676: Can be done in delay slot after `call syscalls_dispatch`: change `nop` to `add %sp, 0x60, %o2`
And please update `syscalls_dispatch` signature in the comment in code

lukileczo (approved):
LGTM

agkaminski (approved):

PR #629, opened 2024-12-16
MCXN94x: Log information whether CPU0 or CPU1 is used

agkaminski:
JIRA: RTOS-988

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (mcx).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-build/pull/219 and https://github.com/phoenix-rtos/plo/pull/373).

### DIFF
--- a/hal/armv8m/cpu.c
+++ b/hal/armv8m/cpu.c
@@ -192,8 +192,13 @@ char *hal_cpuInfo(char *info)
 	}
 
 	if (((cpuinfo >> 4) & 0xfffu) == 0xd21u) {
+#ifdef MCX_USE_CPU1
+		hal_strcpy(info + i, "Micro Cortex-M33 ");
+		i += 17;
+#else
 		hal_strcpy(info + i, "Cortex-M33 ");
 		i += 11;
+#endif
 	}
 
 	*(info + i++) = 'r';


Darchiv:
I wonder if we should have a runtime option to get current core's number. Without it, drivers which need to distinguish cores would have to use defines (i.e. 
`MCX_USE_CPU1`). I suspect there is no easy way to get "self" core info from MCU and this flag must be used (in a centralized manner?)

agkaminski:
> I wonder if we should have a runtime option to get current core's number. Without it, drivers which need to distinguish cores would have to use defines (i.e. `MCX_USE_CPU1`). I suspect there is no easy way to get "self" core info from MCU and this flag must be used (in a centralized manner?)

Perhaps, but I cannot imagine a use case for this information - the only difference is in fact missing CPU features (e.g. FPU). There is no other difference from the programmer perspective (in particular memory map is the same for both cores, excluding ARM region that is private for each core).

Darchiv:
> > I wonder if we should have a runtime option to get current core's number. Without it, drivers which need to distinguish cores would have to use defines (i.e. `MCX_USE_CPU1`). I suspect there is no easy way to get "self" core info from MCU and this flag must be used (in a centralized manner?)
> 
> Perhaps, but I cannot imagine a use case for this information - the only difference is in fact missing CPU features (e.g. FPU). There is no other difference from the programmer perspective (in particular memory map is the same for both cores, excluding ARM region that is private for each core).

I'm thinking about Mailbox use case - we need to know which core is the "remote" one, because we have to write to a specific register: IRQ0 for CPU0 or IRQ1 for CPU1.

Darchiv (approved):

PR #633, opened 2025-01-29
posix/dup2: fix case when fds are the same

badochov:
JIRA: RTOS-1002

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32
## Checklist:

## Special treatment

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -788,7 +788,6 @@ int posix_dup(int fildes)
 }
 
 
-/* FIXME: handle fildes == fildes2 */
 int _posix_dup2(process_info_t *p, int fildes, int fildes2)
 {
 	open_file_t *f, *f2;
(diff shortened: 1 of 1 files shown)


posix/posix.c
```diff
@@ -806,6 +806,10 @@ int _posix_dup2(process_info_t *p, int fildes, int fildes2)
 		return -EBADF;
 	}
 
+	if (fildes == fildes2) {
+		return EOK;
```
Darchiv, line 810: > If fildes2 is already a valid open file descriptor, it shall be closed first, unless fildes is equal to fildes2 in which case dup2() shall return fildes2 without closing it.

According to POSIX, `fildes2` should be returned in this case. `dup2()` returns fd or an error and sets errno, so `EOK` would be interpreted as fd=0

posix/posix.c
```diff
@@ -806,6 +806,10 @@ int _posix_dup2(process_info_t *p, int fildes, int fildes2)
 		return -EBADF;
 	}
 
+	if (fildes == fildes2) {
```
Darchiv, line 809: FIXME at the top of this function can be removed now

posix/posix.c
badochov: Yeah, I ve been testing on fd == 0 so i didnt catch it thanks!

Darchiv (approved):

PR #645, opened 2025-04-02
pmap: allow removing multiple pages in one call

lukileczo:
## Description

This change targets performance optimization on multicore targets. Previously, `pmap_remove` had to be called for each page to be removed, which resulted in overhead in TLB shootdowns - one for each page. With this change, a single TLB shootdown can be issued to remove multiple pages, which limits the number of IPIs sent to other cores.

The drawback is that `pmap_common` spinlock is held longer - for the duration of multiple page unmap.

The behaviour on `aarch64` remains unchanged - the flush is done in every loop.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `sparcv8leon-gr740-mini`, `riscv64-generic-qemu`

## Checklist:

## Special treatment

### DIFF
--- a/hal/aarch64/pmap.c
+++ b/hal/aarch64/pmap.c
@@ -214,7 +214,7 @@ static void _pmap_asidDealloc(pmap_t *pmap)
 }
 
 
-static void _pmap_cacheOpBeforeChange(descr_t oldEntry, descr_t newEntry, void *vaddr, unsigned lvl)
+static void _pmap_cacheOpBeforeChange(descr_t oldEntry, descr_t newEntry, ptr_t vaddr, unsigned lvl)
 {
 	addr_t pa;
 	int oldCachedRW, newNoncached;
@@ -231,10 +231,10 @@ static void _pmap_cacheOpBeforeChange(descr_t oldEntry, descr_t newEntry, void *
 	oldCachedRW = ((oldEntry & DESCR_AP2) == 0) && (ATTR_FROM_DESCR(oldEntry) == MAIR_IDX_CACHED);
 	newNoncached = ((newEntry & DESCR_VALID) == 0) || (ATTR_FROM_DESCR(newEntry) != MAIR_IDX_CACHED);
 	if (oldCachedRW && newNoncached) {
-		pa = _pmap_hwTranslate((ptr_t)vaddr);
+		pa = _pmap_hwTranslate(vaddr);
 		if (((pa & 1) == 0) && (DESCR_PA(oldEntry) == (pa & ((1uL << 48) - (1uL << 12))))) {
 			/* VA is currently mapped - simply flush cache by virtual address */
-			hal_cpuFlushDataCache((ptr_t)vaddr, (ptr_t)vaddr + SIZE_PAGE);
+			hal_cpuFlushDataCache(vaddr, vaddr + SIZE_PAGE);
 		}
 		else {
 			/* Temporarily map to pmap_common.scratch_page */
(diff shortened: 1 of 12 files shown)


lukileczo:
@jmaksymowicz is there an option on `aarch64` to flush a range of addresses? Currently, the flush will still be done every loop.

hal/armv7a/pmap.c
```diff
@@ -353,32 +353,35 @@ int pmap_enter(pmap_t *pmap, addr_t pa, void *va, int attr, page_t *alloc)
...
+		pti = ID_PTABLE(vaddr);
+
+		addr = pmap->pdir[pdi];
+		if (addr == PDIR_TYPE_INVALID) {
+			continue;
+		}
 
-	_pmap_writeEntry(pmap_common.sptab, vaddr, 0, 0, asid);
+		/* Map page table corresponding to vaddr */
+		_pmap_mapScratch(addr, asid);
```
agkaminski, line 377: There's a good chance that we will map the same page after first iteration - would be nice to skip this when possible

jmaksymowicz:
> @jmaksymowicz is there an option on `aarch64` to flush a range of addresses? Currently, the flush will still be done every loop.

On AArch64 TLB invalidation instructions take only one address as argument, so it's not possible. However, invalidation can apply to TLBs in all cores in SMP so there is no need to send IPIs for this purpose - so there is little performance to be gained.

hal/aarch64/pmap.c
```diff
@@ -509,43 +509,46 @@ int pmap_enter(pmap_t *pmap, addr_t paddr, void *vaddr, int attr, page_t *alloc)
...
-		}
-		else if ((lvl == 3) || ((entry & DESCR_TABLE) == 0)) {
-			break;
-		}
-		else {
-			addr = DESCR_PA(entry);
-			_pmap_mapScratch(pmap_common.scratch_tt, addr);
-			tt = pmap_common.scratch_tt;
+	for (vaddr = (ptr_t)vstart; vaddr < (ptr_t)vend; vaddr += SIZE_PAGE) {
+		tt = pmap->ttl1;
```
jmaksymowicz, line 521: Because virtual addresses will increment linearly you can avoid the lookups and mapping the translation table again:
```c
if (first_time || (TTL_IDX(3, vaddr) == 0)) {
 /* code that maps translation table into scratch */
}
```
We only need to map a new 3rd level translation table if this is the first iteration of the loop or the previous iteration reached the end of the translation table (so this iteration is at the start).

hal/armv7a/pmap.c
```diff
@@ -66,6 +66,7 @@ extern unsigned int _etext;
 #define PDIR_TYPE_L2TABLE 0x00001
 #define PDIR_TYPE_INVALID 0x00000
 
+#define PDIR_ATTRS (PGHD_PRESENT | PGHD_READ | PGHD_WRITE)
```
agkaminski, line 69: Nitpick - not really used for pdir, only for ptables, how about
```suggestion
#define SCRATCH_ATTRS (PGHD_PRESENT | PGHD_READ | PGHD_WRITE)
```

agkaminski (commented):
One nitpick, otherwise LGTM. Would like @Darchiv to approve before merging

Darchiv (approved):
LGTM.

I'm only concerned with lengthening the spinlock hold time and thus potentially increasing the maximum response time of a real time system, but that has yet to be analyzed system-wide. For future reference only

PR #649, opened 2025-05-19
Coredump: implementation on MMU targets

etiaro:
## Description

Implement configurable dumping of serialized memory, registers and metadata on a fault that allows for post-mortem debugging of crashed process using GDB after [parsing on host machine](https://github.com/phoenix-rtos/phoenix-rtos-hostutils/pull/58).

## Motivation and Context

[RTOS-1043](https://jira.phoenix-rtos.com/browse/RTOS-1043) Research core dump and backtrace techniques
[RTOS-1049](https://jira.phoenix-rtos.com/browse/RTOS-1049) Implement core dump on MMU targets

<details>
  <summary>Example coredump log</summary>
  
  This is the most minimalistic coredump example (only stack memory and registers, one thread), generated on ia32.
  Allows for backtrace and shows all stack variables in GDB.
  
  ```javascript
  _____________COREDUMP_START_____________
  /usr/bin/coredump_test: 14 #PF
f0VMRgEBAf4JAAQAAwAB/gcANP4LADQAIAAC/gcABAAAAHT+CwAkAQAAJAH+CgABAAAAmAEAAIxf/gYAdAAAAHQAAAAG/gcABQAAAJAAAAABAAAAQ09SRf4QAA7+CwAn/kcAeFY0EiMAAAAjAAAAIwAAADP+BwBchAQIGwAAAEYyAACMXwAAI/4HAAUAAABsAAAAAgAAAENPUkX+BAB/AwAAIAEAAP//AAAAhAQIGwAAAIxfAAAj/jwAgABA/gUAZmbIBED+BQBmZsgEQDL+CwB/D38D/ggAMv4HAHhWNBL+CABmZshCmv4Fmck//hAA+CAAALiIBAj+IAD4IAAAY9k7lA==
  ______________COREDUMP_END______________
  ```
  
</details>

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-qemu, riscv64-qemu, armv7a9-zynq7000-qemu

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/exceptions.c
+++ b/hal/armv7a/exceptions.c
@@ -19,6 +19,7 @@
 #include "hal/console.h"
 #include "hal/string.h"
 #include "include/mman.h"
+#include "proc/elf.h"
 
 
 #define EXC_ASYNC_EXTERNAL      0x16
@@ -49,18 +50,23 @@ struct {
 enum { exc_reset = 0, exc_undef, exc_svc, exc_prefetch, exc_abort };
 
 
-void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
+const char *hal_exceptionMnemonic(int n)
 {
 	static const char *const mnemonics[] = {
-		"0 #Reset",       "1 #Undef",    "2 #Syscall",    "3 #Prefetch",
-		"4 #Abort",       "5 #Reserved", "6 #FIRQ",       "7 #IRQ"
+		"0 #Reset", "1 #Undef", "2 #Syscall", "3 #Prefetch",
+		"4 #Abort", "5 #Reserved", "6 #FIRQ", "7 #IRQ"
 	};
+
+	return mnemonics[n & 0x7];
+}
+
+
+void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
+{
 	size_t i = 0;
 
-	n &= 0x7;
-
 	hal_strcpy(buff, "\nException: ");
-	hal_strcpy(buff += hal_strlen(buff), mnemonics[n]);
+	hal_strcpy(buff += hal_strlen(buff), hal_exceptionMnemonic(n));
 	hal_strcpy(buff += hal_strlen(buff), "\n");
 	buff += hal_strlen(buff);
 
@@ -231,4 +237,84 @@ void _hal_exceptionsInit(void)
 	exceptions.undefHandler = exceptions_defaultHandler;
 	exceptions.abortHandler = exceptions_defaultHandler;
 	exceptions.defaultHandler = exceptions_defaultHandler;
+}
+
+
+cpu_context_t *hal_excToCpuCtx(exc_context_t *ctx)
+{
+	return &ctx->cpuCtx;
+}
+
+
+void hal_coredumpGRegset(void *buff, cpu_context_t *ctx)
+{
+	u32 *regs = (u32 *)buff;
+	*(regs++) = ctx->r0;
+	*(regs++) = ctx->r1;
+	*(regs++) = ctx->r2;
+	*(regs++) = ctx->r3;
+	*(regs++) = ctx->r4;
+	*(regs++) = ctx->r5;
+	*(regs++) = ctx->r6;
+	*(regs++) = ctx->r7;
+	*(regs++) = ctx->r8;
+	*(regs++) = ctx->r9;
+	*(regs++) = ctx->r10;
+	*(regs++) = ctx->fp;
+	*(regs++) = ctx->ip;
+	*(regs++) = ctx->sp;
+	*(regs++) = ctx->lr;
+	*(regs++) = ctx->pc;
+	*(regs++) = ctx->psr;
+}
+
+
+void hal_coredumpThreadAux(void *buff, cpu_context_t *ctx)
+{
+	static const char ARMVFP_NAME[] = "LINUX";
+	Elf32_Nhdr nhdr;
+	nhdr.n_namesz = sizeof(ARMVFP_NAME);
+	nhdr.n_descsz = sizeof(ctx->freg) + sizeof(ctx->fpsr);
+	nhdr.n_type = NT_ARM_VFP;
+	hal_memcpy(buff, &nhdr, sizeof(nhdr));
+	buff = (char *)buff + sizeof(nhdr);
+	hal_memcpy(buff, ARMVFP_NAME, sizeof(ARMVFP_NAME));
+	buff = (char *)buff + ((sizeof(ARMVFP_NAME) + 3) & ~3);
+	hal_memcpy(buff, ctx->freg, sizeof(ctx->freg));
+	buff = (char *)buff + sizeof(ctx->freg);
+	hal_memcpy(buff, &ctx->fpsr, sizeof(ctx->fpsr));
+}
+
+
+#define COMPAT_HWCAP_VFP   (1 << 6)
+#define COMPAT_HWCAP_NEON  (1 << 12)
+#define COMPAT_HWCAP_VFPv3 (1 << 13)
+#define HWCAP_VFPv3        (COMPAT_HWCAP_VFP | COMPAT_HWCAP_NEON | COMPAT_HWCAP_VFPv3)
+
+#define AT_HWCAP 16
+#define AT_NULL  0
+
+
+void hal_coredumpGeneralAux(void *buff)
+{
+	static const char AUXV_NAME[] = "CORE";
+	Elf32_Nhdr nhdr;
+	struct {
+		u32 a_type;
+		u32 a_val;
+	} auxv[2];
+
+	nhdr.n_namesz = sizeof(AUXV_NAME);
+	nhdr.n_descsz = sizeof(auxv);
+	nhdr.n_type = NT_AUXV;
+	hal_memcpy(buff, &nhdr, sizeof(nhdr));
+	buff = (char *)buff + sizeof(nhdr);
+	hal_memcpy(buff, AUXV_NAME, sizeof(AUXV_NAME));
+	buff = (char *)buff + ((sizeof(AUXV_NAME) + 3) & ~3);
+
+	auxv[0].a_type = AT_HWCAP;
+	auxv[0].a_val = HWCAP_VFPv3;
+	auxv[1].a_type = AT_NULL;
+	auxv[1].a_val = 0;
+	hal_memcpy(buff, auxv, sizeof(auxv));
 }
--- a/hal/ia32/exceptions.c
+++ b/hal/ia32/exceptions.c
@@ -23,6 +23,7 @@
 
 #include "include/mman.h"
 #include "include/errno.h"
+#include "proc/elf.h"
 
 
 /* Exception stubs */
@@ -112,18 +113,23 @@ ptr_t hal_exceptionsPC(exc_context_t *ctx)
 }
 
 
-void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
+const char *hal_exceptionMnemonic(int n)
 {
 	static const char *const mnemonics[] = {
-		"0 #DE",  "1 #DB",  "2 #NMI", "3 #BP",      "4 #OF",  "5 #BR",  "6 #UD",  "7 #NM",
-		"8 #BF",  "9 #",    "10 #TS", "11 #NP",     "12 #SS", "13 #GP", "14 #PF", "15 #",
-		"16 #MF", "17 #AC", "18 #MC", "19 #XM/#XF", "20 #VE", "21 #",   "22 #",   "23 #",
-		"24 #",   "25 #",   "26 #",   "27 #",       "28 #",   "29 #",   "30 #SE", "31 #" };
+		"0 #DE", "1 #DB", "2 #NMI", "3 #BP", "4 #OF", "5 #BR", "6 #UD", "7 #NM",
+		"8 #BF", "9 #", "10 #TS", "11 #NP", "12 #SS", "13 #GP", "14 #PF", "15 #",
+		"16 #MF", "17 #AC", "18 #MC", "19 #XM/#XF", "20 #VE", "21 #", "22 #", "23 #",
+		"24 #", "25 #", "26 #", "27 #", "28 #", "29 #", "30 #SE", "31 #"
+	};
 
+	return mnemonics[n & 0x1f];
+}
+
+
+void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
+{
 	size_t i = 0;
 	u32 ss;
-
-	n &= 0x1f;
 
 	/* clang-format off */
 	__asm__ volatile(
@@ -135,7 +141,7 @@ void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
 	/* clang-format on */
 
 	hal_strcpy(buff, "\nException: ");
-	hal_strcpy(buff += hal_strlen(buff), mnemonics[n]);
+	hal_strcpy(buff += hal_strlen(buff), hal_exceptionMnemonic(n));
 	hal_strcpy(buff += hal_strlen(buff), "\n");
 	buff += hal_strlen(buff);
 
@@ -300,4 +306,53 @@ __attribute__ ((section (".init"))) void _hal_exceptionsInit(void)
 	exceptions.handlers[7] = exceptions_exc7_handler;
 
 	return;
+}
+
+
+cpu_context_t *hal_excToCpuCtx(exc_context_t *ctx)
+{
+	return &ctx->cpuCtx;
+}
+
+
+void hal_coredumpGRegset(void *buff, cpu_context_t *ctx)
+{
+	u32 *regs = (u32 *)buff;
+	*(regs++) = ctx->ebx;
+	*(regs++) = ctx->ecx;
+	*(regs++) = ctx->edx;
+	*(regs++) = ctx->esi;
+	*(regs++) = ctx->edi;
+	*(regs++) = ctx->ebp;
+	*(regs++) = ctx->eax;
+	*(regs++) = ctx->ds;
+	*(regs++) = ctx->es;
+	*(regs++) = ctx->fs;
+	*(regs++) = ctx->gs;
+	*(regs++) = 0;
+	*(regs++) = ctx->eip;
+	*(regs++) = ctx->cs;
+	*(regs++) = ctx->eflags;
+	*(regs++) = ctx->esp;
+	*(regs++) = ctx->ss;
+}
+
+
+void hal_coredumpThreadAux(void *buff, cpu_context_t *ctx)
+{
+	static const char FPREGSET_NAME[] = "CORE";
+	Elf32_Nhdr nhdr;
+	nhdr.n_namesz = sizeof(FPREGSET_NAME);
+	nhdr.n_descsz = sizeof(ctx->fpuContext);
+	nhdr.n_type = NT_FPREGSET;
+	hal_memcpy(buff, &nhdr, sizeof(nhdr));
+	buff = (char *)buff + sizeof(nhdr);
+	hal_memcpy(buff, FPREGSET_NAME, sizeof(FPREGSET_NAME));
+	buff = (char *)buff + ((sizeof(FPREGSET_NAME) + 3) & ~3);
+	hal_memcpy(buff, &ctx->fpuContext, sizeof(ctx->fpuContext));
+}
+
+
+void hal_coredumpGeneralAux(void *buff)
+{
 }
--- a/hal/riscv64/exceptions.c
+++ b/hal/riscv64/exceptions.c
@@ -20,6 +20,7 @@
 #include "hal/string.h"
 
 #include "include/mman.h"
+#include "proc/elf.h"
 
 #define SIZE_EXCEPTIONS 16
 
@@ -31,10 +32,8 @@ static struct {
 } exceptions_common;
 
 
-void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
+const char *hal_exceptionMnemonic(int n)
 {
-	unsigned int i = 0;
-
 	static const char *mnemonics[] = {
 		"0 Instruction address missaligned", "1 Instruction access fault", "2 Illegal instruction", "3 Breakpoint",
 		"4 Reserved", "5 Load access fault", "6 AMO address misaligned", "7 Store/AMO access fault",
@@ -42,10 +41,16 @@ void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
 		"12 Instruction page fault", "13 Load page fault", "14 Reserved", "15 Store/AMO page fault"
 	};
 
-	n &= 0xf;
+	return mnemonics[n & 0xf];
+}
+
+
+void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
+{
+	unsigned int i = 0;
 
 	hal_strcpy(buff, "\nException: ");
-	hal_strcpy(buff += hal_strlen(buff), mnemonics[n]);
+	hal_strcpy(buff += hal_strlen(buff), hal_exceptionMnemonic(n));
 	hal_strcpy(buff += hal_strlen(buff), "\n");
 	buff += hal_strlen(buff);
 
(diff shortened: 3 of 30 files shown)


hal/armv7a/exceptions.c
```diff
@@ -49,18 +50,24 @@ struct {
 enum { exc_reset = 0, exc_undef, exc_svc, exc_prefetch, exc_abort };
 
 
-void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
+const char *hal_exceptionMnemonic(int n)
```
agkaminski, line 53: As this is an external HAL API, perhaps `n` should be sanitized

hal/armv7a/exceptions.c
```diff
@@ -232,3 +239,84 @@ void _hal_exceptionsInit(void)
...
+	*(regs++) = ctx->r8;
+	*(regs++) = ctx->r9;
+	*(regs++) = ctx->r10;
+	*(regs++) = ctx->fp;
+	*(regs++) = ctx->ip;
+	*(regs++) = ctx->sp;
+	*(regs++) = ctx->lr;
+	*(regs++) = ctx->pc;
+	*(regs++) = ctx->psr;
+}
```
agkaminski, line 270:
```suggestion
}


```

hal/armv7a/exceptions.c
```diff
@@ -232,3 +239,84 @@ void _hal_exceptionsInit(void)
...
+void hal_coredumpThreadAux(void *buff, cpu_context_t *ctx)
+{
+	const char ARMVFP_NAME[] = "LINUX";
+	Elf32_Nhdr nhdr;
+	nhdr.n_namesz = sizeof(ARMVFP_NAME);
+	nhdr.n_descsz = sizeof(ctx->freg) + sizeof(ctx->fpsr);
+	nhdr.n_type = NT_ARM_VFP;
+	hal_memcpy(buff, &nhdr, sizeof(nhdr));
+	buff += sizeof(nhdr);
```
agkaminski, line 279: `void *` arithmetic, applies in other places too
```suggestion
	buff = (char *)buff + sizeof(nhdr);
```

hal/armv7a/exceptions.c
```diff
@@ -232,3 +239,84 @@ void _hal_exceptionsInit(void)
...
+	Elf32_Nhdr nhdr;
+	nhdr.n_namesz = sizeof(AUXV_NAME);
+	nhdr.n_descsz = sizeof(elf_auxv_t) * 2;
+	nhdr.n_type = NT_AUXV;
+	hal_memcpy(buff, &nhdr, sizeof(nhdr));
+	buff += sizeof(nhdr);
+	hal_memcpy(buff, AUXV_NAME, sizeof(AUXV_NAME));
+	buff += (sizeof(AUXV_NAME) + 3) & ~3;
+
+	elf_auxv_t auxv[2];
```
agkaminski, line 315: Keep var declaration at the beginning of the file

hal/ia32/exceptions.c
```diff
@@ -301,3 +309,55 @@ __attribute__ ((section (".init"))) void _hal_exceptionsInit(void)
...
+{
+	u32 *regs = (u32 *)buff;
+	*(regs++) = ctx->ebx;
+	*(regs++) = ctx->ecx;
+	*(regs++) = ctx->edx;
+	*(regs++) = ctx->esi;
+	*(regs++) = ctx->edi;
+	*(regs++) = ctx->ebp;
+	*(regs++) = ctx->eax;
+	;
```
agkaminski, line 330:
```suggestion

```

hal/ia32/exceptions.c
```diff
@@ -301,3 +309,55 @@ __attribute__ ((section (".init"))) void _hal_exceptionsInit(void)
...
+	*(regs++) = ctx->ds;
+	*(regs++) = ctx->es;
+	*(regs++) = ctx->fs;
+	*(regs++) = ctx->gs;
+	;
```
agkaminski, line 335:
```suggestion

```

hal/riscv64/exceptions.c
```diff
@@ -246,3 +252,71 @@ __attribute__((section(".init"))) void _hal_exceptionsInit(void)
...
+cpu_context_t *hal_excToCpuCtx(exc_context_t *ctx)
+{
+	return ctx;
+}
+
+
+void hal_coredumpGRegset(void *buff, cpu_context_t *ctx)
+{
+	u64 *regs = (u64 *)buff;
+
```
agkaminski, line 266:
```suggestion
```

agkaminski (commented):
Partial review, will tackle the rest later

hal/armv7a/exceptions.c
```diff
@@ -49,18 +50,28 @@ struct {
...
+	}
+	return mnemonics[n];
+}
+
+
+void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
+{
 	size_t i = 0;
 
 	n &= 0x7;
```
kemonats, line 71: This line is unnecessary.
You check in lines 60-62.

hal/armv7a/exceptions.c
```diff
@@ -232,3 +243,85 @@ void _hal_exceptionsInit(void)
...
+	*(regs++) = ctx->pc;
+	*(regs++) = ctx->psr;
+}
+
+
+void hal_coredumpThreadAux(void *buff, cpu_context_t *ctx)
+{
+	const char ARMVFP_NAME[] = "LINUX";
+	Elf32_Nhdr nhdr;
+	nhdr.n_namesz = sizeof(ARMVFP_NAME);
```
kemonats, line 281: With `nul` char ?

hal/armv7a/exceptions.c
```diff
@@ -232,3 +243,85 @@ void _hal_exceptionsInit(void)
...
+
+
+void hal_coredumpGeneralAux(void *buff)
+{
+	const char AUXV_NAME[] = "CORE";
+	Elf32_Nhdr nhdr;
+	elf_auxv_t auxv[2];
+
+	nhdr.n_namesz = sizeof(AUXV_NAME);
+	nhdr.n_descsz = sizeof(elf_auxv_t) * 2;
```
kemonats, line 315: better
```suggestion
	nhdr.n_descsz = sizeof(auxv);
```

hal/armv7a/exceptions.c
etiaro: Yes, that's how it's done in linux kernel

hal/armv7a/exceptions.c
etiaro: I don't know if there can't be some other information encoded in higher bits, if not here then on other architectures. Will move this to line 60 and get rid of if statement.

proc/coredump.c
```diff
@@ -0,0 +1,518 @@
...
+#include "coredump.h"
+#include "arch/cpu.h"
+#include "arch/exceptions.h"
+#include "hal/exceptions.h"
+#include "log/log.h"
+#include "proc/elf.h"
+#include "proc/process.h"
+#include <stddef.h>
+
+#ifdef COREDUMP
```
agkaminski, line 25: Perhaps we could add whole .c or .o to the build process instead of global ifdef?

Side note - name is kinda too generic, there's a change of collision. `PROC_` prefix might be a good idea

proc/coredump.c
```diff
@@ -0,0 +1,518 @@
...
+typedef Elf32_Nhdr Elf_Nhdr;
+typedef Elf32_Addr Elf_Addr;
+#define ELFCLASS ELFCLASS32
+#else
+typedef Elf64_Ehdr Elf_Ehdr;
+typedef Elf64_Phdr Elf_Phdr;
+typedef Elf64_Nhdr Elf_Nhdr;
+typedef Elf64_Addr Elf_Addr;
+#define ELFCLASS ELFCLASS64
+#endif
```
agkaminski, line 52: Not ideal, we can often in theory run 32 bit apps on 64 bit kernel

proc/coredump.c
```diff
@@ -0,0 +1,518 @@
...
+
+	u32 b64_buf;
+	int b64_bits;
+
+	u8 rle_last;
+	size_t rle_count;
+
+	u32 crc32;
+} coredump_state_t;
+
```
agkaminski, line 94:
```suggestion


```

proc/coredump.c
```diff
@@ -0,0 +1,518 @@
...
+{
+	size_t i;
+	for (i = 0; i < len; i++) {
+		state->outBuf[state->outCur++] = data[i];
+		if (state->outCur == sizeof(state->outBuf) - 1) {
+			state->outBuf[state->outCur] = '\0';
+			coredump_write(state->outBuf, state->outCur);
+			state->outCur = 0;
+		}
+	}
```
agkaminski, line 116: memcpy would be better, we can calculate if it fits a priori

proc/coredump.c
```diff
@@ -0,0 +1,518 @@
...
+	coredump_write(textBuff, i);
+	vm_kfree(textBuff);
+}
+
+
+static void coredump_encodeChunk(coredump_state_t *state, const u8 *buf, size_t len)
+{
+	size_t i = 0;
+	size_t b;
+	while (i < len) {
```
agkaminski, line 180: This should be `for`

proc/coredump.c
```diff
@@ -0,0 +1,518 @@
...
+
+	process = proc_current()->process;
+
+	/*
+	Ensure for dumped process that:
+	- saved context is coherent with stack memory,
+	- thread count, section count are fixed,
+	while the rest of the processes can run freely.
+	*/
+	proc_freeze(process);
```
agkaminski, line 479: As we discussed - freezing shouldn't be really needed, as we have to kill the whole process anyway

proc/coredump.c
```diff
@@ -0,0 +1,518 @@
...
+			align4(sizeof(PRSTATUS_NAME)) +
+			sizeof(elf_prstatus) +
+			SIZE_COREDUMP_GREGSET +
+			SIZE_COREDUMP_THREADAUX;
+	const size_t NOTES_SIZE = SIZE_COREDUMP_GENAUX + threadCnt * (THREAD_NOTES_SIZE);
+	size_t currentOffset;
+	size_t stackSize;
+	void *userSp;
+	map_entry_t *e;
+	size_t i;
```
lukileczo, line 382: In some `#define` configurations, some variables and functions in this file are unused - please check compilation warnings with different cases

proc/coredump.c
```diff
@@ -0,0 +1,518 @@
...
+
+#endif /* COREDUMP */
+
+void coredump_dump(unsigned int n, exc_context_t *ctx)
+{
+#ifndef COREDUMP
+	return;
+#else
+
+	char buff[CORE_BUF_SIZE_MAX];
```
lukileczo, line 462: You write to this buffer by words/double words, not all targets support possibly unaligned accesses

```suggestion
	char buff[CORE_BUF_SIZE_MAX] __attribute__((aligned(8)));
```

proc/coredump.c
agkaminski: Ack

proc/coredump.c
etiaro: I understand that this is an exception from the rule about defining variables at the begining of the function, just like [here](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/6ba3e4112ed11284fa0653471f26216022391370/proc/userintr.c#L121)?

lukileczo: Probably, second option is adding `#else` and casting e.g. `(void)i` - like [here](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3e4df5def19d280a33b892f37adbe6bf2495b7fa/proc/process.c#L1052). Maybe @agkaminski has stronger preference for either solution?

agkaminski: Casting to `(void)` in `else` path is ok

etiaro: It gets a little dense with all these defines. 
Maybe it'll be better to create some static constants at the beginning of the file and writing regular if statements instead of #ifdef ?
Ultimately, we want to use some kind of runtime config, but for now there's no such thing

lukileczo: Quick thought, maybe there's a better solution:
- `#define` the options as values: `#define COREDUMP_MEM_ALL 1`, `#define COREDUMP_MEM_ALL_STACKS 2`
- add config parameter to functions, e.g. `void coredump_dump(unsigned int n, exc_context_t *ctx, int dumpOpt)`
- `#define` the `dumpOpt` - e.g. in build system `#define COREDUMP_OPT COREDUMP_MEM_ALL`
- change the `#ifdef` to regular `if`

Then you get closer to runtime configurability and remove the `#ifdefs` altogether.

etiaro: Yeah, sounds good to me

proc/coredump.c
```diff
@@ -0,0 +1,614 @@
...
+}
+
+
+void coredump_dump(unsigned int n, exc_context_t *ctx)
+{
+	char buff[CORE_BUF_SIZE_MAX] __attribute__((aligned(8)));
+	coredump_threadinfo_t threadInfo[PROC_COREDUMP_THREADS_NUM];
+	size_t segCnt;
+	size_t threadCnt;
+	coredump_state_t state;
```
adamgreloch, line 545: `buff`, `threadInfo` and `state` structs seem quite heavy for keeping them on stack. Is it possible (w.r.t. `coredump_dump()` concurrency) to move them to global `struct {} coredump_common` since their sizes are static?

proc/coredump.c
```diff
@@ -0,0 +1,614 @@
...
+	size_t i;
+
+	process = proc_current()->process;
+
+	/*
+	Ensure for dumped process that:
+	- saved context is coherent with stack memory,
+	- thread count, section count are fixed,
+	while the rest of the processes can run freely.
+	*/
```
adamgreloch:
```suggestion
	/*
	 * Ensure for dumped process that:
	 * - saved context is coherent with stack memory,
	 * - thread count, section count are fixed,
	 * while the rest of the processes can run freely.
	 */
```

proc/coredump.c
```diff
@@ -0,0 +1,614 @@
...
+ *
+ * Operating system kernel
+ *
+ * Process coredump
+ *
+ * Copyright 2025 Phoenix Systems
+ * Author: Jakub Klimek
+ *
+ * This file is part of Phoenix-RTOS.
+ *
```
adamgreloch: Nitpick, but this part is no longer necessary for new files

proc/coredump.c
etiaro: They probably wont be static in the future (when some runtime configuration will be possible).
Keeping them on stack is inspired by process_dumpException in process.c.

adamgreloch: Ack

etiaro:
Long tests with multiple processes and threads unveiled a few problems:
- spawning new threads at the same time when the process is being killed sometimes led to remaining dangling, alive threads of presumably killed process (even on master),
- spawning new threads at the same time when the process is being freezed sometimes led to issues with not being frozen and added to ready queue twice after unfreeze,
- occasionally deadlocks happened if frozen threads were enqueued for obtaining kernel locks.

These problems are solved in last two commits.

nalajcie:
> * spawning new threads at the same time when the process is being killed sometimes led to remaining dangling, alive threads of presumably killed process (even on master),

How easy is to trigger the error on master branch? If it's a common occurrence - maybe we want to fix it ASAP on `master` (and maybe backport to release branch?)

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+		int si_errno;
+	} pr_info;
+	short pr_cursig;
+	unsigned long pr_sigpend;
+	unsigned long pr_sighold;
+	pid_t pr_pid;
+	pid_t pr_ppid;
+	pid_t pr_pgrp;
+	pid_t pr_sid;
+	struct timeval {
```
agkaminski, line 68: Name collision with libc name, might cause problems in the future

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+		long tv_sec;
+		long tv_usec;
+	} pr_utime;
+	struct timeval pr_stime;
+	struct timeval pr_cutime;
+	struct timeval pr_cstime;
+	char pr_reg[0]; /* actual size depends on architecture */
+	int pr_fpvalid;
```
agkaminski, line 76: How does this work? Won't this cause `pr_fpvalid` offset to be always invalid?

hal/armv7a/exceptions.c
```diff
@@ -232,3 +238,85 @@ void _hal_exceptionsInit(void)
...
+	*(regs++) = ctx->r8;
+	*(regs++) = ctx->r9;
+	*(regs++) = ctx->r10;
+	*(regs++) = ctx->fp;
+	*(regs++) = ctx->ip;
+	*(regs++) = ctx->sp;
+	*(regs++) = ctx->lr;
+	*(regs++) = ctx->pc;
+	*(regs++) = ctx->psr;
+}
```
agkaminski, line 269: This is somewhat iffy in scope of C standard and MISRA - yes, we provide aligned buffer, but in pedantic view this is still unwanted method.

It is best to pass the desired type all the way down, in this case it depends on the architecture - perhaps per arch typedef?

On the other hand `memcpy` can be used instead, most likely less efficient approach.

Not really a merge blocker, I'll leave this for @Darchiv to decide

hal/armv7a/exceptions.c
```diff
@@ -232,3 +238,85 @@ void _hal_exceptionsInit(void)
...
+#define COMPAT_HWCAP_VFPv3 (1 << 13)
+#define HWCAP_VFPv3        (COMPAT_HWCAP_VFP | COMPAT_HWCAP_NEON | COMPAT_HWCAP_VFPv3)
+
+#define AT_HWCAP 16
+#define AT_NULL  0
+
+typedef struct {
+	u32 a_type;
+	u32 a_val;
+} elf_auxv_t;
```
agkaminski: Nit - typedef not really needed, struct scope can be reduced by declaring anon struct inside `hal_coredumpGeneralAux` body

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+	state->outCur += len;
+	if (state->outCur >= sizeof(state->outBuf) - 1) {
+		state->outBuf[sizeof(state->outBuf) - 1] = '\0';
+		coredump_write(state->outBuf, sizeof(state->outBuf));
+		state->outCur = 0;
+	}
+}
+
+
+static void coredump_b64EncodeByte(coredump_state_t *state, const u8 byte)
```
agkaminski: Something to consider - we might want to move base64 and crc32 stuff to `lib`.

If not used, should be removed during link

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+
+static size_t align4(size_t size)
+{
+	return (size + 3) & ~3;
+}
+
+
+static void coredump_dumpThreadNotes(coredump_threadinfo_t *threads, size_t threadCnt, coredump_state_t *state, void *buff)
+{
+	const u32 zero = 0;
```
agkaminski, line 308: Kinda hacky right now, we pretend to use `PRSTATUS_NAME` while the buffer is fixed 4 bytes
```suggestion
	const u8 zero[PRSTATUS_NAME] = { 0 };
```

agkaminski: Perhaps might be safer to `memset` it though

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+}
+
+
+static size_t coredump_findStack(void **currentSP, void *ustack, process_t *process)
+{
+	map_entry_t t;
+	map_entry_t *e;
+	size_t stackSize;
+
+	t.vaddr = ustack;
```
agkaminski, line 340: Why not use `currentSP` instead? `ustack` seems kinda redundant here

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+	}
+
+	phdr.p_align = 0;
+	coredump_encodeChunk(state, (u8 *)&phdr, sizeof(phdr));
+}
+
+
+static void coredump_dumpPhdr(u32 type, size_t offset, void *vaddr, size_t size, unsigned short flags, coredump_state_t *state)
+{
+	if (sizeof(void *) == 4) {
```
agkaminski: Perhaps we could wrap this condition in a function to be able to change it easily if needed

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+		coredump_dumpPhdr32(type, offset, vaddr, size, flags, state);
+	}
+	else {
+		coredump_dumpPhdr64(type, offset, vaddr, size, flags, state);
+	}
+}
+
+
+static void coredump_dumpPhdrs(coredump_threadinfo_t *threadInfo, size_t threadCnt, size_t segCnt, process_t *process, coredump_state_t *state)
```
agkaminski: Differs `coredump_dumpPhdr` by one char - bad pattern, enables mistakes

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+			if ((e->prot & PROT_READ) && (e->prot & PROT_WRITE)) {
+				coredump_dumpPhdr(PT_LOAD, currentOffset, e->vaddr, e->size, e->prot, state);
+				currentOffset += e->size;
+			}
+			e = lib_treeof(map_entry_t, linkage, lib_rbNext(&e->linkage));
+		}
+		proc_lockClear(&process->mapp->lock);
+	}
+	else if (PROC_COREDUMP_MEM_OPT == MEM_EXC_STACK) {
+		userSp = hal_cpuGetUserSP(threadInfo[0].userContext);
```
agkaminski, line 476: Why id 0 (I guess main thread), not the thread that caused exception?

proc/coredump.c
```diff
@@ -0,0 +1,612 @@
...
+	coredump_dumpThreadNotes(threadInfo, threadCnt, &state, buff);
+	hal_coredumpGeneralAux(buff);
+	coredump_encodeChunk(&state, (u8 *)buff, SIZE_COREDUMP_GENAUX);
+
+	/* MEMORY */
+	if (PROC_COREDUMP_MEM_OPT == MEM_ALL) {
+		coredump_dumpAllMemory(process, &state);
+	}
+	else if (PROC_COREDUMP_MEM_OPT == MEM_EXC_STACK) {
+		coredump_dumpStack(process, &threadInfo[0], &state);
```
agkaminski, line 591: Why id 0?

proc/process.c
```diff
@@ -274,6 +276,7 @@ void process_exception(unsigned int n, exc_context_t *ctx)
 {
 	thread_t *thread = proc_current();
 
+	coredump_dump(n, ctx);
```
agkaminski, line 279: You dereference the hell out of the thread->process here - I guess it should be after `if (thread->process == NULL)` check

proc/threads.c
```diff
@@ -2129,3 +2154,169 @@ int _threads_init(vm_map_t *kmap, vm_object_t *kernel)
...
+		if ((i != hal_cpuGetID()) && (threads_common.current[i]->process == process)) {
+			threads_common.current[i]->freeze = FREEZING;
+		}
+	}
+	current->freeze = FREEZER;
+
+	/* Threads can become frozen when rescheduled from userspace */
+	do {
+		while ((thread->freeze != FROZEN) && (thread != current)) {
+			if ((thread->state == SLEEP) && (_threads_usesKernelOnlyLock(thread) == 0)) {
```
agkaminski, line 2234: Why do this stuff with locks type etc? We can allow thread freeze only and only if either:
- threads is not in supervisor mode,
- thread is sleeping.

This assumption would make matter way simpler and we wouldn't need to classify locks

etiaro: If we freeze the thread that is sleeping, and
1. it can be waiting for a lock - then will sooner or later get that lock, and if it's still freezed, then that lock can cause deadlock, and will block other waiting threads
2. it can't be waiting for any lock - then we have to wait until that thread gets the locks its waiting for - but it can get deadlocked if that lock is "user.mutex" lock.

Another solution would be to only wait to free acquired locks, dequeue freezed threads and put them in queue on unfreeze, but the queue mechanism is used not only for locks and I didn't have an idea how to do that reasonably well that wouldn't be a harsh to maintain. Also, this solution still requires distinguishing "user.mutex", as it can never get freed.

agkaminski: You're freezing all threads, so there's no risk of deadlock regarding user mutexes.

But it's a valid point, something has to be done regarding queues. Perhaps we could use interruption mechanism (from signals) to force threads to vacate interruptible queues and freeze only in user space or on return to user space?

agkaminski: Another solution is to skip frozen  threads on dequeue, but yeah, it might get ugly quick in edge cases (e.g. how to proceed when we only have frozen thread on queue and it becomes unfrozen?).

Side note - IMHO freezing is not really needed for coredump mechanism - we can simply kill process on fault and collect thread stack as they die

proc/coredump.c
etiaro: I agree that requires a comment.
I made silent assumption that id zero is the thread that thrown exception if such exists.

It's handy, because we want this one to always be dumped in all modes, regardless any limits etc..

proc/coredump.c
etiaro: I agree that requires a comment.
I made silent assumption that id zero is the thread that thrown exception if such exists.
It's handy, because we want this one to always be dumped in all modes, regardless any limits etc..

proc/coredump.c
etiaro: It's equivalent to declaring two structures:
elf_prstatus_header,
elf_prstatus_footer,
and dumping in order `header`, `hal_coredumpGRegset()`, `footer`.

I just combined it in single struct and used pr_reg as a marker to put real arch-dependent pr_reg here in actual dump. The intention was to achieve: `&prstatus->pr_reg == &prstatus->pr_fpvalid`.
After quick read, zero-length arrays is legacy GCC extension and it's rather discouraged, not really intended to use this way, just legacy way for C99's [] array (variable length at the end of struct).

Alternatively, whole struct could be moved to HAL (which will affect buffer size and require filling its entries inside HAL function). This adds some redundancy inside HAL. 
Current solution isn't perfect either - it turned out that GDB is not working with SPARC-linux coredumps and require a hack inside its HAL to modify this struct into (luckily larger) Solaris's version of elf_prstatus.

proc/coredump.c
etiaro: It's used to provide 4-bytes alignment after PRSTATUS_NAME, which means we'll read at most 4 bytes of that zero no matter the size of PRSTATUS_NAME

proc/coredump.c
etiaro: tl;dr
It works only on GCC and it's hacky. I wonder if I should:
1. move whole struct into HAL adding redundant code in all architectures except SPARC
2. split it into header/footer and keep hacky solution in SPARC's HAL (next PR)
3. forget about this struct and only remember its size and pr_pid offset, as that's the only used (most probably probably the only useful) field for now

hal/armv7a/exceptions.c
etiaro: Regarding the previous comment, would it be any better to define `struct elf_prstatus` including pr_regs in HAL which would just be filled here, and make a cast to this new type in caller?

proc/threads.c
etiaro: > You're freezing all threads, so there's no risk of deadlock regarding user mutexes.

The exception thread can be holding this lock, if another thread is waiting to acquire it and freezer is waiting for this acquisition then that's a deadlock


> Side note - IMHO freezing is not really needed for coredump mechanism - we can simply kill process on fault and collect thread stack as they die

I'm working on dumping running processes on request. That's where freezing will really be required.
Also, the thread that gets a SIGSEGV can handle that signal and keep working (probably not recommended, but possible)

> Perhaps we could use interruption mechanism (from signals) to force threads to vacate interruptible queues

I missed that `proc_mutexLock` is interruptible. That indeed solves the problem. I'll remove the attribute and just interrupt them out of the queues.

proc/coredump.c
etiaro: It helps with handling stackoverflow case (or just invalid esp value).
If esp is outside of the expected range (memory segment containing ustack), then we include whole segment in the dump.
Also, currentSP is updated to put correct section address into PT_LOAD elf header.

Note: if esp is valid then we only dump memory from esp to the end of segment. It reduces dump size.

proc/coredump.c
etiaro: I changed the code such that this behavior is not longer hidden inside threads.c `coredump_threadsInfo()`

proc/coredump.c
etiaro: changed `coredump_dumpPhdrs` to `coredump_dumpAllPhdrs`

proc/coredump.c
etiaro: Moved to lib/encoding.h(.c)

proc/process.c
etiaro: I don't think there is any case when thread->process can be NULL apart from a few kernel-only threads.
But until kernel dumping is implemented this can go after mentioned check

etiaro:
Rewriting to userspace coredump-server and ELF core file creation.
Bugfixes moved to separate PRs

PR #651, opened 2025-05-19
proc/lock: fix owner NULL dereference

### LINKED ISSUE #650: proc/lockSet: null dereference
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/25c3b9e6400404a15561a6c3aa963ab631c1f0db/proc/threads.c#L1604

Here lock->owner can be NULL, this leads to pagefault inside _proc_threadGetPriority.

It happens relatively often when current and owner are destroyed during lock handling.

etiaro:
## Motivation and Context

Resolves #650 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: riscv64-qemu, ia32-qemu.

## Checklist:

## Special treatment

### DIFF
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -1596,6 +1596,10 @@ static int _proc_lockSet(lock_t *lock, int interruptible, spinlock_ctx_t *scp)
 		do {
 			/* _proc_lockUnlock will give us a lock by it's own */
 			if (proc_threadWaitEx(&lock->queue, &lock->spinlock, 0, interruptible, scp) == -EINTR) {
+				/* Can happen when thread_destroy is called on lock owner and current */
+				if (lock->owner == NULL) {
+					return -EINTR;
+				}
 				/* Don't return EINTR if we got lock anyway */
 				if (lock->owner != current) {
 					hal_spinlockSet(&threads_common.spinlock, &sc);


proc/threads.c
```diff
@@ -1596,6 +1596,10 @@ static int _proc_lockSet(lock_t *lock, int interruptible, spinlock_ctx_t *scp)
 		do {
 			/* _proc_lockUnlock will give us a lock by it's own */
 			if (proc_threadWaitEx(&lock->queue, &lock->spinlock, 0, interruptible, scp) == -EINTR) {
+				/* Can happen when thread_destroy is called on lock owner and current */
+				if (lock->owner == NULL) {
+					return -EINTR;
+				}
```
agkaminski, line 1602: IMHO the core of the problem is that we clear the `threads_common.spinlock` in line 1594, which can lead to a race on a SMP system.

So I think this check should be performed under `threads_common.spinlock` in the line 1606 instead. Current solution just makes the race less probable

agkaminski: Eh, I guess the whole `if (lock->owner != current)` should be under spinlock too, I think

agkaminski (approved):
Perhaps move the condition later and set the priority when `lock->owner != NULL` instead. Personal preference I guess, so just an option to consider.

etiaro:
That's my first though too, but I didn't want to add even more layers of nesting here

PR #656, opened 2025-05-26
ioctl: serialize special ioctls outside of kernel

julianuziemblo:
## Description
For the kernel, the ioctl syscall data should be transparent, so special ioctls (e.g. ones containing a pointer in the input structure) should be serialized/deserialized in userspace. Additional size parameter is passed to allow for not-const sized ioctl payloads (e.g. for SIOCGIFCONF).
Related (serialization in userspace): https://github.com/phoenix-rtos/libphoenix/pull/414

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change):
    - change signature of ioctl syscall proxy function: `int sys_ioctl(int fildes, unsigned long request, void *val, size_t size)`
    This was done to pass the size of the size of the whole structure to the kernel without affecting the `request` variable, so we can still match on it in the device server handling the ioctl. This eliminates the need to handle special cases in the kernel: userspace serializes the ioctl, and device server unpacks it.

## How Has This Been Tested?

- [x] Tested by hand on: `armv7m7-imxrt106x-evk`, `ia32-generic-qemu`

## Checklist:

    - requires change to documentation regarding ioctl syscall (accepting 4th argument `size`)

## Special treatment

    - TODO

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -1617,19 +1617,11 @@ int posix_fcntl(int fd, unsigned int cmd, char *ustack)
 #define IOC_OUT                      0x40000000
 #define IOC_IN                       0x80000000
 #define IOC_INOUT                    (IOC_IN | IOC_OUT)
-#define _IOC(inout, group, num, len) ((unsigned long)(inout | ((len & IOCPARM_MASK) << 16) | ((group) << 8) | (num)))
-
-#define SIOCGIFCONF _IOC(IOC_INOUT, 'S', 0x12, sizeof(struct ifconf))
-#define SIOCADDRT   _IOC(IOC_IN, 'S', 0x44, sizeof(struct rtentry))
-#define SIOCDELRT   _IOC(IOC_IN, 'S', 0x45, sizeof(struct rtentry))
 
 
-static void ioctl_pack(msg_t *msg, unsigned long request, void *data, oid_t *oid)
+static void ioctl_pack(msg_t *msg, unsigned long request, void *data, size_t size, oid_t *oid)
 {
-	size_t size = IOCPARM_LEN(request);
 	ioctl_in_t *ioctl = (ioctl_in_t *)msg->i.raw;
-	struct ifconf *ifc;
-	struct rtentry *rt;
 
 	hal_memcpy(&msg->oid, oid, sizeof(*oid));
 	msg->type = mtDevCtl;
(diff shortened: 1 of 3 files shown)


agkaminski (commented):
LGTM, would like @Darchiv to approve

posix/posix.c
```diff
@@ -1715,14 +1688,16 @@ int posix_ioctl(int fildes, unsigned long request, char *ustack)
 			default:
 				if (((request & IOC_INOUT) != 0) || (IOCPARM_LEN(request) > 0)) {
 					GETFROMSTACK(ustack, void *, data, 2);
+					GETFROMSTACK(ustack, size_t, size, 3);
```
Darchiv: As we have the size of `data` now, `data` pointer can be checked with `vm_mapBelongs()` (see other syscalls for reference). This is an optional syscall parameter, so it should probably be done here - in this case include a comment in `syscalls_sys_ioctl()` so that it is apparent you took care of securing this syscall.

posix/posix.c
```diff
@@ -1643,22 +1643,16 @@ int posix_fcntl(int fd, unsigned int cmd, u8 *ustack)
...
 
-#define IOC_OUT                      0x40000000UL
-#define IOC_IN                       0x80000000UL
-#define IOC_INOUT                    (IOC_IN | IOC_OUT)
-#define _IOC(inout, group, num, len) ((unsigned long)((inout) | (((len) & IOCPARM_MASK) << 16) | (((unsigned int)(group)) << 8) | (num)))
+#define IOC_VOID   0x00000000UL
+#define IOC_NESTED 0x20000000UL
+#define IOC_OUT    0x40000000UL
+#define IOC_IN     0x80000000UL
+#define IOC_INOUT  (IOC_IN | IOC_OUT)
```
Darchiv: I have just seen that libphoenix and kernel duplicate these macros, ugh. The _phoenix-rtos-kernel/include_ directory is installed for libphoenix to be used via `#include <phoenix/*.h>`, so both components can use common definitions. It makes sense to have ABI-related definitions in a single shared maintenance point. (For example, see: posix-statvfs.h)

Maybe move these macros to _phoenix-rtos-kernel/include/ioctl.h_ and in _include/sys/ioctl.h_ just include it? Not sure how many of them - for sure ABI-defining (with bit positions), so maybe just `IOC*`, `IOC_*`, and `_IOC`, while e.g. `_IO` can be left in libphoenix

PR #556, opened 2024-06-04
proc: Rename kernel thread to kthread instead of idle

agkaminski:
to avoid confusion when using 'ps' applet

DONE: RTOS-845

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

- [x] Already covered by automatic testing.
- [ ] New test added: (add PR link here).
- [ ] Tested by hand on: (list targets here).

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing linter checks and tests passed.
- [x] My changes generate no new compilation warnings for any of the targets.

## Special treatment

- [ ] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
- [ ] I will merge this PR by myself when appropriate.

PR #661, opened 2025-05-30
!perf: introduce performance analysis subsystem

adamgreloch:
Introduces a subsystem intended for various performance analysis
mechanisms: tracing, sample-based methods, MPU utilization, etc.

Breaking change: This commit generalizes the `perf_*` syscalls to
support various methods desribed above. Previous perf semantics are
preserved in the `perf_mode_threads` mode.

The subsystem currently implements `perf_mode_trace` mode - a mechanism
for tracing syscalls, interrupts, locks and scheduler events. The events
are outputed in CTF format to a cbuffer that can be retrieved from
userspace using `perf_read(perf_mode_trace)` and later processed by 3rd
party libraries like babeltrace2.

JIRA: RTOS-1057

## Types of changes

## How Has This Been Tested?

- [X] Tested by hand on: ia32-generic-qemu

## Checklist:

## Special treatment


proc/lock.h
```diff
@@ -59,4 +63,10 @@ extern int proc_lockInit(lock_t *lock, const struct lockAttr *attr, const char *
 extern int proc_lockDone(lock_t *lock);
 
 
+static inline int proc_lockGetId(const lock_t *lock)
+{
+	return lock->idlinkage.id;
+}
```
agkaminski: Why is this needed? `lock` is a per process resource, not a global one. lock_id:process_id provides unique identifier

adamgreloch: After discussion, we have decided to use lock address as unique id

perf/buffer-mem.c
```diff
@@ -0,0 +1,125 @@
...
+
+
+struct {
+	vm_map_t *kmap;
+
+	time_t lastTimestamp;
+	cbuffer_t buffer;
+	page_t *pages;
+
+	time_t prev;
```
etiaro: what are `lastTimestamp` and `prev` for?

perf/buffer-mem.c
etiaro, line 1: Why is name of this file different than header `buffer.h`?

proc/process.c
```diff
@@ -29,6 +29,7 @@
 #include "msg.h"
 #include "ports.h"
 #include "userintr.h"
+#include "perf/events.h"
```
etiaro: Is this necessary?

proc/threads.c
```diff
@@ -668,9 +674,17 @@ int threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
 {
 	spinlock_ctx_t sc;
 	int ret;
+
+	int cpuId = hal_cpuGetID();
+
+	perf_traceEventsSchedEnter(cpuId);
```
etiaro: Shouldn't this be inside _threads_schedule?
On some targets (eg ia32), there's a direct call to _threads_schedule from hal_cpuReschedule implementation.
I don't know how often it happens (there's a threads_schedule call too)

perf/trace.c
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+
+#include "buffer.h"
+#include "events.h"
+#include "trace.h"
+
+
+struct {
+	time_t lastTimestamp;
```
etiaro: Is `lastTimestamp` necessary?

perf/trace.c
```diff
@@ -0,0 +1,159 @@
...
+
+	if (trace_common.gather == 0) {
+		return -EINVAL;
+	}
+
+	if (trace_common.nonMonotonicity != 0) {
+		lib_printf("kernel (%s:%d): timer non-monotonicity detected during event gathering", __func__, __LINE__);
+	}
+
+	if (trace_common.bufferFull != 0) {
```
etiaro: I guess `gather`, `nonMonotonicity`, `bufferFull` should be protected by spinlock

proc/process.c
adamgreloch: Not anymore, thanks

perf/buffer-mem.c
adamgreloch: There may be multiple `buffer.h` implementations, i.e. `buffer-rtt.c` that pushes the events via RTT instead

perf/trace.c
adamgreloch: `gather` is a specific case, but I agree about the rest

proc/threads.c
adamgreloch: Yeah, you're right, thanks

hal/arm/rtt.c
```diff
@@ -140,10 +148,20 @@ int _hal_rttReset(unsigned int chan, rtt_dir_t dir)
 }
 
 
+int _hal_rttInitialized(void)
+{
+	return common.rtt != NULL;
```
etiaro, line 157: Is that guaranteed to be NULL?

include/perf.h
```diff
@@ -0,0 +1,28 @@
...
+#define _PHOENIX_PERF_H_
+
+
+/* clang-format off */
+typedef enum { perf_mode_trace, perf_mode_count } perf_mode_t;
+typedef enum { trace_channel_meta, trace_channel_event, trace_channel_count } trace_channel_t;
+/* clang-format on */
+
+
+#define PERF_TRACE_FLAG_ROLLING (1 << 1) /* treat event channel as rolling window */
```
etiaro: nitpick: first bit is (1 << 0)

proc/threads.c
```diff
@@ -2111,6 +1893,19 @@ int proc_threadsOther(thread_t *t)
 }
 
 
+void proc_threadsListCb(void *arg, int i, threadinfo_t *tinfo)
```
etiaro: This should be static

proc/threads.c
```diff
@@ -104,289 +104,50 @@ static int _proc_threadWakeup(thread_t **queue);
...
 
-static unsigned perf_idpack(unsigned id)
-{
-	return id >> 8;
-}
-
-
 /* Note: always called with threads_common.spinlock set */
-static void _perf_event(thread_t *t, int type)
+static void _updateWaits(thread_t *t, int type)
```
etiaro: Why is the module name prefix removed in all these functions?

perf/trace-events.h
```diff
@@ -0,0 +1,289 @@
...
+		ev.tid = proc_getTid(t);
+		ev.priority = t->priority;
+
+		if (t->process != NULL) {
+			ev.pid = process_getPid(t->process);
+			process_getName(t->process, ev.name, sizeof(ev.name));
+		}
+		else {
+			ev.pid = 0;
+			hal_memcpy(ev.name, "[idle]", sizeof("[idle]"));
```
etiaro: There can be kernel threads other than idle (eg usrv)

perf/trace.c
```diff
@@ -0,0 +1,341 @@
...
+	 * Treat `gather` as atomic to reduce overhead on the kernel when the tracing is disabled
+	 * - there is only one writer at a time (perf_trace{Start,Finish}()) and multiple readers
+	 * (trace events doing trace_isRunning()). Due to eventual consistency in the readers
+	 * we may lose some events, but we may lose them anyway as the invocation of
+	 * trace_start() naturally races with kernel events occurring in the meantime.
+	 *
+	 * `running` is an always consistent copy of `gather` used for proper synchronization of
+	 * tracing state.
+	 */
+	volatile int gather;
```
etiaro: `running` with added volatile can be used for checks without spinlock, which makes this duplication unnecessary

proc/threads.c
adamgreloch: They are static, so the module name is unnecessary. They are not really a part of `perf` anymore and serve just as hooks for invoking arbitrary code when a specific scheduler event occurs

adamgreloch: Upon further discussion, let's stay consistent with the rest of the kernel code. Added `_threads` prefix

include/syscalls.h
```diff
@@ -124,5 +124,7 @@
 	ID(sys_mprotect) \
 	\
 	ID(sys_statvfs) \
-	ID(sys_uname)
+	ID(sys_uname) \
+	\
+	ID(perf_stop)
```
Darchiv: Since this is a breaking change I think `perf_stop` can be moved to its sibling (`perf_*`) syscalls

syscalls.c
```diff
@@ -21,12 +21,16 @@
...
 #include "include/utsname.h"
+#include "include/perf.h"
 #include "lib/lib.h"
 #include "proc/proc.h"
 #include "vm/object.h"
 #include "posix/posix.h"
 #include "syspage.h"
 
+#include "perf/perf.h"
+#include "perf/trace-events.h"
```
Darchiv, line 35: Why separate these includes? They are a module just like lib/, proc/, vm/ etc.

syscalls.c
```diff
@@ -482,11 +486,22 @@ int syscalls_syspageprog(void *ustack)
...
 {
-	unsigned pid;
+	process_t *proc = proc_current()->process;
+	perf_mode_t mode;
+	unsigned flags;
+	void *arg;
+	size_t sz;
 
-	GETFROMSTACK(ustack, unsigned, pid, 0);
+	GETFROMSTACK(ustack, perf_mode_t, mode, 0);
```
Darchiv: I'm not convinced about enums as syscall arguments due to potential ABI incompatibilities, but maybe we should not assume that enums may **not** be `int` (due to `-fshort-enums` or others)

perf/trace-events.h
```diff
@@ -0,0 +1,289 @@
...
+	TRACE_EVENT_SYSCALL_EXIT = 0x29,
+	TRACE_EVENT_SCHED_ENTER = 0x2a,
+	TRACE_EVENT_SCHED_EXIT = 0x2b,
+	TRACE_EVENT_LOCK_NAME = 0x2c,
+	TRACE_EVENT_LOCK_SET_ENTER = 0x2d,
+	TRACE_EVENT_LOCK_SET_ACQUIRED = 0x2e,
+	TRACE_EVENT_LOCK_SET_EXIT = 0x2f,
+	TRACE_EVENT_LOCK_CLEAR = 0x30,
+	TRACE_EVENT_THREAD_PRIORITY = 0x31,
+	TRACE_EVENT_PROCESS_KLL = 0x32,
```
Darchiv: `KILL`?

perf/tsdl/metadata
```diff
@@ -0,0 +1,188 @@
...
+};
+
+event {
+	name = thread_priority;
+	id = 0x31;
+	fields := struct {
+		u16 tid;
+		u8 priority;
+	};
+};
```
Darchiv, line 188: Definition for `TRACE_EVENT_PROCESS_KLL` missing

perf/trace-events.h
```diff
@@ -0,0 +1,289 @@
...
+		ev.tid = proc_getTid(t);
+		ev.priority = t->priority;
+
+		if (t->process != NULL) {
+			ev.pid = process_getPid(t->process);
+			process_getName(t->process, ev.name, sizeof(ev.name));
+		}
+		else {
+			ev.pid = 0;
+			hal_memcpy(ev.name, "[kthr]", sizeof("[kthr]"));
```
Darchiv: We should probably agree on naming of various system entities. Is `[kthr]` already used anywhere? If we merge https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/556, then should this also be `[kthread]`?

perf/trace.c
```diff
@@ -0,0 +1,336 @@
...
+
+
+#define TRACE_NON_MONOTONICITY (1 << 1)
+#define TRACE_EVENT_DELAYED    (1 << 2)
+#define TRACE_BUFFER_WRITE_ERR (1 << 3)
+
+
+static u32 _gettimeRaw(void)
+{
+	u32 now = hal_timerGetUs();
```
Darchiv: I assume that `u32` is used in event frames for a timestamp and that is the reason to down-size the result of `time_t hal_timerGetUs()`?

adamgreloch: Correct

perf/trace-events.h
adamgreloch: Yes, it should be `[kthread]` then. The initial `ctf_to_perfetto` converter used this thread as a fallback for any kernel events (IRQs, sched, etc.), so the `[idle]` seemed misleading to me. In the current version it doesn't matter anymore (converter manages entity naming in its own way).

syscalls.c
adamgreloch: Subconsciously, to emphasize that these are purely "meta" includes, not really a part of the kernel code as they just instrument it. I can remove the newline if you wish

syscalls.c
adamgreloch: Fair, changed `mode` to `int`

perf/trace-events.h
adamgreloch: Upon discussion, changed to `[kernel]` to remove any mental connection to actual kernel thread naming

Darchiv (approved):

hal/sparcv8leon/gaisler/irqamp.c
```diff
@@ -232,6 +232,12 @@ char *hal_interruptsFeatures(char *features, size_t len)
 }
 
 
+void _hal_interruptsTrace(int enable)
+{
+	/* TODO */
```
Darchiv, line 237: Maybe indicate (in commit msg) that interrupts are not traced on gaisler?

adamgreloch: Done

PR #664, opened 2025-06-25
hal: Bulletproof kernel operations on user memory

etiaro:
## Description

Inspired by BSD/openQNX exception handling routines, this draft introduces a mechanism with semantics similar to try - catch and based on idea similar to setjmp/longjmp by introducing 
- `hal_createexcjmp` that prepares "try-catch" context to be saved in TCB
- `threads_setexcjmp`/`threads_getexcjmp` which manages access to dedicated `thread_t` struct field

Introduced a few basic function and preprocessor macro for easy and safe use of this exception handling mechanism. 

On NOMMU targets, thanks to fixed configuration of accessible memory, simple access checks are good enough. That being said protection (R/W/E) checks are included in this PR to ensure no illegal accesses via kernel.

Here is an example AI-generated test case that triggers hal_excjmpInstall (and cascades CPU faults on master): [test_usermem_toctou.c](https://github.com/user-attachments/files/31069023/test_usermem_toctou.c)
TBD: phoenix-rtos-tests PR with TOCTOU test cases

## Motivation and Context

There are some events in kernel code that require operating on user's memory., which is vulnerable to TOCTOU class races.
There's no good way to ensure such memory operations won't trigger a processor exception apart from interrupting all threads that could access that memory fragment.

These mechanisms can also be used in the future to gently handle user stackoverflows and faulty signal trampolines that currently can become a cascade of kernel exceptions.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/_interrupts.S
+++ b/hal/armv7a/_interrupts.S
@@ -296,3 +296,19 @@ _syscalls_dispatch:
 	add sp, sp, #8
 	b _hal_cpuRestoreCtx
 .size _syscalls_dispatch, .-_syscalls_dispatch
+
+
+/* int hal_createexcjmp(excjmp_context_t *ctx) */
+.globl hal_createexcjmp
+.type hal_createexcjmp, %function
+hal_createexcjmp:
+	mrs r2, cpsr
+	str r2, [r0, #0]
+	str lr, [r0, #4]
+	add r1, r0, #8
+	stm r1, {r4-r11}
+	str sp, [r0, #40]
+
+	mov r0, #0
+	bx lr
+.size hal_createexcjmp, .-hal_createexcjmp
--- a/hal/armv7a/cpu.c
+++ b/hal/armv7a/cpu.c
@@ -15,6 +15,8 @@
 #include "hal/string.h"
 #include "hal/spinlock.h"
 #include "hal/hal.h"
+#include "arch/exceptions.h"
+#include "include/errno.h"
 
 #include "armv7a.h"
 #include "config.h"
--- a/hal/armv7a/exceptions.c
+++ b/hal/armv7a/exceptions.c
@@ -35,6 +35,8 @@
 #define EXC_DEBUG               0x02U
 #define EXC_ALIGMENT            0x01U
 
+#define PSR_THUMB_STATE (1U << 5)
+
 
 static struct {
 	excHandlerFn_t undefHandler;
@@ -113,6 +115,33 @@ static void exceptions_defaultHandler(unsigned int n, exc_context_t *ctx)
 		hal_cpuHalt();
 	}
 #endif
+}
+
+
+int hal_excjmpInstall(volatile excjmp_context_t *jmpctx, exc_context_t *excctx)
+{
+	cpu_context_t *ctx = &excctx->cpuCtx;
+	ctx->psr = jmpctx->psr;
+	if ((jmpctx->ret & 1U) != 0U) {
+		ctx->psr |= PSR_THUMB_STATE;
+	}
+	else {
+		ctx->psr &= ~PSR_THUMB_STATE;
+	}
+	ctx->r4 = jmpctx->r4;
+	ctx->r5 = jmpctx->r5;
+	ctx->r6 = jmpctx->r6;
+	ctx->r7 = jmpctx->r7;
+	ctx->r8 = jmpctx->r8;
+	ctx->r9 = jmpctx->r9;
+	ctx->r10 = jmpctx->r10;
+	ctx->fp = jmpctx->fp;
+	ctx->pc = jmpctx->ret & ~1U;
+	ctx->sp = jmpctx->sp;
+
+	/* Mark fault happened in return value */
+	ctx->r0 = 1;
+	return 0;
 }
 
 
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -355,6 +355,28 @@ __attribute__((noreturn)) void proc_longjmp(cpu_context_t *ctx)
 }
 
 
+#ifdef EXCJMP_SUPPORTED
+void threads_setexcjmp(volatile excjmp_context_t *ctx)
+{
+	thread_t *current = proc_current();
+	if (current == NULL) {
+		return;
+	}
+	current->excjmpctx = ctx;
+}
+
+
+volatile excjmp_context_t *threads_getexcjmp(void)
+{
+	thread_t *current = proc_current();
+	if (current == NULL) {
+		return NULL;
+	}
+	return current->excjmpctx;
+}
+#endif
+
+
 static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context_t *signalCtx, unsigned int oldmask, const int src);
 
 
@@ -595,6 +617,9 @@ int proc_threadCreate(process_t *process, startFn_t start, int *id, u8 priority,
 	t->startTime = t->readyTime;
 	t->lastTime = t->readyTime;
 	t->longjmpctx = NULL;
+#ifdef EXCJMP_SUPPORTED
+	t->excjmpctx = NULL;
+#endif
 
 	if (process != NULL && (process->tls.tdata_sz != 0U || process->tls.tbss_sz != 0U)) {
 		err = process_tlsInit(&t->tls, &process->tls, process->mapp);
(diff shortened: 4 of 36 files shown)


hal/armv7a/_interrupts.S
```diff
@@ -80,6 +80,13 @@ _exceptions_dispatch:
 	 * r0 - exception number
 	 * r2 - PC of the instruction that caused the exception
 	 * registers r0-r4 from previous context are saved below SP */
+
+    push {r0}
+    bl threads_getexcjmp
+    cmp r0, #0
+    bne hal_excjmp
+    pop {r0}
```
agkaminski: Perhaps we could perform this check in C? If we're jumping from it anyway, we could just safely drop the exception context, so it should not hurt to do this this way.

side note - you are exiting in the wrong state here! On Cortex-A each exception type has it's own banked registers (along with `sp`) - we need to switch to `SYS_MODE` before leaving!

hal/armv7a/cpu.c
```diff
@@ -110,19 +114,26 @@ int hal_cpuPushSignal(void *kstack, void (*handler)(void), cpu_context_t *signal
...
+		}
+		else {
+			signalCtx->psr &= ~THUMB_STATE;
+		}
+
+		hal_stackPutArgs((void **)&signalCtx->sp, sizeof(args) / sizeof(args[0]), args);
 	}
 	else {
-		signalCtx->psr &= ~THUMB_STATE;
+		threads_setexcjmp(oldctx, NULL);
```
agkaminski: Perhaps we could clear old ctx on `hal_excjmp`

proc/threads.c
```diff
@@ -35,11 +36,15 @@ struct {
 	vm_map_t *kmap;
 	spinlock_t spinlock;
 	lock_t lock;
+
+	/* Synchronized by spinlock */
 	thread_t *ready[8];
+	/*
+	 * Write to current[i] only by i'th cpu, with spinlock set.
+	 * Read from current[i] by i'th cpu allowed without spinlock.
+	 */
```
agkaminski: Spinlock does nothing if it's used only for writes - let's just assume atomicity (and make these **pointers** volatile)

proc/threads.c
```diff
@@ -1464,16 +1469,18 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
 
 	unsigned int sig;
+	int ret;
 
 	sig = (selected->sigpend | proc->sigpend) & ~selected->sigmask;
 	if ((sig != 0) && (proc->sighandler != NULL)) {
 		sig = hal_cpuGetLastBit(sig);
 
-		if (hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sighandler, signalCtx, sig, oldmask, src) == 0) {
+		if ((ret = hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sighandler, signalCtx, sig, oldmask, src)) == 0) {
```
agkaminski:
```suggestion
		ret = hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sighandler, signalCtx, sig, oldmask, src);
		if (ret == 0) {
```

proc/threads.c
```diff
@@ -1491,18 +1498,29 @@ void threads_setupUserReturn(void *retval, cpu_context_t *ctx)
...
-	thread = _proc_current();
+	thread = proc_current();
 
 	kstackTop = thread->kstack + thread->kstacksz;
 	signalCtx = (void *)((char *)hal_cpuGetUserSP(ctx) - sizeof(*signalCtx));
 	hal_cpuSetReturnValue(ctx, retval);
 
-	if (_threads_checkSignal(thread, thread->process, signalCtx, thread->sigmask, SIG_SRC_SCALL) == 0) {
+	// hal_spinlockClear(&threads_common.spinlock, &sc);  // TODO first load sigmask etc
+	int ret = _threads_checkSignal(thread, thread->process, signalCtx, thread->sigmask, SIG_SRC_SCALL);
```
agkaminski: Declare variables at the start (kernel, plo)

proc/threads.c
```diff
@@ -1491,18 +1498,29 @@ void threads_setupUserReturn(void *retval, cpu_context_t *ctx)
...
 		f = thread->process->sighandler;
 		hal_spinlockClear(&threads_common.spinlock, &sc);
 		hal_jmp(f, kstackTop, hal_cpuGetUserSP(signalCtx), 0, NULL);
 		/* no return */
 	}
+	else if (ret == -EFAULT) {
+#ifndef NDEBUG
+		hal_consolePrint(ATTR_BOLD, "EFAULT in _threads_checkSignal, killing whole process\n");
+#endif
+		hal_spinlockClear(&threads_common.spinlock, &sc);
```
agkaminski: spinlock double unlock on line 1525

etiaro: That thread is effectively dying on reschedule, but you're right, a return won't make a harm.

proc/threads.c
etiaro: It's not used only for writes - it's used for reads on other cores.
The idea was that on the i'th core it's not required for reading because it cannot be writing at the same time as it's reading.

hal/armv7a/cpu.c
etiaro: That's problematic, as it would require putting pointer to oldctx in struct `excjmp_context_t` (as `hal_excjmp` doesn't have access to local variables from this context).
And we need to call threads_setexcjmp with oldctx after else anyways.

hal/ia32/cpu.c
```diff
@@ -186,13 +188,21 @@ int hal_cpuPushSignal(void *kstack, void (*handler)(void), cpu_context_t *signal
 
 	(void)src;
 
-	hal_memcpy(signalCtx, ctx, sizeof(cpu_context_t));
 
-	/* parasoft-suppress-next-line MISRAC2012-RULE_11_1 "Need to assign function address to processor register" */
-	signalCtx->eip = (u32)handler;
-	signalCtx->esp -= sizeof(cpu_context_t);
+	if (!hal_setexcjmp(&excctx, &oldctx)) {
+		hal_memcpy(signalCtx, ctx, sizeof(cpu_context_t));
```
etiaro: Additional check for access to kernel memory map should be added

syscalls.c
```diff
@@ -915,24 +916,22 @@ int syscalls_msgRespond(u8 *ustack)
...
 	GETFROMSTACK(ustack, u32, port, 0U);
 	GETFROMSTACK(ustack, msg_t *, msg, 1U);
 	GETFROMSTACK(ustack, msg_rid_t, rid, 2U);
 
-	if (vm_mapBelongs(proc, msg, sizeof(*msg)) < 0) {
+	if (vm_mapBelongs(proc, msg, sizeof(*msg), PROT_READ) < 0) {
 		return -EFAULT;
 	}
 
-#ifndef NOMMU /* o.data has client memory pointer on NOMMU */
```
ziemleszcz, line 933: What about `NOMMU_MAPBELONGS_VALIDATE_MAP` that you have added earlier?

hal/armv7a/exceptions.c
```diff
@@ -118,12 +120,46 @@ static void exceptions_defaultHandler(unsigned int n, exc_context_t *ctx)
...
 void threads_setupUserReturn(void *retval, cpu_context_t *ctx);
 
 
+excjmp_context_t *threads_getexcjmp(void);
+
+
 /* parasoft-suppress-next-line MISRAC2012-RULE_8_4 "Usage in assembly" */
 void exceptions_dispatch(unsigned int n, exc_context_t *ctx)
 {
+	excjmp_context_t *excctx = threads_getexcjmp();
```
ziemleszcz: Shouldn't this be called from `map_pageFault()` on `vm_mapForce()` failure for USERMEM_TRY section?

etiaro: Yeah, you're right. Previous solution would break on lazy allocation (if it would be enabled), now it should be fine, however exception dump before those checks from the master is kept.

syscalls.c
etiaro: This ifndef seems outdated to me. 
With that flag, the check is still reasonable (if the process has write access to the output buffer it will return EOK, if the process does not have the access, the output buffer would be copied into his memory in msg-nommu.c).

Before, or without that flag, mapBelongs implementation was disabled on NOMMU for this very reason, so this seems like a double-disable.

PR #668, opened 2025-07-10
proc: Add futexes

kamilaboba123:
Implement futexes

## Description
Implement futexes, add new syscalls: futex_wait(), futex_wakeup()

## Motivation and Context
Futexes fix performance issues with the current mutex implementation, by reducing
the amount of syscalls.

## Types of changes

## How Has This Been Tested?

ia32-generic-qemu

## Checklist:

## Special treatment

### DIFF
--- a/include/syscalls.h
+++ b/include/syscalls.h
@@ -124,5 +124,8 @@
 	ID(sys_mprotect) \
 	\
 	ID(sys_statvfs) \
-	ID(sys_uname)
+	ID(sys_uname) \
+    \
+    ID(futex_wait) \
+    ID(futex_wakeup)
 /* clang-format on */
--- a/proc/futex.c
+++ b/proc/futex.c
@@ -31,7 +31,6 @@ int futex_wait(unsigned int *address, unsigned int value, time_t timeout)
 	LIST_ADD(&sleepqueue->futex_list, &futex);
 	proc_lockClear(&process->lock);
 
-	lib_printf("address = %p, *address = %u, value = %u, timeout = %d\n", address, *address, value, timeout);
 	if (*address != value) {
 		proc_lockSet(&process->lock);
 		LIST_REMOVE(&sleepqueue->futex_list, &futex);
@@ -39,7 +38,7 @@ int futex_wait(unsigned int *address, unsigned int value, time_t timeout)
 		return -EAGAIN;
 	}
 
-	int err = proc_threadSleep(timeout * 1000 * 1000);
+	int err = proc_threadSleep2(futex.waiting_thread, timeout * 1000 * 1000);
 	if (err < 0) {
 		proc_lockSet(&process->lock);
 		LIST_REMOVE(&sleepqueue->futex_list, &futex);
@@ -50,8 +49,6 @@ int futex_wait(unsigned int *address, unsigned int value, time_t timeout)
 
 int futex_wakeup(unsigned int *address, unsigned int n_threads)
 {
-	lib_printf("address = %p, *address = %u, n_threads = %u\n", address, *address, n_threads);
-
 	if (n_threads == 0) {
 		return 0;
 	}
@@ -73,7 +70,6 @@ int futex_wakeup(unsigned int *address, unsigned int n_threads)
 		if (i == n_threads) {
 			break;
 		}
-		lib_printf("WAKEUP\n");
 		int err = proc_threadWakeup(&f->waiting_thread);
 		if (err < 0) {
 			return err;
--- a/proc/futex.c
+++ b/proc/futex.c
@@ -1,0 +1,85 @@
+#include "hal/hal.h"
+#include "lib/lib.h"
+#include "include/errno.h"
+#include "process.h"
+#include "threads.h"
+#include "futex.h"
+
+static inline futex_sleepqueue_t *futex_getSleepQueue(process_t *process, futex_t *f)
+{
+    unsigned int key = f->address >> 3;
+    key ^= key >> FUTEX_SLEEPQUEUES_BITS;
+    return &process->futex_sleepqueues[key & FUTEX_SLEEPQUEUES_MASK];
+}
+
+int futex_wait(unsigned int *address, unsigned int value, time_t timeout)
+{
+    thread_t *thread = proc_current();
+    process_t *process = thread->process;
+
+    futex_t futex;
+    hal_memset(&futex, 0, sizeof(futex));
+    futex_sleepqueue_t *sleepqueue = NULL;
+
+    addr_t off = (addr_t)address;
+    futex.address = off;
+
+    sleepqueue = futex_getSleepQueue(process, &futex);
+
+    futex.waiting_thread = thread;
+    proc_lockSet(&process->lock);
+    LIST_ADD(&sleepqueue->futex_list, &futex);
+    proc_lockClear(&process->lock);
+
+    lib_printf("address = %p, *address = %u, value = %u, timeout = %d\n", address, *address, value, timeout);
+    if (*address != value) {
+        proc_lockSet(&process->lock);
+        LIST_REMOVE(&sleepqueue->futex_list, &futex);
+        proc_lockClear(&process->lock);
+        return -EAGAIN;
+    }
+
+    int err = proc_threadSleep(timeout * 1000 * 1000);
+    if (err < 0) {
+        proc_lockSet(&process->lock);
+        LIST_REMOVE(&sleepqueue->futex_list, &futex);
+        proc_lockClear(&process->lock);
+    }
+    return err;
+}
+
+int futex_wakeup(unsigned int *address, unsigned int n_threads)
+{
+    lib_printf("address = %p, *address = %u, n_threads = %u\n", address, *address, n_threads);
+
+    if (n_threads == 0) {
+        return 0;
+    }
+    
+    thread_t *thread = proc_current();
+    process_t *process = thread->process;
+
+    futex_t futex;
+    hal_memset(&futex, 0, sizeof(futex));
+    
+    addr_t off = (addr_t)address;
+    futex.address = off;
+
+    futex_sleepqueue_t *sleepqueue = futex_getSleepQueue(process, &futex);
+
+    futex_t *f = sleepqueue->futex_list;
+    int i = 0;
+    while (f != NULL) {
+        if (i == n_threads) {
+            break;
+        }
+        lib_printf("WAKEUP\n");
+        int err = proc_threadWakeup(&f->waiting_thread);
+        if (err < 0) {
+            return err;
+        }
+        f = f->next, i++;
+    }
+    return i;
+}
+
--- a/proc/futex.c
+++ b/proc/futex.c
@@ -7,79 +7,78 @@
 
 static inline futex_sleepqueue_t *futex_getSleepQueue(process_t *process, futex_t *f)
 {
-    unsigned int key = f->address >> 3;
-    key ^= key >> FUTEX_SLEEPQUEUES_BITS;
-    return &process->futex_sleepqueues[key & FUTEX_SLEEPQUEUES_MASK];
+	unsigned int key = f->address >> 3;
+	key ^= key >> FUTEX_SLEEPQUEUES_BITS;
+	return &process->futex_sleepqueues[key & FUTEX_SLEEPQUEUES_MASK];
 }
 
 int futex_wait(unsigned int *address, unsigned int value, time_t timeout)
 {
-    thread_t *thread = proc_current();
-    process_t *process = thread->process;
+	thread_t *thread = proc_current();
+	process_t *process = thread->process;
 
-    futex_t futex;
-    hal_memset(&futex, 0, sizeof(futex));
-    futex_sleepqueue_t *sleepqueue = NULL;
+	futex_t futex;
+	hal_memset(&futex, 0, sizeof(futex));
+	futex_sleepqueue_t *sleepqueue = NULL;
 
-    addr_t off = (addr_t)address;
-    futex.address = off;
+	addr_t off = (addr_t)address;
+	futex.address = off;
 
-    sleepqueue = futex_getSleepQueue(process, &futex);
+	sleepqueue = futex_getSleepQueue(process, &futex);
 
-    futex.waiting_thread = thread;
-    proc_lockSet(&process->lock);
-    LIST_ADD(&sleepqueue->futex_list, &futex);
-    proc_lockClear(&process->lock);
+	futex.waiting_thread = thread;
+	proc_lockSet(&process->lock);
+	LIST_ADD(&sleepqueue->futex_list, &futex);
+	proc_lockClear(&process->lock);
 
-    lib_printf("address = %p, *address = %u, value = %u, timeout = %d\n", address, *address, value, timeout);
-    if (*address != value) {
-        proc_lockSet(&process->lock);
-        LIST_REMOVE(&sleepqueue->futex_list, &futex);
-        proc_lockClear(&process->lock);
-        return -EAGAIN;
-    }
+	lib_printf("address = %p, *address = %u, value = %u, timeout = %d\n", address, *address, value, timeout);
+	if (*address != value) {
+		proc_lockSet(&process->lock);
+		LIST_REMOVE(&sleepqueue->futex_list, &futex);
+		proc_lockClear(&process->lock);
+		return -EAGAIN;
+	}
 
-    int err = proc_threadSleep(timeout * 1000 * 1000);
-    if (err < 0) {
-        proc_lockSet(&process->lock);
-        LIST_REMOVE(&sleepqueue->futex_list, &futex);
-        proc_lockClear(&process->lock);
-    }
-    return err;
+	int err = proc_threadSleep(timeout * 1000 * 1000);
+	if (err < 0) {
+		proc_lockSet(&process->lock);
+		LIST_REMOVE(&sleepqueue->futex_list, &futex);
+		proc_lockClear(&process->lock);
+	}
+	return err;
 }
 
 int futex_wakeup(unsigned int *address, unsigned int n_threads)
 {
-    lib_printf("address = %p, *address = %u, n_threads = %u\n", address, *address, n_threads);
+	lib_printf("address = %p, *address = %u, n_threads = %u\n", address, *address, n_threads);
 
-    if (n_threads == 0) {
-        return 0;
-    }
-    
-    thread_t *thread = proc_current();
-    process_t *process = thread->process;
+	if (n_threads == 0) {
+		return 0;
+	}
 
-    futex_t futex;
-    hal_memset(&futex, 0, sizeof(futex));
-    
-    addr_t off = (addr_t)address;
-    futex.address = off;
+	thread_t *thread = proc_current();
+	process_t *process = thread->process;
 
-    futex_sleepqueue_t *sleepqueue = futex_getSleepQueue(process, &futex);
+	futex_t futex;
+	hal_memset(&futex, 0, sizeof(futex));
 
-    futex_t *f = sleepqueue->futex_list;
-    int i = 0;
-    while (f != NULL) {
-        if (i == n_threads) {
-            break;
-        }
-        lib_printf("WAKEUP\n");
-        int err = proc_threadWakeup(&f->waiting_thread);
-        if (err < 0) {
-            return err;
-        }
-        f = f->next, i++;
-    }
-    return i;
+	addr_t off = (addr_t)address;
+	futex.address = off;
+
+	futex_sleepqueue_t *sleepqueue = futex_getSleepQueue(process, &futex);
+
+	futex_t *f = sleepqueue->futex_list;
+	int i = 0;
+	while (f != NULL) {
+		if (i == n_threads) {
+			break;
+		}
+		lib_printf("WAKEUP\n");
+		int err = proc_threadWakeup(&f->waiting_thread);
+		if (err < 0) {
+			return err;
+		}
+		f = f->next, i++;
+	}
+	return i;
 }
-
--- a/proc/futex.h
+++ b/proc/futex.h
@@ -5,21 +5,21 @@
 
 /* Implementation inspired by: https://github.com/openbsd/src/blob/master/sys/kern/sys_futex.c */
 
-#define FUTEX_SLEEPQUEUES_BITS  6
-#define FUTEX_SLEEPQUEUES_SIZE  (1U << FUTEX_SLEEPQUEUES_BITS)
-#define FUTEX_SLEEPQUEUES_MASK  (FUTEX_SLEEPQUEUES_SIZE - 1)
+#define FUTEX_SLEEPQUEUES_BITS 6
+#define FUTEX_SLEEPQUEUES_SIZE (1U << FUTEX_SLEEPQUEUES_BITS)
+#define FUTEX_SLEEPQUEUES_MASK (FUTEX_SLEEPQUEUES_SIZE - 1)
 
 struct _thread_t;
 
 typedef struct _futex_t {
-    struct _futex_t *prev, *next;
-    addr_t address;
+	struct _futex_t *prev, *next;
+	addr_t address;
 
-    struct _thread_t *waiting_thread;
+	struct _thread_t *waiting_thread;
 } futex_t;
 
 typedef struct {
-    futex_t *futex_list;
+	futex_t *futex_list;
 } futex_sleepqueue_t;
 
 int futex_wait(unsigned int *address, unsigned int value, time_t timeout);
--- a/proc/futex.h
+++ b/proc/futex.h
@@ -1,0 +1,28 @@
+#ifndef _PROC_FUTEX_H_
+#define _PROC_FUTEX_H_
+
+#include "include/types.h"
+
+/* Implementation inspired by: https://github.com/openbsd/src/blob/master/sys/kern/sys_futex.c */
+
+#define FUTEX_SLEEPQUEUES_BITS  6
+#define FUTEX_SLEEPQUEUES_SIZE  (1U << FUTEX_SLEEPQUEUES_BITS)
+#define FUTEX_SLEEPQUEUES_MASK  (FUTEX_SLEEPQUEUES_SIZE - 1)
+
+struct _thread_t;
+
+typedef struct _futex_t {
+    struct _futex_t *prev, *next;
+    addr_t address;
+
+    struct _thread_t *waiting_thread;
+} futex_t;
+
+typedef struct {
+    futex_t *futex_list;
+} futex_sleepqueue_t;
+
+int futex_wait(unsigned int *address, unsigned int value, time_t timeout);
+int futex_wakeup(unsigned int *address, unsigned int n_threads);
+
+#endif
--- a/proc/process.h
+++ b/proc/process.h
@@ -72,7 +72,7 @@ typedef struct _process_t {
 	void *got;
 	hal_tls_t tls;
 
-    futex_sleepqueue_t futex_sleepqueues[FUTEX_SLEEPQUEUES_SIZE];
+	futex_sleepqueue_t futex_sleepqueues[FUTEX_SLEEPQUEUES_SIZE];
 } process_t;
 
 
--- a/proc/process.h
+++ b/proc/process.h
@@ -23,6 +23,7 @@
 #include "vm/amap.h"
 #include "syspage.h"
 #include "lib/lib.h"
+#include "futex.h"
 
 #define MAX_PID MAX_ID
 
@@ -70,6 +71,8 @@ typedef struct _process_t {
 
 	void *got;
 	hal_tls_t tls;
+
+    futex_sleepqueue_t futex_sleepqueues[FUTEX_SLEEPQUEUES_SIZE];
 } process_t;
 
 
--- a/proc/threads.h
+++ b/proc/threads.h
@@ -28,12 +28,17 @@
 #define THREAD_END_NOW 2
 
 /* Parent thread states */
-enum { PREFORK = 0, FORKING = 1, FORKED };
+enum { PREFORK = 0,
+	FORKING = 1,
+	FORKED };
 
 /* Child thread states */
-enum { OWNSTACK = 0, PARENTSTACK };
+enum { OWNSTACK = 0,
+	PARENTSTACK };
 
-enum { READY = 0, SLEEP, GHOST };
+enum { READY = 0,
+	SLEEP,
+	GHOST };
 
 
 typedef struct _thread_t {
--- a/proc/threads.h
+++ b/proc/threads.h
@@ -165,6 +165,9 @@ extern void proc_zombie(process_t *proc);
 extern int proc_threadSleep(time_t us);
 
 
+extern int proc_threadSleep2(thread_t *thread, time_t us);
+
+
 extern int proc_threadWait(thread_t **queue, spinlock_t *spinlock, time_t timeout, spinlock_ctx_t *scp);
 
 
--- a/proc/userintr.c
+++ b/proc/userintr.c
@@ -31,11 +31,11 @@ void userintr_put(userintr_t *ui)
 	int rem;
 
 	LIB_ASSERT(ui != NULL, "process: %s, pid: %d, tid: %d, ui == NULL",
-		t->process->path, process_getPid(t->process), proc_getTid(t));
+			t->process->path, process_getPid(t->process), proc_getTid(t));
 
 	rem = resource_put(t->process, &ui->resource);
 	LIB_ASSERT(rem >= 0, "process: %s, pid: %d, tid: %d, refcnt below zero",
-		t->process->path, process_getPid(t->process), proc_getTid(t));
+			t->process->path, process_getPid(t->process), proc_getTid(t));
 	if (rem <= 0) {
 		hal_interruptsDeleteHandler(&ui->handler);
 
--- a/syscalls.c
+++ b/syscalls.c
@@ -37,36 +37,36 @@
 
 int syscalls_futex_wait(void *ustack)
 {
-    unsigned int *addr;
-    GETFROMSTACK(ustack, unsigned int *, addr, 0);
-    unsigned int value;
-    GETFROMSTACK(ustack, unsigned int, value, 1);
-    time_t timeout;
-    GETFROMSTACK(ustack, time_t, timeout, 2);
+	unsigned int *addr;
+	GETFROMSTACK(ustack, unsigned int *, addr, 0);
+	unsigned int value;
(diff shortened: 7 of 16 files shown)


proc/futex.c
```diff
@@ -0,0 +1,84 @@
...
+int futex_wait(unsigned int *address, unsigned int value, time_t timeout)
+{
+	thread_t *thread = proc_current();
+	process_t *process = thread->process;
+
+	futex_t futex;
+	hal_memset(&futex, 0, sizeof(futex));
+	futex_sleepqueue_t *sleepqueue = NULL;
+
+	addr_t off = (addr_t)address;
```
ziemleszcz: Local variables should be defined before the function code.

proc/futex.c
```diff
@@ -0,0 +1,84 @@
...
+#include "futex.h"
+
+static inline futex_sleepqueue_t *futex_getSleepQueue(process_t *process, futex_t *f)
+{
+	unsigned int key = f->address >> 3;
+	key ^= key >> FUTEX_SLEEPQUEUES_BITS;
+	return &process->futex_sleepqueues[key & FUTEX_SLEEPQUEUES_MASK];
+}
+
+int futex_wait(unsigned int *address, unsigned int value, time_t timeout)
```
ziemleszcz: Use `u32` for futex instead of `unsigned int`.

proc/futex.c
```diff
@@ -0,0 +1,84 @@
...
+
+	futex_t futex;
+	hal_memset(&futex, 0, sizeof(futex));
+
+	addr_t off = (addr_t)address;
+	futex.address = off;
+
+	futex_sleepqueue_t *sleepqueue = futex_getSleepQueue(process, &futex);
+
+	futex_t *f = sleepqueue->futex_list;
```
ziemleszcz: Unprotected access to the list.

proc/futex.c
```diff
@@ -0,0 +1,84 @@
...
+	int i = 0;
+	while (f != NULL) {
+		if (i == n_threads) {
+			break;
+		}
+		lib_printf("WAKEUP\n");
+		int err = proc_threadWakeup(&f->waiting_thread);
```
ziemleszcz: Incorrect usage of `proc_threadWakeup` - the argument should be a wait queue. Please refer to other usage examples.

proc/threads.h
```diff
@@ -160,6 +160,9 @@ extern void proc_zombie(process_t *proc);
 extern int proc_threadSleep(time_t us);
 
 
+extern int proc_threadSleep2(thread_t *thread, time_t us);
```
ziemleszcz: Unused.

proc/futex.c
```diff
@@ -0,0 +1,74 @@
...
+	key = address >> 3;
+	key ^= key >> FUTEX_SLEEPQUEUES_BITS;
+	futex_sleepqueue = &process->futex_sleepqueues[key & FUTEX_SLEEPQUEUES_MASK];
+	proc_lockClear(&process->lock);
+	return futex_sleepqueue;
+}
+
+int proc_futexWait(u32 *address, u32 value, time_t timeout)
+{
+	thread_t *current_thread = proc_current();
```
ziemleszcz: Prefer lazy initialization, if the function returns EAGAIN  there is no need to assign `current_thread`. In fact, `current_thread` and `current_process` variables can be omitted entirely by directly referencing `proc_current()->process`.

proc/futex.c
```diff
@@ -0,0 +1,74 @@
...
+	process_t *current_process = current_thread->process;
+	addr_t key = (addr_t)address;
+	spinlock_ctx_t ctx;
+	futex_sleepqueue_t *sleepqueue;
+	int err;
+	time_t now;
+
+	if (*address != value) {
```
ziemleszcz: Another thread may modify futex value between lines 31 and 38. For now, it can be left as-is to observe the effects of incorrect futex synchronization during testing.

proc/futex.c
```diff
@@ -0,0 +1,74 @@
...
+		return -EAGAIN;
+	}
+	sleepqueue = futex_getSleepQueue(current_process, key);
+
+	proc_gettime(&now, NULL);
+
+	hal_spinlockSet(&sleepqueue->spinlock, &ctx);
+	sleepqueue->count += 1;
+	err = proc_threadWaitInterruptible(&sleepqueue->threads, &sleepqueue->spinlock, now + timeout, &ctx);
```
ziemleszcz: The infinite timeout case (timeout value of 0) is not handled correctly.

proc/futex.c
```diff
@@ -0,0 +1,74 @@
...
+		return 0;
+	}
+
+	sleepqueue = futex_getSleepQueue(current_process, key);
+	hal_spinlockSet(&sleepqueue->spinlock, &ctx);
+	if (n_threads == FUTEX_WAKEUP_ALL) {
+		n_threads = sleepqueue->count;
+	}
+
+	for (int i = 0; i < n_threads && sleepqueue->count > 0; i++, sleepqueue->count -= 1) {
```
ziemleszcz: Move definition of `i` to the top.

proc/futex.c
```diff
@@ -0,0 +1,74 @@
...
+		err = proc_threadWakeup(&sleepqueue->threads);
```
ziemleszcz: Please verify the return value semantics, it returns either 0 or 1.

proc/futex.c
```diff
@@ -0,0 +1,74 @@
...
+		if (err < 0) {
+			hal_spinlockClear(&sleepqueue->spinlock, &ctx);
+			return err;
+		}
+	}
+	hal_spinlockClear(&sleepqueue->spinlock, &ctx);
+	return n_threads;
```
ziemleszcz: It should return number of threads that were woken up (please see `proc_threadWakeup` return value).

proc/futex.h
```diff
@@ -0,0 +1,22 @@
...
+
+/* Implementation inspired by: https://github.com/openbsd/src/blob/master/sys/kern/sys_futex.c */
+
+#define FUTEX_SLEEPQUEUES_BITS 6
+#define FUTEX_SLEEPQUEUES_SIZE (1U << FUTEX_SLEEPQUEUES_BITS)
+#define FUTEX_SLEEPQUEUES_MASK (FUTEX_SLEEPQUEUES_SIZE - 1)
+#define FUTEX_WAKEUP_ALL       ((u32) - 1)
+
+typedef struct {
+	struct _thread_t *threads;
```
ziemleszcz, line 38: You can use `thread_t` type.

proc/futex.c
```diff
@@ -0,0 +1,74 @@
...
+	addr_t key = (addr_t)address;
+	int err;
+	spinlock_ctx_t ctx;
+	futex_sleepqueue_t *sleepqueue;
+
+	if (n_threads == 0) {
+		return 0;
+	}
+
+	sleepqueue = futex_getSleepQueue(current_process, key);
```
ziemleszcz: Handling of futex `key` collisions is not implemented yet. Please add a TODO comment about this.

proc/futex.c
```diff
@@ -0,0 +1,78 @@
...
+	if (*address != value) {
+		return -EAGAIN;
+	}
+	current_process = proc_current()->process;
+	key = (addr_t)address;
+	sleepqueue = futex_getSleepQueue(current_process, key);
+
+	if (timeout > 0) {
+		proc_gettime(&now, NULL);
+		total_timeout = now + timeout;
```
ziemleszcz: `total_timeout` variable is redundant, you can add `now` directly to `timeout` instead.

proc/futex.c
```diff
@@ -0,0 +1,78 @@
...
+	addr_t key;
+	spinlock_ctx_t ctx;
+	futex_sleepqueue_t *sleepqueue;
+	int err;
+	time_t now, total_timeout = 0;
+
+	if (*address != value) {
+		return -EAGAIN;
+	}
+	current_process = proc_current()->process;
```
ziemleszcz: There is a single reference to `current_process` variable, it can be replaced with `proc_current()->process`, allowing the removal of this variable.

proc/futex.c
```diff
@@ -0,0 +1,78 @@
...
+	key = (addr_t)address;
+	sleepqueue = futex_getSleepQueue(current_process, key);
+
+	if (timeout > 0) {
+		proc_gettime(&now, NULL);
+		total_timeout = now + timeout;
+	}
+
+	hal_spinlockSet(&sleepqueue->spinlock, &ctx);
+	sleepqueue->count += 1;
```
ziemleszcz: Is this counter really needed?

proc/futex.c
```diff
@@ -0,0 +1,78 @@
...
+
+	if (n_threads == 0) {
+		return 0;
+	}
+
+	current_process = proc_current()->process;
+	key = (addr_t)address;
+	sleepqueue = futex_getSleepQueue(current_process, key);
+	hal_spinlockSet(&sleepqueue->spinlock, &ctx);
+	if (n_threads == FUTEX_WAKEUP_ALL) {
```
ziemleszcz: Please check out `_proc_threadBroadcast` function (it wakes up all threads waiting for cond).

proc/futex.c
```diff
@@ -0,0 +1,148 @@
+#include <stdatomic.h>
```
ziemleszcz, line 15: Please add a copyright header to each new file.

proc/futex.c
```diff
@@ -0,0 +1,148 @@
...
+	proc_lockSet(&process->lock);
+	key = address >> 3;
+	key ^= key >> FUTEX_SLEEPQUEUES_BITS;
+	idx = key & FUTEX_SLEEPQUEUES_MASK;
+
+	/* Find a taken slot with the same address using linear probing */
+	i = idx;
+	while (1) {
+		if (process->futex_sleepqueues[i].address == address) {
+			futex_sleepqueue = &process->futex_sleepqueues[i];
```
ziemleszcz: Why not break here, you have already found the address?

proc/futex.c
```diff
@@ -0,0 +1,148 @@
...
+		}
+		i = (i + 1) % FUTEX_SLEEPQUEUES_SIZE;
+		if (i == idx) {
```
ziemleszcz: You can stop searching on first NULL address.

proc/futex.c
```diff
@@ -0,0 +1,148 @@
+#include <stdatomic.h>
+#include "hal/hal.h"
+#include "lib/lib.h"
+#include "include/errno.h"
+#include "process.h"
+#include "threads.h"
+#include "futex.h"
+
+static futex_sleepqueue_t *futex_getSleepQueue(process_t *process, addr_t address, u32 allocate)
```
ziemleszcz: It's probably better to split this into separate functions for searching and allocating.

proc/futex.c
```diff
@@ -0,0 +1,148 @@
...
+int proc_futexWait(_Atomic(u32) *address, u32 value, time_t timeout)
+{
+	spinlock_ctx_t ctx;
+	futex_sleepqueue_t *sleepqueue;
+	futex_waitctx_t wc;
+	thread_t *current;
+	int err = EOK;
+	time_t total_timeout = 0;
+
+	if (timeout > 0) {
```
ziemleszcz: Could you reorder the code? We might return early without ever using `total_timeout`.

proc/futex.c
```diff
@@ -0,0 +1,148 @@
...
+		}
+	}
+
+	proc_lockClear(&process->lock);
+	return futex_sleepqueue;
+}
+
+static int proc_futexUnwait(futex_sleepqueue_t *sq, futex_waitctx_t *wc)
+{
+	int r;
```
ziemleszcz: Fix the type.

proc/futex.c
```diff
@@ -0,0 +1,148 @@
...
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&sq->spinlock, &sc);
+	r = atomic_load(&wc->thread) != NULL;
+	if (r) {
```
ziemleszcz, line 86: r != NULL

proc/futex.c
```diff
@@ -0,0 +1,148 @@
...
+
+	if (n_threads == 0) {
+		return 0;
+	}
+
+	current = proc_current();
+
+	sleepqueue = futex_getSleepQueue(current->process, (addr_t)address, 0);
+	if (sleepqueue == NULL) {
+		return 0;
```
ziemleszcz, line 185: It should return an error.

proc/futex.c
```diff
@@ -0,0 +1,178 @@
+/*
+ * phoenix-rtos
```
ziemleszcz: No need to lowercase it, just copy it exactly from the original source file.

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+#include "process.h"
+#include "threads.h"
+#include "futex.h"
+
+static futex_sleepqueue_t *proc_allocFutexSleepQueue(process_t *process, addr_t address)
+{
+	u32 i;
+	futex_sleepqueue_t *futex_sleepqueue = NULL;
+
+	/* Not found. Look for an free slot */
```
ziemleszcz: Please update the comment to match the updated code, and fix the grammar (`an` -> `a`).

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+#include <stdatomic.h>
+#include <stdbool.h>
+#include "hal/hal.h"
+#include "lib/lib.h"
+#include "include/errno.h"
+#include "process.h"
+#include "threads.h"
+#include "futex.h"
+
+static futex_sleepqueue_t *proc_allocFutexSleepQueue(process_t *process, addr_t address)
```
ziemleszcz: Functions that aren’t protected should start with an underscore.

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+{
+	u32 i;
+	futex_sleepqueue_t *futex_sleepqueue = NULL;
```
ziemleszcz: Rename it to `sleepqueue` or just `queue`. It's clear from the context that it's futex sleep queue.

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+
+	/* Not found. Look for an free slot */
+	for (i = 0; i < FUTEX_SLEEPQUEUES_SIZE; i++) {
```
ziemleszcz: With linear probing, you should start from the key's index, not from 0.

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+
+	key = address >> 3;
+	key ^= key >> FUTEX_SLEEPQUEUES_BITS;
+	idx = key & FUTEX_SLEEPQUEUES_MASK;
+
+	/* Find a taken slot with the same address using linear probing */
+	i = idx;
+	while (1) {
+		if (process->futex_sleepqueues[i].address == address) {
+			futex_sleepqueue = &process->futex_sleepqueues[i];
```
ziemleszcz: Just return the sleep queue pointer here, and NULL at the end. No need for an extra variable.

proc/futex.c
ziemleszcz: This comment is still not addressed. With linear probing, you can stop as soon as you hit the first NULL slot.

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+	u32 key, idx, i;
+	futex_sleepqueue_t *futex_sleepqueue = NULL;
+
+	key = address >> 3;
+	key ^= key >> FUTEX_SLEEPQUEUES_BITS;
+	idx = key & FUTEX_SLEEPQUEUES_MASK;
+
+	/* Find a taken slot with the same address using linear probing */
+	i = idx;
+	while (1) {
```
ziemleszcz: This version is easier to read:
do {
} while (i != idx);

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+
+	current = proc_current();
+
+	proc_lockSet(&current->process->lock);
+	sleepqueue = proc_getFutexSleepQueue(current->process, (addr_t)address);
+	if (sleepqueue == NULL) {
+		sleepqueue = proc_allocFutexSleepQueue(current->process, (addr_t)address);
+	}
+	if (sleepqueue == NULL) {
+		proc_lockClear(&current->process->lock);
```
ziemleszcz: You can clear the lock before `if` block.

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+	if (n_threads == 0) {
+		return 0;
+	}
+
+	current = proc_current();
+
+	proc_lockSet(&current->process->lock);
+	sleepqueue = proc_getFutexSleepQueue(current->process, (addr_t)address);
+	if (sleepqueue == NULL) {
+		proc_lockClear(&current->process->lock);
```
ziemleszcz: You can clear the lock before if block.

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+	thread_t *current;
+	int err = EOK;
+	time_t total_timeout = 0;
+
+	current = proc_current();
+
+	proc_lockSet(&current->process->lock);
+	sleepqueue = proc_getFutexSleepQueue(current->process, (addr_t)address);
+	if (sleepqueue == NULL) {
+		sleepqueue = proc_allocFutexSleepQueue(current->process, (addr_t)address);
```
ziemleszcz: It seems sleep queues are never freed. What happens if a futex based mutex gets destroyed?

proc/futex.c
```diff
@@ -0,0 +1,178 @@
...
+	if (sleepqueue == NULL) {
+		proc_lockClear(&current->process->lock);
+		return 0;
+	}
+	proc_lockClear(&current->process->lock);
+
+	hal_spinlockSet(&sleepqueue->spinlock, &ctx);
+
+	while ((wc = sleepqueue->waitctxs) != NULL) {
+		LIST_REMOVE(&sleepqueue->waitctxs, wc);
```
ziemleszcz, line 164: You're assuming `wc->thread` is the same as the thread being woken up. Why is that?

proc/futex.c
```diff
@@ -0,0 +1,175 @@
...
+		if (process->futex_sleepqueues[i].address == 0) {
+			process->futex_sleepqueues[i].address = address;
+			return &process->futex_sleepqueues[i];
+		}
+		i = (i + 1) % FUTEX_SLEEPQUEUES_SIZE;
+	} while (i != idx);
+
+	return NULL;
+}
+
```
ziemleszcz, line 55: Please follow our coding conventions: `Functions should be separated with two newline characters.`

proc/futex.c
```diff
@@ -0,0 +1,175 @@
...
+	if (r) {
+		LIST_REMOVE(&sq->waitctxs, wc);
+	}
+	hal_spinlockClear(&sq->spinlock, &sc);
+	return r;
+}
+
+int proc_futexWait(_Atomic(u32) *address, u32 value, time_t timeout)
+{
+	spinlock_ctx_t ctx, sqsl_ctx;
```
ziemleszcz: Try to keep variable names consistent and adhere to our coding convention (camelCase). Use `sc` for spinlock context, and since you’re using `sq` for sleepqueue in `proc_futexUnwait`, that works well too. For example:


spinlock_ctx_t sc, sqSc;
futex_sleepqueue_t *sq;


Stick to these names across the code to keep things clear and uniform.

proc/futex.c
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return err;
+done:
+	if (atomic_load(&wc.thread) != NULL) {
+		proc_futexUnwait(sleepqueue, &wc);
+	}
+	return err;
+}
+
+int proc_futexWakeup(process_t *process, _Atomic(u32) *address, u32 n_threads)
```
ziemleszcz: `n_threads` violates our coding convention. Please change, for example to `cnt` or `wakeCnt`.

proc/futex.c
```diff
@@ -0,0 +1,175 @@
...
+	hal_spinlockSet(&process->futex_sq_spinlock, &sqsl_ctx);
+	sleepqueue = _proc_getFutexSleepQueue(process, (addr_t)address);
+	hal_spinlockClear(&process->futex_sq_spinlock, &sqsl_ctx);
+	if (sleepqueue == NULL) {
+		return 0;
+	}
+
+	hal_spinlockSet(&sleepqueue->spinlock, &ctx);
+
+	while ((wc = sleepqueue->waitctxs) != NULL) {
```
ziemleszcz: You're holding the spinlock for too long here. This block should just move the threads to a separate list. The actual wake-up should happen after releasing the spinlock, like how it's done in OpenBSD.

proc/futex.c
```diff
@@ -0,0 +1,175 @@
...
+		LIST_REMOVE(&sleepqueue->waitctxs, wc);
+		tmp = atomic_load(&wc->thread);
+		atomic_store(&wc->thread, NULL);
+		if (proc_threadWakeupOne(tmp)) {
```
ziemleszcz: What if this thread was killed while it was still waiting? Are we handling that case properly here?

proc/process.h
```diff
@@ -70,6 +71,9 @@ typedef struct _process_t {
 
 	void *got;
 	hal_tls_t tls;
+
+	spinlock_t futex_sq_spinlock;
```
ziemleszcz: Use camelCase.

proc/userintr.c
```diff
@@ -13,12 +13,13 @@
...
  */
 
+#include <stdatomic.h>
 #include "lib/lib.h"
 #include "resource.h"
 #include "userintr.h"
 #include "cond.h"
 #include "proc.h"
-
+#include "futex.h"
```
ziemleszcz, line 22: Keep two blank lines between code blocks or function definitions.

syscalls.c
```diff
@@ -35,6 +35,40 @@
  */
 
 
+int syscalls_phFutexWait(void *ustack)
+{
+	_Atomic(u32) *addr;
```
ziemleszcz, line 41: Local variables should be defined before the function code.

syscalls.c
```diff
@@ -35,6 +35,40 @@
...
+	if (vm_mapBelongs(process, addr, sizeof(*addr)) < 0) {
+		return -EFAULT;
+	}
+
+	return proc_futexWait(addr, value, timeout);
+}
+
+int syscalls_phFutexWakeup(void *ustack)
+{
+	_Atomic(u32) *addr;
```
ziemleszcz, line 63: Local variables should be defined before the function code.

proc/userintr.h
```diff
@@ -27,7 +28,7 @@ typedef struct _userintr_t {
 	process_t *process;
 	int (*f)(unsigned int, void *);
 	void *arg;
-	cond_t *cond;
+	_Atomic(u32) *cond_futex;
```
ziemleszcz: Use camelCase.

proc/userintr.c
```diff
@@ -81,21 +79,17 @@ int userintr_setHandler(unsigned int n, int (*f)(unsigned int, void *), void *ar
...
 	process_t *process = proc_current()->process;
 	userintr_t *ui;
-	cond_t *cond = NULL;
 	int id, res;
+	_Atomic(u32) *cond_futex = (_Atomic(u32) *)(addr_t)c;
 
 	if (c > 0) {
-		cond = cond_get(c);
-		if (cond == NULL) {
+		if (vm_mapBelongs(process, cond_futex, sizeof(*cond_futex)) < 0) {
```
ziemleszcz: This check, along with `handle_t` conversion, should be moved into `syscall_interrupt`.

proc/futex.c
```diff
@@ -0,0 +1,190 @@
...
+	hal_spinlockClear(&process->futex_sq_spinlock, &sqSc);
+	if (sq == NULL) {
+		return 0;
+	}
+
+	hal_spinlockSet(&sq->spinlock, &sc);
+
+	while ((wc = sq->waitctxs) != NULL) {
+		LIST_REMOVE(&sq->waitctxs, wc);
+		tmpWc = vm_kmalloc(sizeof(*tmpWc));
```
ziemleszcz: Why are you allocating a new structure instead of just reusing and relinking the existing one?

proc/futex.c
```diff
@@ -0,0 +1,190 @@
...
+		atomic_store(&wc->thread, NULL);
+
+		i++;
+		if (i == wakeCount && wakeCount != FUTEX_WAKEUP_ALL) {
+			break;
+		}
+	}
+	hal_spinlockClear(&sq->spinlock, &sc);
+
+	while ((wc = wakeupList) != NULL) {
```
ziemleszcz: No need to remove items from the list. You just need to loop through them.

proc/futex.h
```diff
@@ -0,0 +1,44 @@
...
+	spinlock_t spinlock;
+	addr_t address;
+	futex_waitctx_t *waitctxs;
+} futex_sleepqueue_t;
+
+int proc_futexWait(_Atomic(u32) *address, u32 value, time_t timeout, int clockType);
+int proc_futexWakeup(struct _process_t *process, _Atomic(u32) *address, u32 n_threads);
```
ziemleszcz: n_threads -> wakeCount

syscalls.c
```diff
@@ -658,27 +571,31 @@ int syscalls_resourceDestroy(void *ustack)
...
+	GETFROMSTACK(ustack, _Atomic(u32) *, condFutex, 3);
 	GETFROMSTACK(ustack, handle_t *, handle, 4);
 
 	if ((handle != NULL) && (vm_mapBelongs(proc, handle, sizeof(*handle)) < 0)) {
 		return -EFAULT;
 	}
 
-	res = userintr_setHandler(n, f, data, cond);
+	if (condFutex != NULL && vm_mapBelongs(proc, condFutex, sizeof(*condFutex)) < 0) {
+		return -EINVAL;
```
ziemleszcz: EFAULT

proc/futex.c
```diff
@@ -0,0 +1,211 @@
...
+	sq = _proc_getFutexSleepQueue(current->process, (addr_t)address);
+	if (sq == NULL) {
+		sq = _proc_allocFutexSleepQueue(current->process, (addr_t)address);
+	}
+	hal_spinlockClear(&current->process->futexSqSpinlock, &sqSc);
+	if (sq == NULL) {
+		return -ENOMEM;
+	}
+
+	atomic_store(&wc.thread, current);
```
ziemleszcz, line 142: `atomic_init` is missing. `atomic_store` without it is UB.

proc/futex.c
```diff
@@ -0,0 +1,211 @@
...
+		LIST_ADD(&wakeupList, wc);
+		i++;
+		if (i == wakeCount && wakeCount != FUTEX_WAKEUP_ALL) {
+			break;
+		}
+	}
+	hal_spinlockClear(&sq->spinlock, &sc);
+
+	if (wakeupList != NULL) {
+		tmpwc = wakeupList;
```
ziemleszcz: Why not reuse `wc`?

proc/futex.c
```diff
@@ -0,0 +1,211 @@
...
+		do {
+			tmp = atomic_load(&tmpwc->thread);
+			tmpwc = tmpwc->next;
+			atomic_store(&tmpwc->thread, NULL);
```
ziemleszcz, line 206: Big fail, checkout Gemini comment.

proc/futex.c
```diff
@@ -0,0 +1,214 @@
...
+	hal_spinlockClear(&sq->spinlock, &sc);
+
+	if (wakeupList != NULL) {
+		wc = wakeupList;
+		do {
+			/* TODO: Figure out why atomic_exchange doesn't work here */
+			/* tmp = atomic_exchange(&tmpwc->thread, NULL); */
+			tmp = atomic_load(&wc->thread);
+			atomic_store(&wc->thread, NULL);
+			wc = wc->next;
```
ziemleszcz, line 205: It's not safe to use `wc` after setting `wc->thread` to NULL. You should take next element pointer before `atomic_store()`.

include/syscalls.h
```diff
@@ -124,5 +116,8 @@
 	ID(sys_mprotect) \
 	\
 	ID(sys_statvfs) \
-	ID(sys_uname)
+	ID(sys_uname) \
+	\
+	ID(phFutexWait) \
+	ID(phFutexWakeup)
```
Darchiv, line 122: FYI: You remove syscalls, so it doesn't matter if you add new ones in a backward compatible manner (as last) or not, because it is already a breaking change which requires a rebuild. `phFutexWait` etc. can be placed by similar primitive syscalls, e.g. `phInterrupt`.

proc/futex.h
```diff
@@ -0,0 +1,44 @@
...
+#ifndef _PROC_FUTEX_H_
+#define _PROC_FUTEX_H_
+
+#include "include/types.h"
+
+/* Implementation inspired by: https://github.com/openbsd/src/blob/master/sys/kern/sys_futex.c */
+
+#define FUTEX_SLEEPQUEUES_BITS 6
+#define FUTEX_SLEEPQUEUES_SIZE (1U << FUTEX_SLEEPQUEUES_BITS)
+#define FUTEX_SLEEPQUEUES_MASK (FUTEX_SLEEPQUEUES_SIZE - 1)
```
Darchiv, line 27: Sleep queue implementation detail. Should be hidden in futex.c.

proc/futex.c
```diff
@@ -0,0 +1,214 @@
...
+#include "futex.h"
+
+
+static futex_sleepqueue_t *_proc_allocFutexSleepQueue(process_t *process, addr_t address)
+{
+	u32 key, idx, i;
+
+	key = address >> 3;
+	key ^= key >> FUTEX_SLEEPQUEUES_BITS;
+	idx = key & FUTEX_SLEEPQUEUES_MASK;
```
Darchiv: Not obvious what this bit-mangling does. Add a comment which describes that this is a simple hash-table implementation or something.

Or make generic macros which will contain these operations (and deduplicate with `_proc_getFutexSleepQueue()`), with a self-documenting name (as we all love self-documenting code) - this way a comment will not be necessary. Yes, the implementation should be comprehensible by just looking at the code.

proc/futex.h
```diff
@@ -0,0 +1,44 @@
...
+typedef struct {
+	struct _thread_t *threads;
+	spinlock_t spinlock;
+	addr_t address;
+	futex_waitctx_t *waitctxs;
+} futex_sleepqueue_t;
+
+int proc_futexWait(_Atomic(u32) *address, u32 value, time_t timeout, int clockType);
+int proc_futexWakeup(struct _process_t *process, _Atomic(u32) *address, u32 wakeCount);
+futex_sleepqueue_t *_proc_getFutexSleepQueue(struct _process_t *process, addr_t address);
```
Darchiv: Why export if not used outside of futex.c?

proc/futex.c
```diff
@@ -0,0 +1,214 @@
...
+int proc_futexWait(_Atomic(u32) *address, u32 value, time_t timeout, int clockType)
+{
+	spinlock_ctx_t sc, sqSc;
+	futex_sleepqueue_t *sq;
+	futex_waitctx_t wc;
+	thread_t *current;
+	int err = EOK;
+	time_t waitTime = 0, offs;
+
+	if (clockType != PH_CLOCK_REALTIME && clockType != PH_CLOCK_RELATIVE && clockType != PH_CLOCK_MONOTONIC) {
```
Darchiv: This can be replaced by just adding a `default` case for the switch-case below.

`clockType` is used only when  `timeout != 0`, so it makes sense to only check it in this case (performance + code size). Also, if not using `timeout`, the user should be able to use this function without looking up valid values for `clockType`, i.e. just pass 0 or -1. If `PH_CLOCK_*` values were completely arbitrary and did not include 0, then the user would be forced to use a specific `PH_CLOCK_*` value even if not using `timeout` or an additional value would have to be added (e.g. `PH_CLOCK_NONE`).

proc/futex.c
```diff
@@ -0,0 +1,214 @@
...
+				break;
+		}
+	}
+
+	current = proc_current();
+
+	hal_spinlockSet(&current->process->futexSqSpinlock, &sqSc);
+	sq = _proc_getFutexSleepQueue(current->process, (addr_t)address);
+	if (sq == NULL) {
+		sq = _proc_allocFutexSleepQueue(current->process, (addr_t)address);
```
Darchiv, line 134: So, basically, once we allocate a sleep queue, we never free it.

I assume that you wanted to mimic futex syscall from other OSes, but can't we actually have a `futexCreate` syscall which will allocate a sleep queue (kernel side of futex, not including the userspace context) on idtree and release it via `resourcesDestroy()`?

proc/futex.h
kamilaboba123: I'd keep those definitions public, because they need to be visible to proc/process.h.
```
// ...
        void *got;
	hal_tls_t tls;

	spinlock_t futexSqSpinlock;
	futex_sleepqueue_t futexSleepQueues[FUTEX_SLEEPQUEUES_SIZE]; // depends on FUTEX_SLEEPQUEUE_BITS
} process_t;
```

proc/futex.c
```diff
@@ -0,0 +1,221 @@
...
+#include "hal/hal.h"
+#include "lib/lib.h"
+#include "include/errno.h"
+#include "include/time.h"
+#include "process.h"
+#include "threads.h"
+#include "futex.h"
+
+
+u32 _proc_futexTableHash(addr_t address)
```
Darchiv: Not exported, used internally. Should be `static`

proc/futex.c
```diff
@@ -0,0 +1,221 @@
...
+	atomic_init(&wc.thread, NULL);
+
+	atomic_store(&wc.thread, current);
+	hal_spinlockSet(&sq->spinlock, &sc);
+	LIST_ADD(&sq->waitctxs, &wc);
+	hal_spinlockClear(&sq->spinlock, &sc);
+
+	if (atomic_load(address) != value) {
+		err = -EAGAIN;
+		goto done;
```
Darchiv: Please avoid using `goto` (see: coding convention on docs). If you really need to use the non-structured paradigm (or it vastly increases readability), consider using a do-while. For example:
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/e35fd78ecc48b0957fedb99950f1f7f214cb00e9/posix/unix.c#L407-L463

But it seems an overkill for just one branch (goto).

proc/futex.h
```diff
@@ -0,0 +1,43 @@
...
+
+typedef struct {
+	struct _thread_t *threads;
+	spinlock_t spinlock;
+	addr_t address;
+	futex_waitctx_t *waitctxs;
+} futex_sleepqueue_t;
+
+int proc_futexWait(_Atomic(u32) *address, u32 value, time_t timeout, int clockType);
+int proc_futexWakeup(struct _process_t *process, _Atomic(u32) *address, u32 wakeCount);
```
Darchiv, line 48: Type and function declarations should be separated by two empty lines. See other headers for example.

PR #670, opened 2025-07-14
!syscalls: move signal handling to kernel

### LINKED ISSUE #1367: libphoenix: calling signal or sigaction causes temporary unmasking of other signals
Current implementation of signal() and sigaction() unmasks all signals during masking modified one.
Code example:
```C
int main(void) {
	sigset_t s;
	sigemptyset(&s);
	sigaddset(&s, SIGUSR1);
	sigprocmask(SIG_SETMASK, &s, NULL);
	raise(SIGUSR1);

	signal(SIGUSR2, SIG_IGN);
	return 0;
}
```
Behavior: process exits with status 158 (SIGUSR1 termination)
Expected: process exits with status 0

### LINKED ISSUE #1432: `wait()` can be interrupted even when signal is ignored
quote from `POSIX` [wait()](https://pubs.opengroup.org/onlinepubs/009696899/functions/wait.html) description:
_The wait() function shall suspend execution of the calling thread until status information for one of the terminated child processes of the calling process is available, or **until delivery of a signal whose action is either to execute a signal-catching function or to terminate the process**._ 

Reproduction code (ia32-generic-qemu):
```c
#include <stdio.h>
#include <unistd.h>
#include <signal.h>
#include <stdlib.h>
#include <errno.h>
#include <sys/wait.h>

int main(void)
{
        struct sigaction sa;
        sa.sa_handler = SIG_IGN;
        sa.sa_flags = 0;
        sigemptyset(&sa.sa_mask);
        sigaction(SIGALRM, &sa, NULL);

        if (fork() == 0) {
                sleep(5);
                exit(0);
        }

        alarm(1);
        int ret = wait(NULL);

        printf("ret: %d errno: %d\n", ret, errno);

        return 0;
}
```

Image

### LINKED ISSUE #1448: `exit()`: `WIFEXITED` returns false and `WIFSIGNALED` true for exit codes above 255
Quote from `POSIX` [`exit()`](https://pubs.opengroup.org/onlinepubs/009695299/functions/exit.html) description:

_The value of status may be 0, EXIT_SUCCESS, EXIT_FAILURE, or any other value, **though only the least significant 8 bits (that is, status & 0377) shall be available to a waiting parent process**._

Reproduction (ia32-generic-qemu):
```c
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main(void)
{
        if (fork() == 0) {
                sleep(2);
                exit(0xFF);
        }

        if (fork() == 0) {
                sleep(5);
                exit(0xFF + 1);
        }

        int status;

        for (int i = 0; i < 2; i++) {
                wait(&status);

                if (WIFEXITED(status)) {
                        printf("exited correctly\n");
                }
                else if (WIFSIGNALED(status)) {
                        printf("wrong exit without signal\n");
                }
                else {
                        printf("error\n");
                }
        }

        return 0;
}
```

Image

### LINKED ISSUE #1453: `kill()` doesn't fail on invalid negative signal number
Reproduction (ia32-generic-qemu):
```c
#include <stdio.h>
#include <errno.h>
#include <unistd.h>
#include <signal.h>

int main(void)
{
        int ret, dummy = -1;

        errno = 0;
        ret = kill(getpid(), dummy);
        printf("kill ret: %d errno: %d\n", ret, errno);

        dummy = 123;
        errno = 0;
        ret = kill(getpid(), dummy);
        printf("kill ret: %d errno: %d\n", ret, errno);

        return 0;
}
```

Image

### LINKED ISSUE #1456: child process signal mask is not inherited from the parent
Reproduction (ia32-generic-qemu):
```c
#include <stdio.h>
#include <unistd.h>
#include <stdlib.h>
#include <signal.h>
#include <sys/wait.h>

int main(void)
{
        sigset_t set;
        sigset_t oldset;
        int ret;

        sigemptyset(&set);
        sigaddset(&set, SIGHUP);
        sigaddset(&set, SIGTERM);
        sigaddset(&set, SIGINT);
        sigaddset(&set, SIGTTIN);
        sigaddset(&set, SIGTTOU);

        ret = sigprocmask(SIG_SETMASK, &set, NULL);
        if (ret != 0) {
                perror("sigprocmask");
                return 1;
        }

        if (fork() == 0) {
                ret = sigprocmask(SIG_SETMASK, NULL, &oldset);
                if (ret != 0) {
                        perror("sigprocmask");
                        return 1;
                }

                printf("SIGHUP: %d\n", sigismember(&oldset, SIGHUP));
                printf("SITERM: %d\n", sigismember(&oldset, SIGTERM));
                printf("SIINT: %d\n", sigismember(&oldset, SIGINT));
                printf("SIGTTIN: %d\n", sigismember(&oldset, SIGTTIN));
                printf("SIGTTOU: %d\n", sigismember(&oldset, SIGTTOU));

                exit(0);
        }

        wait(NULL);

        return 0;
}
```

Image

host:

Image

### LINKED ISSUE #1464: `exec()` doesn't keep ignored signals from parent process
Quote from `POSIX` [`exec()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html) description:

_Signals set to the default action (SIG_DFL) in the calling process image shall be set to the default action in the new process image. **Except for SIGCHLD, signals set to be ignored (SIG_IGN) by the calling process image shall be set to be ignored by the new process image.** Signals set to be caught by the calling process image shall be set to the default action in the new process image_

etiaro:
## Description

* keep actions for every signal and signal trampoline in process_t struct
  * allocate actions lazily on first change
* remove unused sigmask from process_t struct, rely on thread_t sigmask field
* add handler address as signal_trampoline stack parameter
* use POSIX signal codes and default behaviors
* replaced signalHandle syscall with signalAction
* handled vforked process signal termination
* remove signalPost syscall as duplicate to sys_tkill

## Motivation and Context

Moving singal handlers to kernel enables it to know whether coredump should be run, by knowing whether a signal handler is installed for given signal.
Simplifies handling exceptions with problems like stackoverflow (JIRA: RTOS-1066, currently stackoverflow causes kernel exception during sigsegv handling even with default sigsegv action).
Allows improved POSIX compliability (eg. not interrupting sleep by ignored signals).

Resolves phoenix-rtos/phoenix-rtos-project#1367
Resolves phoenix-rtos/phoenix-rtos-project#1456
Resolves phoenix-rtos/phoenix-rtos-project#1464
Resolves phoenix-rtos/phoenix-rtos-project#1453
Resolves phoenix-rtos/phoenix-rtos-project#1432
Resolves phoenix-rtos/phoenix-rtos-project#1448
JIRA: RTOS-736

Depends-On: phoenix-rtos-tests:etiaro/signals

## Types of changes

## How Has This Been Tested?

- [x] New test added: phoenix-rtos/phoenix-rtos-tests#410
- [x] Tested by hand on:
- qemu with ia32, aarch64 and riscv, with SMP between 1 and 10 cores
- nightly tests
- additional test crafted for detecting smp/race issues ([can be found here](https://github.com/phoenix-rtos/phoenix-rtos-tests/blob/221c4738a3bdf490238cb8bf5ef473b19938ec86/libc/signal/bulk.c))

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/libphoenix/pull/420

### DIFF
--- a/hal/armv7m/cpu.c
+++ b/hal/armv7m/cpu.c
@@ -155,7 +155,7 @@ int hal_cpuCreateContext(cpu_context_t **nctx, startFn_t start, void *kstack, si
 }
 
 
-int hal_cpuPushSignal(void *kstack, void (*handler)(void), cpu_context_t *signalCtx, int n, unsigned int oldmask, const int src)
+int hal_cpuPushSignal(void *kstack, void (*trampoline)(void), void (*handler)(int signo), cpu_context_t *signalCtx, int n, unsigned int oldmask, const int src)
 {
 	cpu_context_t *ctx = (void *)((char *)kstack - sizeof(cpu_context_t));
 	struct stackArg args[] = {
@@ -164,6 +164,7 @@ int hal_cpuPushSignal(void *kstack, void (*handler)(void), cpu_context_t *signal
 		{ &ctx->hwctx.pc, sizeof(ctx->hwctx.pc) },
 		{ &signalCtx, sizeof(signalCtx) },
 		{ &oldmask, sizeof(oldmask) },
+		{ &handler, sizeof(handler) },
 		{ &n, sizeof(n) },
 		{ 0, 0 } /* Reserve space for optional HWCTX */
 	};
@@ -174,7 +175,7 @@ int hal_cpuPushSignal(void *kstack, void (*handler)(void), cpu_context_t *signal
 	signalCtx->psp -= sizeof(cpu_context_t);
 
 	/* parasoft-suppress-next-line MISRAC2012-RULE_11_1 "Need to assign function address to processor register" */
-	signalCtx->hwctx.pc = (u32)handler;
+	signalCtx->hwctx.pc = (u32)trampoline;
 
 	/* Set default PSR, clear potential ICI/IT flags */
 	signalCtx->hwctx.psr = 0x01000000U;
--- a/include/signal.h
+++ b/include/signal.h
@@ -17,8 +17,108 @@
 #ifndef _PH_SIGNAL_H_
 #define _PH_SIGNAL_H_
 
-enum { signal_kill = 1, signal_segv, signal_illegal, signal_cancel = 32 };
+#include "types.h"
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+typedef void (*sighandler_t)(int signo);
+
+
+#define SIGNULL      0
+#define SIGHUP       1
+#define SIGINT       2
+#define SIGQUIT      3
+#define SIGILL       4
+#define SIGTRAP      5
+#define SIGABRT      6
+#define SIGIOT       SIGABRT
+#define SIGEMT       7
+#define SIGFPE       8
+#define SIGKILL      9
+#define SIGBUS       10
+#define SIGSEGV      11
+#define SIGSYS       12
+#define SIGPIPE      13
+#define SIGALRM      14
+#define SIGTERM      15
+#define SIGURG       16
+#define SIGSTOP      17
+#define SIGTSTP      18
+#define SIGCONT      19
+#define SIGCHLD      20
+#define SIGTTIN      21
+#define SIGTTOU      22
+#define SIGIO        23
+#define SIGXCPU      24
+#define SIGXFSZ      25
+#define SIGVTALRM    26
+#define SIGPROF      27
+#define SIGWINCH     28
+#define SIGINFO      29
+#define SIGUSR1      30
+#define SIGUSR2      31
+#define PH_SIGCANCEL 32
 
 #define NSIG 32
+
+/* clang-format off */
+#define SIG_DFL ((sighandler_t)0)
+#define SIG_IGN ((sighandler_t)-1)
+#define SIG_ERR ((sighandler_t)-2)
+/* clang-format on */
+
+
+enum { SIG_BLOCK,
+	SIG_SETMASK,
+	SIG_UNBLOCK };
+
+
+#define SA_NOCLDSTOP (1U << 0)
+#define SA_NOCLDWAIT (1U << 1)
+#define SA_NODEFER   (1U << 2)
+#define SA_ONSTACK   (1U << 3)
+#define SA_RESETHAND (1U << 4)
+#define SA_RESTART   (1U << 5)
+#define SA_RESTORER  (1U << 6)
+#define SA_SIGINFO   (1U << 7)
+
+
+typedef unsigned int sigset_t;
+typedef int sig_atomic_t;
+
+
+union sigval {
+	int sival_int;
+	void *sival_ptr;
+};
+
+
+typedef struct {
+	int si_signo;
+	int si_code;
+	pid_t si_pid;
+	uid_t si_uid;
+	void *si_addr;
+	int si_status;
+	union sigval si_value;
+} siginfo_t;
+
+
+struct sigaction {
+	union {
+		sighandler_t sa_handler;
+		void (*sa_sigaction)(int signo, siginfo_t *info, void *context);
+	};
+	sigset_t sa_mask;
+	int sa_flags;
+};
+
+#ifdef __cplusplus
+}
+#endif
 
 #endif
--- a/main.c
+++ b/main.c
@@ -48,6 +48,9 @@ static void main_initthr(void *unused)
 	_hal_start();
 	_usrv_start();
 
+	/* Disable "killing" init thread */
+	proc_current()->sigmask = ~0U;
+
 	lib_printf("main: Starting syspage programs:");
 	syspage_progShow();
 
--- a/proc/process.c
+++ b/proc/process.c
@@ -32,11 +32,6 @@
 #include "userintr.h"
 #include "perf/trace-events.h"
 
-/* Process states */
-#define PREFORK 0
-#define FORKING 1
-#define FORKED  2
-
 typedef struct _process_spawn_t {
 	spinlock_t sl;
 	thread_t *wq;
@@ -1345,25 +1340,37 @@ static void process_restoreParentKstack(thread_t *current, thread_t *parent)
 }
 
 
-__attribute__((noreturn)) static void proc_vforkedExit(thread_t *current, process_spawn_t *spawn, int state)
+void proc_vforkedDied(thread_t *thread, int state)
 {
-	current->ustack = NULL;
-	proc_changeMap(current->process, NULL, NULL, NULL);
+	spinlock_ctx_t sc;
+	process_spawn_t *spawn = thread->execdata;
 
-	proc_kill(current->process);
+	thread->execdata = NULL;
+	thread->ustack = NULL;
+	proc_changeMap(thread->process, NULL, NULL, NULL);
+
+	proc_kill(thread->process);
 
 	/* Only possible in the case of `initthread` exit or failure to fork. */
 	if (spawn->parent == NULL) {
 		hal_spinlockDestroy(&spawn->sl);
 		(void)vm_objectPut(spawn->object);
-
-		proc_threadEnd();
 	}
 	else {
-		process_restoreParentKstack(current, spawn->parent);
+		process_restoreParentKstack(thread, spawn->parent);
 
-		proc_spawnThreadEnd(spawn, state);
+		hal_spinlockSet(&spawn->sl, &sc);
+		spawn->state = state;
+		(void)proc_threadWakeup(&spawn->wq);
+		hal_spinlockClear(&spawn->sl, &sc);
 	}
+}
+
+
+__attribute__((noreturn)) static void proc_vforkedExit(thread_t *current, int state)
+{
+	proc_vforkedDied(current, state);
+	proc_threadEnd();
 }
 
 
@@ -1371,7 +1378,7 @@ void proc_exit(int code)
 {
 	thread_t *current = proc_current();
 	process_spawn_t *spawn = current->execdata;
-	arg_t args[3];
+	arg_t args[2];
 
 	current->process->exit = code;
 
@@ -1383,10 +1390,9 @@ void proc_exit(int code)
 		}
 
 		args[0] = (arg_t)current;
-		args[1] = (arg_t)spawn;
-		args[2] = (arg_t)FORKED;
+		args[1] = (arg_t)FORKED;
 		/* parasoft-suppress-next-line MISRAC2012-RULE_11_1 "Function can accept two different types of first argument" */
-		hal_jmp(proc_vforkedExit, current->kstack + current->kstacksz, NULL, 3, args);
+		hal_jmp(proc_vforkedExit, current->kstack + current->kstacksz, NULL, 2, args);
 	}
 
 	proc_kill(current->process);
@@ -1626,7 +1632,7 @@ int proc_fork(void)
 #ifndef NOMMU
 	thread_t *current, *parent;
 	unsigned int sigmask;
-	arg_t args[3];
+	arg_t args[2];
 
 	err = proc_vfork();
 	if (err == 0) {
@@ -1652,10 +1658,9 @@ int proc_fork(void)
 
 		if (err < 0) {
 			args[0] = (arg_t)current;
-			args[1] = (arg_t)current->execdata;
-			args[2] = (arg_t)err;
+			args[1] = (arg_t)err;
 			/* parasoft-suppress-next-line MISRAC2012-RULE_11_1 "Function can accept two different types of first argument" */
-			hal_jmp(proc_vforkedExit, (unsigned char *)current->kstack + current->kstacksz, NULL, 3, args);
+			hal_jmp(proc_vforkedExit, (unsigned char *)current->kstack + current->kstacksz, NULL, 2, args);
 		}
 		else {
 			hal_cpuEnableInterrupts();
--- a/proc/process.c
+++ b/proc/process.c
@@ -113,6 +113,7 @@ static void process_destroy(process_t *p)
 		vm_kfree(ghost);
 	}
 
+	vm_kfree(p->sigactions);
 	vm_kfree(p->path);
 	vm_kfree(p->argv);
 	vm_kfree(p->envp);
@@ -211,7 +212,7 @@ int proc_start(startFn_t start, void *arg, const char *path)
 	process->ports = NULL;
 
 	process->sigpend = 0;
-	hal_memset(process->sigactions, 0, sizeof(process->sigactions));
+	process->sigactions = NULL;
 	process->tls.tls_base = 0;
 	process->tls.tbss_sz = 0;
 	process->tls.tdata_sz = 0;
@@ -1410,8 +1411,10 @@ static void process_vforkThread(void *arg)
 	/* POSIX: A child created via fork inherits a copy of its parent's signal mask */
 	current->sigmask = parent->sigmask;
 
-	current->process->sigtrampoline = parent->process->sigtrampoline;
-	hal_memcpy(current->process->sigactions, parent->process->sigactions, sizeof(parent->process->sigactions));
+	ret = proc_cloneSigactions(parent->process, current->process);
+	if (ret < 0) {
+		proc_spawnThreadEnd(spawn, ret);
+	}
 
 	hal_spinlockSet(&spawn->sl, &sc);
 	while (spawn->state < FORKING) {
@@ -1668,7 +1671,7 @@ static int process_execve(thread_t *current)
 	process_spawn_t *spawn = current->execdata;
 	thread_t *parent = spawn->parent;
 	vm_map_t *map, *imap;
-	int i;
+	int i, keep;
 
 	/* The old user stack is no longer valid */
 	current->ustack = NULL;
@@ -1709,10 +1712,20 @@ static int process_execve(thread_t *current)
 	current->process->sigpend = 0;
 
 	/* POSIX: signals ignored by the calling process should remain ignored */
-	for (i = 1; i < NSIG; ++i) {
-		/* parasoft-suppress-next-line MISRAC2012-RULE_11_1-a "POSIX compliant definition" */
-		if (current->process->sigactions[i - 1].sa_handler != SIG_IGN) {
-			current->process->sigactions[i - 1].sa_handler = SIG_DFL;
+	if (current->process->sigactions != NULL) {
+		keep = 0;
+		for (i = 1; i < NSIG; ++i) {
+			/* parasoft-suppress-next-line MISRAC2012-RULE_11_1-a "POSIX compliant definition" */
+			if (current->process->sigactions[i - 1].sa_handler == SIG_IGN) {
+				keep = 1;
+			}
+			else {
+				current->process->sigactions[i - 1].sa_handler = SIG_DFL;
+			}
+		}
+		if (keep == 0) {
+			vm_kfree(current->process->sigactions);
+			current->process->sigactions = NULL;
 		}
 	}
 
--- a/proc/process.c
+++ b/proc/process.c
@@ -211,7 +211,7 @@ int proc_start(startFn_t start, void *arg, const char *path)
 	process->ports = NULL;
 
 	process->sigpend = 0;
-	process->sighandler = NULL;
+	hal_memset(process->sigactions, 0, sizeof(process->sigactions));
 	process->tls.tls_base = 0;
 	process->tls.tbss_sz = 0;
 	process->tls.tdata_sz = 0;
@@ -239,12 +239,6 @@ int proc_start(startFn_t start, void *arg, const char *path)
 	}
 
 	return process_getPid(process);
-}
-
-
-void proc_kill(process_t *proc)
-{
-	proc_threadsDestroy(&proc->threads, NULL);
 }
 
 
@@ -292,7 +286,7 @@ static void process_exception(unsigned int n, exc_context_t *ctx)
 		hal_cpuHalt();
 	}
 
-	(void)threads_sigpost(thread->process, thread, signal_kill);
+	(void)threads_sigpost(thread->process, thread, SIGKILL);
 
 	/* Don't allow current thread to return to the userspace,
 	 * it will crash anyway. */
@@ -311,7 +305,11 @@ static void process_illegal(unsigned int n, exc_context_t *ctx)
 		hal_cpuHalt();
 	}
 
-	(void)threads_sigpost(process, thread, signal_illegal);
+	(void)threads_sigpost(process, thread, SIGILL);
+
+	if (thread->exit != 0U) {
+		proc_threadEnd();
+	}
 }
 
 
@@ -1411,7 +1409,9 @@ static void process_vforkThread(void *arg)
 
 	/* POSIX: A child created via fork inherits a copy of its parent's signal mask */
 	current->sigmask = parent->sigmask;
-	current->process->sighandler = parent->process->sighandler;
+
+	current->process->sigtrampoline = parent->process->sigtrampoline;
+	hal_memcpy(current->process->sigactions, parent->process->sigactions, sizeof(parent->process->sigactions));
 
 	hal_spinlockSet(&spawn->sl, &sc);
 	while (spawn->state < FORKING) {
@@ -1668,6 +1668,7 @@ static int process_execve(thread_t *current)
 	process_spawn_t *spawn = current->execdata;
 	thread_t *parent = spawn->parent;
 	vm_map_t *map, *imap;
+	int i;
 
 	/* The old user stack is no longer valid */
 	current->ustack = NULL;
@@ -1705,8 +1706,15 @@ static int process_execve(thread_t *current)
 	current->parentkstack = NULL;
 	current->execdata = NULL;
 
-	current->process->sighandler = NULL;
 	current->process->sigpend = 0;
+
+	/* POSIX: signals ignored by the calling process should remain ignored */
+	for (i = 1; i < NSIG; ++i) {
+		/* parasoft-suppress-next-line MISRAC2012-RULE_11_1-a "POSIX compliant definition" */
+		if (current->process->sigactions[i - 1].sa_handler != SIG_IGN) {
+			current->process->sigactions[i - 1].sa_handler = SIG_DFL;
+		}
+	}
 
 	/* Close cloexec file descriptors */
 	(void)posix_exec();
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -41,6 +41,13 @@ const struct lockAttr proc_lockAttrDefault = { .type = PH_LOCK_NORMAL };
 /* Special empty queue value used to wakeup next enqueued thread. This is used to implement sticky conditions */
 static thread_t *const wakeupPending = (void *)-1;
 
+/* Signal default actions */
+enum {
+	SIGNAL_TERMINATE = 0,
+	SIGNAL_TERMINATE_THREAD,
+	SIGNAL_IGNORE,
+};
+
 static struct {
 	vm_map_t *kmap;
 	spinlock_t spinlock;
@@ -812,21 +819,29 @@ void proc_threadDestroy(thread_t *t)
 }
 
 
-void proc_threadsDestroy(thread_t **threads, const thread_t *except)
+static void _proc_threadsDestroy(thread_t **threads, const thread_t *except)
 {
 	thread_t *t;
-	spinlock_ctx_t sc;
 
-	hal_spinlockSet(&threads_common.spinlock, &sc);
 	t = *threads;
 	if (t != NULL) {
 		do {
 			if (t != except) {
 				_proc_threadExit(t);
 			}
+			/* parasoft-suppress-next-line MISRAC2012-DIR_4_1 "procnext is never NULL, and *threads is checked earlier" */
 			t = t->procnext;
 		} while (t != *threads);
 	}
+}
+
+
+void proc_threadsDestroy(thread_t **threads, const thread_t *except)
+{
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	_proc_threadsDestroy(threads, except);
 	hal_spinlockClear(&threads_common.spinlock, &sc);
 }
 
@@ -845,6 +860,15 @@ void proc_reap(void)
 	hal_spinlockClear(&threads_common.spinlock, &sc);
 
 	threads_put(ghost);
+}
+
+
+void proc_kill(process_t *proc)
+{
+	spinlock_ctx_t sc;
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	_proc_threadsDestroy(&proc->threads, NULL);
+	hal_spinlockClear(&threads_common.spinlock, &sc);
 }
 
 
@@ -1279,42 +1303,99 @@ static time_t _proc_nextWakeup(void)
  */
 
 
-int threads_sigpost(process_t *process, thread_t *thread, int sig)
+static int threads_sigmutable(int sig)
 {
-	spinlock_ctx_t sc;
-	u32 sigbit;
-
 	switch (sig) {
-		case signal_segv:
-		/* parasoft-suppress-next-line MISRAC2012-RULE_16_1 MISRAC2012-RULE_16_3 "Intentional fall-through" */
-		case signal_illegal:
-			if (process->sighandler != NULL) {
-				break;
+		/* POSIX: SIGKILL and SIGSTOP cannot be caught or ignored */
+		case SIGKILL:
+		case SIGSTOP:
+		case PH_SIGCANCEL:
+		case SIGNULL:
+			return 0;
+		default:
+			return ((sig < 0) || (sig > NSIG)) ? 0 : 1;
+	}
+}
+
+
+static int _threads_sigdefault(process_t *process, thread_t *thread, int sig)
+{
+	switch (sig) {
+		case SIGHUP:
+		case SIGINT:
+		case SIGQUIT:
+		case SIGILL:
+		case SIGTRAP:
+		case SIGABRT: /* And SIGIOT */
+		case SIGEMT:
+		case SIGFPE:
+		case SIGBUS:
+		case SIGSEGV:
+		case SIGSYS:
+		case SIGPIPE:
+		case SIGALRM:
+		case SIGTERM:
+		case SIGIO:
+		case SIGXCPU:
+		case SIGXFSZ:
+		case SIGVTALRM:
+		case SIGPROF:
+		case SIGUSR1:
+		case SIGUSR2:
+		case SIGKILL:
+			process->exit = sig * (int)(1UL << 8U);
+			_proc_threadsDestroy(&process->threads, NULL);
+			return SIGNAL_TERMINATE;
+
+		case SIGURG:
+		case SIGCHLD:
+		case SIGWINCH:
+		case SIGINFO:
+		case SIGCONT: /* TODO: Continue process. */
+		case SIGTSTP: /* TODO: Stop process. */
+		case SIGTTIN: /* TODO: Stop process. */
+		case SIGTTOU: /* TODO: Stop process. */
+		case SIGSTOP: /* TODO: Stop process. */
+		case SIGNULL:
+			return SIGNAL_IGNORE;
+
+		case PH_SIGCANCEL:
+			if (thread != NULL) {
+				_proc_threadExit(thread);
 			}
-
-		/* Fall-through */
-		case signal_kill:
-			proc_kill(process);
-			return EOK;
-
-		case signal_cancel:
-			proc_threadDestroy(thread);
-			return EOK;
-
-		case 0:
-			return EOK;
+			return SIGNAL_TERMINATE_THREAD;
 
 		default:
-			/* Handles any value of 'sig' not covered by the case labels. */
-			break;
+			return -EINVAL;
 	}
+}
 
-	if ((sig < 0) || (sig >= NSIG)) {
-		return -EINVAL;
-	}
-	sigbit = (u32)1U << (unsigned int)sig;
+
+int threads_sigpost(process_t *process, thread_t *thread, int sig)
+{
+	u32 sigbit;
+	spinlock_ctx_t sc;
 
 	hal_spinlockSet(&threads_common.spinlock, &sc);
+
+	if ((sig < 0) || (sig > NSIG)) {
+		hal_spinlockClear(&threads_common.spinlock, &sc);
+		return -EINVAL;
+	}
+
+	if (sig == PH_SIGCANCEL || sig == SIGNULL) {
+		(void)_threads_sigdefault(process, thread, sig);
+		hal_spinlockClear(&threads_common.spinlock, &sc);
+		return EOK;
+	}
+
+	/* parasoft-suppress-next-line MISRAC2012-RULE_11_1-a "POSIX compliant definition" */
+	if (process->sigactions[sig - 1].sa_handler == SIG_IGN) {
+		hal_spinlockClear(&threads_common.spinlock, &sc);
+		return EOK;
+	}
+
+	sigbit = (u32)1U << (unsigned int)sig;
 
 	if (thread != NULL) {
 		thread->sigpend |= sigbit;
@@ -1345,7 +1426,13 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
 		}
 	}
 
-	(void)hal_cpuReschedule(&threads_common.spinlock, &sc);
+	if (((sigbit & ~thread->sigmask) != 0U) && (process->sigactions[sig - 1].sa_handler == SIG_DFL)) {
+		(void)_threads_sigdefault(process, thread, sig);
+		thread->sigpend &= ~sigbit;
+		process->sigpend &= ~sigbit;
+	}
+
+	hal_spinlockClear(&threads_common.spinlock, &sc);
 
 	return EOK;
 }
@@ -1356,14 +1443,45 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
 #ifndef KERNEL_SIGNALS_DISABLE
 
 	unsigned int sig;
+	sighandler_t handler;
+	int defaultAction;
 
 	sig = (selected->sigpend | proc->sigpend) & ~selected->sigmask;
-	if ((sig != 0U) && (proc->sighandler != NULL)) {
+	while (sig != 0U) {
 		sig = hal_cpuGetLastBit(sig);
+		handler = proc->sigactions[sig - 1U].sa_handler;
 
-		if (hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sighandler, signalCtx, (int)sig, oldmask, src) == 0) {
-			selected->sigpend &= ~(0x1U << sig);
-			proc->sigpend &= ~(0x1U << sig);
+		if (handler == SIG_DFL) {
+			defaultAction = _threads_sigdefault(proc, selected, (int)sig);
+		}
+		else {
+			defaultAction = -1;
+		}
+
+		if ((defaultAction == SIGNAL_TERMINATE) || (defaultAction == SIGNAL_TERMINATE_THREAD)) {
+			return -1;
+		}
+
+		/* parasoft-suppress-next-line MISRAC2012-RULE_11_1-a "POSIX compliant definition" */
+		if ((handler == SIG_IGN) || (defaultAction == SIGNAL_IGNORE) || (defaultAction == -EINVAL)) {
+			selected->sigpend &= ~(u32)(1UL << sig);
+			proc->sigpend &= ~(u32)(1UL << sig);
+
+			/* Check for other signals */
+			sig = (selected->sigpend | proc->sigpend) & ~selected->sigmask;
+			continue;
+		}
+
+		/* POSIX: sa_mask should be ORed with current process signal mask */
+		selected->sigmask |= proc->sigactions[sig - 1U].sa_mask;
+		if (((unsigned int)proc->sigactions[sig - 1U].sa_flags & SA_NODEFER) == 0U) {
+			selected->sigmask |= (u32)(1UL << sig);
+		}
+		/* TODO: Handle other sa_flags */
+
+		if (hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sigtrampoline, handler, signalCtx, (int)sig, oldmask, src) == 0) {
+			selected->sigpend &= ~(u32)(1UL << sig);
+			proc->sigpend &= ~(u32)(1UL << sig);
 			return 0;
 		}
 	}
@@ -1371,6 +1489,38 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
 #endif
 
 	return -1;
+}
+
+
+int threads_setSigaction(int sig, sigtrampolineFn_t trampoline, const struct sigaction *act, struct sigaction *old)
+{
+	process_t *process;
+	spinlock_ctx_t sc;
+
+	if ((sig <= 0) || (sig >= NSIG)) {
+		return -EINVAL;
+	}
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	process = _proc_current()->process;
+
+	if (trampoline != NULL) {
+		process->sigtrampoline = trampoline;
+	}
+	if (old != NULL) {
+		hal_memcpy(old, &process->sigactions[sig - 1], sizeof(struct sigaction));
+	}
+	if (act != NULL) {
+		if (threads_sigmutable(sig) == 0) {
+			hal_spinlockClear(&threads_common.spinlock, &sc);
+			return -EINVAL;
+		}
+
+		hal_memcpy(&process->sigactions[sig - 1], act, sizeof(struct sigaction));
+	}
+
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+	return 0;
 }
 
 
@@ -1391,7 +1541,7 @@ void threads_setupUserReturn(void *retval, cpu_context_t *ctx)
 
 	if (_threads_checkSignal(thread, thread->process, signalCtx, thread->sigmask, SIG_SRC_SCALL) == 0) {
 		/* parasoft-suppress-next-line MISRAC2012-RULE_11_1 "f is passed to function hal_jmp which need void * type" */
-		f = thread->process->sighandler;
+		f = thread->process->sigtrampoline;
 		hal_spinlockClear(&threads_common.spinlock, &sc);
 		hal_jmp(f, kstackTop, hal_cpuGetUserSP(signalCtx), 0, NULL);
 		/* no return */
@@ -1426,9 +1576,16 @@ int threads_sigsuspend(unsigned int mask)
 	/* check for pending signals before sleep - with the new mask */
 	if (_threads_checkSignal(thread, thread->process, signalCtx, oldmask, SIG_SRC_SCALL) == 0) {
 		/* parasoft-suppress-next-line MISRAC2012-RULE_11_1 "f is passed to function hal_jmp which need void * type" */
-		f = thread->process->sighandler;
+		f = thread->process->sigtrampoline;
 		hal_spinlockClear(&threads_common.spinlock, &sc);
 		hal_jmp(f, kstackTop, hal_cpuGetUserSP(signalCtx), 0, NULL);
+		/* no return */
+	}
+
+	/* check if thread wasn't killed by signal */
+	if (thread->exit != 0U) {
+		hal_spinlockClear(&threads_common.spinlock, &sc);
+		proc_threadEnd();
 		/* no return */
 	}
 
@@ -1442,7 +1599,7 @@ int threads_sigsuspend(unsigned int mask)
 	hal_spinlockSet(&threads_common.spinlock, &sc);
 	if (_threads_checkSignal(thread, thread->process, signalCtx, oldmask, SIG_SRC_SCALL) == 0) {
 		/* parasoft-suppress-next-line MISRAC2012-RULE_11_1 "f is passed to function hal_jmp which need void * type" */
-		f = thread->process->sighandler;
+		f = thread->process->sigtrampoline;
 		hal_spinlockClear(&threads_common.spinlock, &sc);
 		hal_jmp(f, kstackTop, hal_cpuGetUserSP(signalCtx), 0, NULL);
 		/* no return */
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -1407,10 +1407,6 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
 		if (thread != NULL) {
 			do {
 				if ((sigbit & ~thread->sigmask) != 0U) {
-					if (thread->interruptible != 0U) {
-						_thread_interrupt(thread);
-					}
-
 					break;
 				}
 				thread = thread->procnext;
@@ -1426,11 +1422,16 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
 		}
 	}
 
-	if (((sigbit & ~thread->sigmask) != 0U) &&
-	    ((process->sigactions == NULL) || (process->sigactions[sig - 1].sa_handler == SIG_DFL))) {
-		(void)_threads_sigdefault(process, thread, sig);
-		thread->sigpend &= ~sigbit;
-		process->sigpend &= ~sigbit;
+	if ((sigbit & ~thread->sigmask) != 0U) {
+		if (thread->interruptible != 0U) {
+			_thread_interrupt(thread);
+		}
+
+		if ((process->sigactions == NULL) || (process->sigactions[sig - 1].sa_handler == SIG_DFL)) {
+			(void)_threads_sigdefault(process, thread, sig);
+			thread->sigpend &= ~sigbit;
+			process->sigpend &= ~sigbit;
+		}
 	}
 
 	hal_spinlockClear(&threads_common.spinlock, &sc);
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -1390,7 +1390,7 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
 	}
 
 	/* parasoft-suppress-next-line MISRAC2012-RULE_11_1-a "POSIX compliant definition" */
-	if (process->sigactions[sig - 1].sa_handler == SIG_IGN) {
+	if ((process->sigactions != NULL) && (process->sigactions[sig - 1].sa_handler == SIG_IGN)) {
 		hal_spinlockClear(&threads_common.spinlock, &sc);
 		return EOK;
 	}
@@ -1426,7 +1426,8 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
 		}
 	}
 
-	if (((sigbit & ~thread->sigmask) != 0U) && (process->sigactions[sig - 1].sa_handler == SIG_DFL)) {
+	if (((sigbit & ~thread->sigmask) != 0U) &&
+	    ((process->sigactions == NULL) || (process->sigactions[sig - 1].sa_handler == SIG_DFL))) {
 		(void)_threads_sigdefault(process, thread, sig);
 		thread->sigpend &= ~sigbit;
 		process->sigpend &= ~sigbit;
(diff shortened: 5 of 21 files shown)


proc/threads.c
```diff
@@ -1401,28 +1401,67 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
 
-		case 0:
+		case PH_SIGCANCEL:
+			proc_threadDestroy(thread);
 			return EOK;
 
 		default:
-			break;
+			return EOK;
+	}
```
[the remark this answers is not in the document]
etiaro: invalid signal number does fallback to default

proc/threads.c
```diff
@@ -1466,13 +1507,21 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
-			return 0;
+		if (proc->sigactions[sig].sa_handler != NULL) {
+			selected->sigmask |= proc->sigactions[sig].sa_mask;
+			if ((proc->sigactions[sig].sa_flags & SA_NODEFER) == 0) {
+				selected->sigmask |= (1 << sig);
+			}
+			if ((proc->sigactions[sig].sa_flags & SA_RESETHAND) != 0) {
+				proc->sigactions[sig].sa_handler = SIG_DFL;
+			}
+			if (hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sigtrampoline, proc->sigactions[sig].sa_handler, signalCtx, sig, oldmask, src) == 0) {
```
nalajcie: passing `SIG_DFL` as function pointer you would `blx` to in userspace.

Please add the neccessary test to the `phoenix-rtos-tests` and only after then fix the code

proc/threads.c
```diff
@@ -1466,13 +1507,21 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
-	if ((sig != 0) && (proc->sighandler != NULL)) {
+	if (sig != 0) {
 		sig = hal_cpuGetLastBit(sig);
-
-		if (hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sighandler, signalCtx, sig, oldmask, src) == 0) {
-			selected->sigpend &= ~(1 << sig);
-			proc->sigpend &= ~(1 << sig);
-			return 0;
+		if (proc->sigactions[sig].sa_handler != NULL) {
+			selected->sigmask |= proc->sigactions[sig].sa_mask;
```
nalajcie: please keep the important comments when copying code from `libphoenix`

proc/threads.c
```diff
@@ -1434,11 +1473,13 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
 		if (thread != NULL) {
 			do {
 				if (sigbit & ~thread->sigmask) {
-					if (thread->interruptible) {
+					if (thread->interruptible != 0) {
 						_thread_interrupt(thread);
 					}
 
-					break;
+					if (thread->state == READY) {
```
nalajcie: unrelated change, seems like a bugfix to some obscure but, should be done in a separate commit with description as to WHY it's needed.

Personally I think it might be better to block all signals when the thread is dead (using per-thread signal mask).
Not sure if we shouldn't return `-ESRCH` in this case.
Mayyybe it would be possible to write automatic test for this case.

proc/threads.c
```diff
@@ -1466,13 +1507,21 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
 	sig = (selected->sigpend | proc->sigpend) & ~selected->sigmask;
-	if ((sig != 0) && (proc->sighandler != NULL)) {
+	if (sig != 0) {
 		sig = hal_cpuGetLastBit(sig);
-
-		if (hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sighandler, signalCtx, sig, oldmask, src) == 0) {
-			selected->sigpend &= ~(1 << sig);
-			proc->sigpend &= ~(1 << sig);
-			return 0;
+		if (proc->sigactions[sig].sa_handler != NULL) {
```
nalajcie: race condition: the sa_handler here might be `SIG_DFL` or `SIG_IGN` - signal arrival (set pending flag) -> `signal(SIG_DFL)` -> `_threads_checkSignal` (on syscall `signal` exit).

Not sure if this is the only path, but for sure this needs a more careful approach.

proc/process.c
```diff
@@ -1500,16 +1498,16 @@ int proc_fork(void)
...
+
 	err = proc_vfork();
 	if (err == 0) {
 		current = proc_current();
 
 		/* Mask all signals - during process_copy(), incoming signal might try
 		 * to access our not-yet existent stack */
-		sigmask = current->sigmask;
-		current->sigmask = 0xffffffff;
+		threads_updateSigmask(0xffffffff, 0xffffffff);
```
nalajcie: why not `sigmask = ` here?

include/signal.h
```diff
@@ -17,6 +17,106 @@
...
+#define SIG_IGN ((sighandler_t) - 1)
+#define SIG_ERR ((sighandler_t) - 2)
+
+
+enum { SIG_BLOCK,
+	SIG_SETMASK,
+	SIG_UNBLOCK };
+
+
+#define SA_NOCLDSTOP 1 << 0
```
nalajcie: add `( )`

proc/process.c
```diff
@@ -1571,8 +1573,12 @@ static int process_execve(thread_t *current)
 	current->parentkstack = NULL;
 	current->execdata = NULL;
 
-	current->process->sighandler = NULL;
 	current->process->sigpend = 0;
+	for (i = 0; i < NSIG; ++i) {
```
nalajcie: maybe we should mention that it's a posix requirement?

proc/threads.c
```diff
@@ -1401,28 +1401,67 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+		case SIGTERM:
+		case SIGIO:
+		case SIGXCPU:
+		case SIGXFSZ:
+		case SIGVTALRM:
+		case SIGPROF:
+		case SIGUSR1:
+		case SIGUSR2:
+			hal_spinlockSet(&threads_common.spinlock, &sc);
+			if (process->sigactions[sig].sa_handler != NULL) {
```
nalajcie: IMHO `!= SIG_DFL` would be more semantic

syscalls.c
```diff
@@ -942,19 +942,23 @@ addr_t syscalls_va2pa(void *ustack)
...
-	GETFROMSTACK(ustack, unsigned, mmask, 2);
+	GETFROMSTACK(ustack, int, sig, 0);
+	GETFROMSTACK(ustack, struct sigaction *, act, 1);
+	GETFROMSTACK(ustack, struct sigaction *, old, 2);
+	GETFROMSTACK(ustack, void *, trampoline, 3);
 
-	thread = proc_current();
-	thread->process->sigmask = (mask & mmask) | (thread->process->sigmask & ~mmask);
-	thread->process->sighandler = handler;
+	if ((sig < 0) || (sig >= NSIG) || (sig == SIGKILL) || (sig == SIGSTOP)) {
```
nalajcie: `sig == 0` is valid? what's kept under `process->sigaction[0]`?

Also: we mix invalid input values with posix requirements, we should at least have some comment here.

proc/threads.c
```diff
@@ -2129,3 +2178,39 @@ int _threads_init(vm_map_t *kmap, vm_object_t *kernel)
...
+}
+
+
+unsigned threads_updateSigmask(unsigned mask, unsigned mmask)
+{
+	thread_t *t;
+	unsigned old;
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
```
nalajcie: is it really necessary acquire global lock for mask change?

My understanding is that the thread mask can be changed only within the scope of the current thread - so assuming the (int) write is atomic - no lock is necessary? (if so - maybe we should add explicit comment regarding that in thread_t struct).

I'm not completely against it - but in many places the straightforward `t->sigmask = val` was changed into call with mask 0xfffffffffffff (with hopefully not a typo in number of `f`s) making it less readable and more error-prone.

proc/threads.c
```diff
@@ -1401,28 +1401,67 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
-			proc_threadDestroy(thread);
+		case SIGURG:
+		case SIGCHLD:
+		case SIGWINCH:
+		case SIGINFO:
+		case SIGCONT: /* TODO: Continue process. */
+		case SIGTSTP: /* TODO: Stop process. */
+		case SIGTTIN: /* TODO: Stop process. */
+		case SIGTTOU: /* TODO: Stop process. */
+			hal_spinlockSet(&threads_common.spinlock, &sc);
```
nalajcie: convoluted way of acquiring spinlock, IMHO doesn't solve the race condition when calling the handler completely.

Please think of a more straightforward way (make code more readable).

include/signal.h
```diff
@@ -17,6 +17,106 @@
...
+} siginfo_t;
+
+
+struct sigaction {
+	union {
+		void (*sa_handler)(int);
+		void (*sa_sigaction)(int, siginfo_t *, void *);
+	};
+	sigset_t sa_mask;
+	int sa_flags : 8;
```
nalajcie: this struct is exported to userspace, IMHO field specification needs to be POSIX-compliant.

Also: this bitfield doesn't optimize memory usage and would make the access/store slower

proc/threads.c
```diff
@@ -2129,3 +2178,39 @@ int _threads_init(vm_map_t *kmap, vm_object_t *kernel)
...
+{
+	thread_t *thread;
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	thread = _proc_current();
+
+	thread->process->sigtrampoline = trampoline;
+	if (old != NULL) {
+		*old = thread->process->sigactions[sig];
```
nalajcie: WARN: may trigger call to `memcpy` which is currently not available in kernel (we're working on providing alias for `hal_memcpy` - but as of today this may fail to compile depending on target/compilation flags)

nalajcie (dismissed):
Please don't add me as a reviewer to draft PRs without any extra context (as per our coding guidelines - we're not reviewing drafts, they are work-in-progress until you explicitly mark it as ready-for-review).

As an exception I did a code review of the kernel part, I won't be looking at `libphoenix` part unless explicitly asked.

I'm missing a context regarding as to Why the change is needed.

proc/threads.c
etiaro: It should be testable indeed.
It's not only about being dead, but also uninterruptible sleeping threads:
1st thread is in uninterruptible sleep OR it's dying with state == GHOST
2nd thread is in interruptible sleep

Without this condition we won't deliver signal until one of these threads wakes up.

etiaro: Will be separated into another commit.

proc/threads.c
etiaro: You're right. NULL == SIG_DFL, but SIG_IGN could indeed cause a fail here. As process_t.sigactions changes are protected by thread_common spinlock and so is this whole function, it's enough to explicitly check for both SIG_DFL and SIG_IGN .

include/signal.h
etiaro: Yes, bitfiled will be removed.

The sa_handler/sa_sigaction fields can overlap according to POSIX.

proc/threads.c
etiaro: I tried to address an issue with one thread masking a signal during sigpost, that could lead to not delivering a signal that could be delivered to other thread in interruptible sleep.

Now I see that this lock doesn't really solve anything the way it is, and would require additional work here (eg. _threads_checkSignal)
I'm not sure if this is an important enough consideration to justify the use of global lock?

syscalls.c
etiaro: Right, sig == 0 is invalid, and process->sigaction[0] is unused. I changed the process_t.sigaction size to [31], but now it requires "- 1" on every access.

proc/threads.c
etiaro: Alternative approach would be to indeed set sigmask to block all signals when thread is dead and when it's enqueued in uninterruptible sleep.

nalajcie: IMHO blocking all signals before marking thread as dead is better (we avoid some strange `if` here, just using the standard sigmask behavior. Please note that the thread might still die between being selected for handling the signal and actually running the handler (so we "loose" the signal anyway) - but IMHO it's ok in POSIX (generally speaking - threads and signals don't mix well together)

proc/threads.c
nalajcie: You're not running `DFL` action (maybe killing the process) or `IGN` action (remove the signal from pending mask) here.

proc/threads.c
nalajcie: not sure I follow your use case, I assume it's some king of race condition - please provide a more detailed description.

proc/threads.c
nalajcie: not changed in 1444, please be more thorough

proc/threads.c
nalajcie: don't try to unlock the spinlock only on some paths of of switch-case, this is still error-prone.

IMHO better would be:
```c
/* maybe hal_spinlockSet(&threads_common.spinlock, &sc); if needed */
if (process->sigactions[sig].sa_handler == SIG_DFL) {
      /* hal_spinlockClear() */
      return signal_default(); /* TODO: better function name, maybe reuse it in sigpost */
}
else if /*...*/
```

syscalls.c
nalajcie: IMHO `-1` on every access is error-prone, if we don't loose too many bytes, we might keep the unused entry (but it should be written in the comment that it's unused?)

proc/threads.c
etiaro: Yes, it's about a race condition:
Assume we have a process with 2 threads with empty sigmasks

0. thread2 enters interruptible wait
1. sigpost sets SIGX in sigpend of thread1
2. thread1 changes its sigmask so SIGX is masked
SIGX is not delivered, but it could be delivered to thread2.

It seems that there will be some more similar problems when threads mix with signals, where we effectively loose some signals, eg. during thread death like you mentioned above. 
I'm not sure if I should try to deal them here?

nalajcie: `thread->sigpend` is only set when we want to send signal to a particular thread, not to the whole process. In that case we need to wait for this thread to be available.

Other cases set `proc->sigpend` - and we choose interruptible thread under spinlock?

etiaro: We do not choose an interruptible thread.
We choose any thread with matching sigmask, and then interrupt it if it's interruptible.

You're right about thread/proc->sigpend, but this doesn't change the result of the above situation.

proc/threads.c
etiaro: I will switch to full sigmask idea (in uninterruptible calls to proc_threadWaitEx and thread_exit)

proc/threads.c
nalajcie: ok, so it's not a race condition - but probably choosing thread which we should deliver signal to could be optimized (so - it's not related to any additional locking, etc.).

If so - it should be done at least in separate commit (if not in separate PR as it's an optimization topic).

I'm not familiar enough with thread states and typical use cases to be sure it can be optimized better in usual / worst case (eg. it might be better to post signal to thread which is now running instead of sleeping in any state). I'm not even sure it's worth optimizing - as programs using signals for IPC are either single-threaded or should have one thread dedicated to handling signals with all other threads having them blocked.

Please note that the current implementation is POSIX-conformant (choosing the thread the signal is delivered to is up to the implementation)

proc/threads.c
nalajcie: as I stated in the other comment - this is purely an optimization, not an error/race-condition/POSIX requirement. Treat thread selection as a lowest priority subtask in this PR. First make it work, than maybe optimize.

proc/threads.c
etiaro: I'm out of ideas how to fix this issue in a less cluttered way.
Right now ignoring of signal that should happen due to this race is handled in checkSignal, but termination is handled in setSigaction.
It's tricky to handle termination in checkSignal, as it's being called after thread is already selected inside _thread_schedule. Scheduling it anyway could eg. trigger second SIGSEGV, I'm not convinced that's a good idea. There Also, proc_kill takes threads_common lock.

I'm looking for suggestions of cleaner solutions

include/signal.h
```diff
@@ -17,6 +17,108 @@
...
+#define SIGWINCH  28
+#define SIGINFO   29
+#define SIGUSR1   30
+#define SIGUSR2   31
+
+#define NSIG 32
+
+#define PH_SIGCANCEL 32
+
+#define PH_SIGMAX 32
```
nalajcie: what's the difference between `NSIG` and this value?

proc/process.c
```diff
@@ -1500,15 +1498,18 @@ int proc_fork(void)
 	unsigned sigmask;
 	arg_t args[3];
 
+	current = proc_current();
```
nalajcie: why moved to outer scope? (I asked this previously)

include/signal.h
```diff
@@ -17,6 +17,108 @@
...
+#define SIGVTALRM 26
+#define SIGPROF   27
+#define SIGWINCH  28
+#define SIGINFO   29
+#define SIGUSR1   30
+#define SIGUSR2   31
+
+#define NSIG 32
+
+#define PH_SIGCANCEL 32
```
nalajcie: maybe define `PH_SIGCANCEL` to `0` - that way you might avoid `-1` on access (if we want to keep config structs for this signal). Even if not - it's needed to avoid strange/undefined behavior like `int sigbit = 1 << 32` (undefined because of singed, strange because the result is 0 - so we probably can't mask it?)

Edit: I see in linux they also do `-1` on access, they actually written a getter for sa_handler to avoid stupid errors :)

include/signal.h
nalajcie: what's the difference between `NSIG` and this value?

proc/process.c
nalajcie: why moved to outer scope? (I asked this previously). Seems like this function is unchanged after all.

proc/threads.c
```diff
@@ -1395,34 +1402,97 @@ static time_t _proc_nextWakeup(void)
...
+		switch (threads_sigdefault(sig)) {
+			case SIGNAL_TERMINATE:
+				proc_kill(process);
+				return EOK;
 
-		case signal_cancel:
-			proc_threadDestroy(thread);
-			return EOK;
+			case SIGNAL_TERMINATE_THREAD:
+				proc_threadDestroy(thread);
```
nalajcie: assuming `thread != NULL`

proc/threads.c
```diff
@@ -1464,12 +1534,27 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
-	if ((sig != 0) && (proc->sighandler != NULL)) {
+	if (sig != 0) {
 		sig = hal_cpuGetLastBit(sig);
+		handler = proc->sigactions[sig - 1].sa_handler;
+
+		if ((handler == SIG_IGN) || (handler == SIG_DFL)) {
+			selected->sigpend &= ~(1 << sig);
+			proc->sigpend &= ~(1 << sig);
+			sig = (selected->sigpend | proc->sigpend) & ~selected->sigmask;
+			return 0;
```
nalajcie: return 0 means "we have signal, jump to the trampoline" so this would break spectacularly.

IMHO you need to design and write testcases which highlight the harder code paths and ensure they're handled correctly. Otherwise you won't be able to design the code correctly and I don't want to do it for you.

proc/threads.c
```diff
@@ -1395,34 +1402,97 @@ static time_t _proc_nextWakeup(void)
...
+
+		case SIGURG:
+		case SIGCHLD:
+		case SIGWINCH:
+		case SIGINFO:
+		case SIGCONT: /* TODO: Continue process. */
+		case SIGTSTP: /* TODO: Stop process. */
+		case SIGTTIN: /* TODO: Stop process. */
+		case SIGTTOU: /* TODO: Stop process. */
+		/* passthrough*/
```
nalajcie: ?

hal/armv7m/cpu.c
```diff
@@ -138,7 +138,7 @@ int hal_cpuCreateContext(cpu_context_t **nctx, void *start, void *kstack, size_t
 }
 
 
-int hal_cpuPushSignal(void *kstack, void (*handler)(void), cpu_context_t *signalCtx, int n, unsigned int oldmask, const int src)
+int hal_cpuPushSignal(void *kstack, void (*trampoline)(void), void *handler, cpu_context_t *signalCtx, int n, unsigned int oldmask, const int src)
```
nalajcie: probably more kosher (but harder to implement currently) would be to create 2 stack frames - with trampoline (calling sigreturn) below the handler - and jump directly to the handler

nalajcie (commented):
As you've summarised - the signal handling is now cluttered and really not readable. Probably all of the actual signal handling should be done in `_threads_checkSignal` (or maybe some wrapper function for that) - all other code paths are "optimizations" which are either prone to a race condition or a little bit convoluted (I'm not saying that we shouldn't eg. handle `SIGKILL` as early as possible as an optimization).

You may want to look at other OSes regarding how signal handling is being implemented there (usually in more complex way but the basics - like deciding on action to perform on signal dequeue in the context of receiving thread - are universal).

Please - present  the additional test cases You've implemented the code you're pushing here passes - eg. I don't see setting return code of the process killed by a signal anywhere.

hal/armv7m/cpu.c
etiaro: I guess i can give it a try once the threads.c logic would be fine

proc/threads.c
etiaro: Same as original, the if condition is inside proc_threadDestroy

proc/process.c
etiaro: This was invalid implementation of POSIX specified inheriting parent's sigmask during fork.
Will be moved to process_vforkThread

include/signal.h
etiaro: 0 seems like a cleaner solution indeed
I don't think we want to be able to handle this signal.
Masking could be useful if pthread cancelability states would be handled by kernel

EDIT: 0 conflicts with POSIX definition of kill: 
> If sig is 0 (the null signal), error checking is performed but no signal is actually sent. The null signal can be used to check the validity of pid.

I'll go with 0 = null signal and 32 = cancel, as I think we don't need masks and handlers for any of these two.

proc/threads.c
etiaro: Tests created, referenced in the description.

hal/armv7m/cpu.c
etiaro: Not sure if it's worth the trouble, would require some additional changes apart from stack modification (eg. setting arm lr register in or before hal_jmp to handler)

include/signal.h
```diff
@@ -17,6 +17,106 @@
...
+#define SIGINFO      29
+#define SIGUSR1      30
+#define SIGUSR2      31
+#define PH_SIGCANCEL 32
+
+#define NSIG 32
+
+#define SIG_DFL ((sighandler_t)0)
+#define SIG_IGN ((sighandler_t) - 1)
+#define SIG_ERR ((sighandler_t) - 2)
```
julianuziemblo: nit: formatting bug? format by hand, add clang-format ignores if necessary

include/signal.h
```diff
@@ -17,6 +17,106 @@
...
+
+
+#define SA_NOCLDSTOP (1U << 0U)
+#define SA_NOCLDWAIT (1U << 1U)
+#define SA_NODEFER   (1U << 2U)
+#define SA_ONSTACK   (1U << 3U)
+#define SA_RESETHAND (1U << 4U)
+#define SA_RESTART   (1U << 5U)
+#define SA_RESTORER  (1U << 6U)
+#define SA_SIGINFO   (1U << 7U)
```
julianuziemblo: nit: no need for the latter to be unsigned: just `(1U << 1)` is sufficient

include/signal.h
```diff
@@ -17,6 +17,106 @@
...
+
+
+typedef unsigned int sigset_t;
+typedef int sig_atomic_t;
+
+
+union sigval {
+	int sival_int;
+	void *sival_ptr;
+};
```
julianuziemblo, line 92: I see it was defined separately, but is there a reason this couldn't just be an anonymous union inside of `siginfo_t`?

proc/threads.h
```diff
@@ -186,4 +186,7 @@ int threads_sigsuspend(unsigned int mask);
 void threads_setupUserReturn(void *retval, cpu_context_t *ctx);
 
 
+extern int threads_setSigaction(int sig, void (*trampoline)(void), const struct sigaction *act, struct sigaction *old);
```
julianuziemblo: remove extern

julianuziemblo (commented):
small things. I'm not an expert but for me it looks good.

include/signal.h
etiaro: Yes, POSIX.

etiaro:
Waiting with rerun for master nightly to work properly.

include/signal.h
```diff
@@ -17,8 +17,108 @@
...
+/* clang-format off */
+#define SIG_DFL ((sighandler_t)0)
+#define SIG_IGN ((sighandler_t)-1)
+#define SIG_ERR ((sighandler_t)-2)
+/* clang-format on */
+
+
+enum { SIG_BLOCK,
+	SIG_SETMASK,
+	SIG_UNBLOCK };
```
adamgreloch:
```suggestion
enum { SIG_BLOCK, SIG_SETMASK, SIG_UNBLOCK };
/* clang-format on */
```

include/signal.h
adamgreloch: Update the year and authorship to yourself + people from previous libphoenix signal.h? These changes are significant

syscalls.c
```diff
@@ -1080,7 +1060,14 @@ unsigned int syscalls_signalMask(u8 *ustack)
 	t = proc_current();
 
 	old = t->sigmask;
-	t->sigmask = (mask & mmask) | (t->sigmask & ~mmask);
+	new = (mask & mmask) | (old & ~mmask);
+
+	/* POSIX: It is not possible to block those signals which cannot be ignored.
+	 * This shall be enforced by the system without causing an error to be indicated.
+	 */
```
adamgreloch:
```suggestion
	/* 
	 * POSIX: It is not possible to block those signals which cannot be ignored.
	 * This shall be enforced by the system without causing an error to be indicated.
	 */
```

proc/threads.c
```diff
@@ -1374,6 +1499,102 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+
+	if (parent->sigactions != NULL) {
+		for (i = 1; i < NSIG; ++i) {
+			if (parent->sigactions[i - 1].sa_handler == SIG_DFL) {
+				continue;
+			}
+
+			hal_spinlockClear(&threads_common.spinlock, &sc);
+			child->sigactions = vm_kmalloc(sizeof(struct sigaction) * (u8)(NSIG - 1));
```
adamgreloch:
```suggestion
			child->sigactions = vm_kmalloc(sizeof(struct sigaction) * (size_t)(NSIG - 1));
```

proc/threads.c
```diff
@@ -848,6 +868,15 @@ void proc_reap(void)
 }
 
 
+void proc_kill(process_t *proc)
+{
+	spinlock_ctx_t sc;
+	hal_spinlockSet(&threads_common.spinlock, &sc);
```
adamgreloch, line 913: nit: style seems inconsistent - sometimes there is a newline (proc_threadsDestroy) between var definitions and the code, sometimes there's none (proc_kill)

proc/threads.c
```diff
@@ -41,6 +41,13 @@ const struct lockAttr proc_lockAttrDefault = { .type = PH_LOCK_NORMAL };
 /* Special empty queue value used to wakeup next enqueued thread. This is used to implement sticky conditions */
 static thread_t *const wakeupPending = (void *)-1;
 
+/* Signal default actions */
+enum {
+	SIGNAL_TERMINATE = 0,
+	SIGNAL_TERMINATE_THREAD,
+	SIGNAL_IGNORE,
+};
```
adamgreloch: maybe let's `#define` these to stay consistent with the rest of similar "actions" above

adamgreloch (commented):
Looks good to me, only some minor style remarks

syscalls.c
```diff
@@ -1015,63 +1019,39 @@ addr_t syscalls_va2pa(u8 *ustack)
...
-	if (t != NULL) {
-		threads_put(t);
+	/* parasoft-suppress-next-line MISRAC2012-RULE_11_1 "Use of common address verification routine" */
+	if ((trampoline != NULL) && (vm_mapBelongs(proc, (void *)trampoline, 1U) < 0)) {
+		return -EFAULT;
 	}
 
-	return err;
+	return threads_setSigaction(sig, trampoline, act, old);
 }
```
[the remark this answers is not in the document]
etiaro: A general syscall-safety problem, that will be solved separately in the future by kernel exception handling.

proc/threads.c
```diff
@@ -1374,6 +1499,102 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
+			hal_memcpy(child->sigactions, parent->sigactions, sizeof(struct sigaction) * (u8)(NSIG - 1));
+			break;
+		}
+	}
+
+	child->sigtrampoline = parent->sigtrampoline;
+
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+	return 0;
+}
```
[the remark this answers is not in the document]
etiaro: There is no race condition, as parent process is waiting inside vfork during execution, so it won't be killed until vfork is done.
Will introduce comments regarding required external guarantee of not reaping parent process during `proc_cloneSigactions` execution.

proc/process.c
```diff
@@ -305,7 +295,11 @@ static void process_illegal(unsigned int n, exc_context_t *ctx)
 
 	LIB_ASSERT_ALWAYS(process != NULL, "exception in kernel");
 
-	(void)threads_sigpost(process, thread, signal_illegal);
+	(void)threads_sigpost(process, thread, SIGILL);
+
+	if (thread->exit != 0U) {
+		proc_threadEnd();
+	}
 }
```
adamgreloch, line 303: If the thread has installed a SIGILL handler, it will have thread->exit == 0, so this will return normally back to `exceptions_dispatch` and then potentially back to userspace. Most architectures call threads_setupUserReturn before returning from the dispatcher, but I'm not so sure about the armv7m and armv8m. Have you checked this?

syscalls.c
adamgreloch, line 1259: mask from userspace is written raw through sigsuspend, so SIGKILL and SIGSTOP are maskable. Maybe extract the write guard from L1234 into a common threads_ setter and use it in threads_sigsuspend?

main.c
```diff
@@ -46,6 +46,9 @@ static void main_initthr(void *unused)
 	_hal_start();
 	_usrv_start();
 
+	/* Disable "killing" init thread */
+	proc_current()->sigmask = ~0U;
```
adamgreloch: Doesn't seem to help: sys_tkill(pid, tid, 32) -> posix_tkill -> threads_sigpost -> goes straight to _threads_sigdefault and _proc_threadExit without consulting the sigmask

proc/process.c
```diff
@@ -1696,9 +1706,26 @@ static int process_execve(thread_t *current)
...
+			}
+			else {
+				current->process->sigactions[i - 1].sa_handler = SIG_DFL;
+			}
+		}
+		if (keep == 0) {
+			vm_kfree(current->process->sigactions);
+			current->process->sigactions = NULL;
+		}
+	}
```
adamgreloch: isn't this racy? everyone else accesses `sigactions` under a threads spinlock.

also: doesn't the vm_kfree here break the assumption from `proc_closeSigactions`, that:
```
/* For a running process sigactions should never get freed */
```
?

proc/threads.c
```diff
@@ -1444,6 +1568,103 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
+
+	if ((act != NULL) && (threads_sigmutable(sig) == 0)) {
+		return -EINVAL;
+	}
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	process = _proc_current()->process;
+
+	/* allocate sigactions array if required */
+	if ((act != NULL) && (process->sigactions == NULL) && (act->sa_handler != SIG_DFL)) {
```
adamgreloch, line 1722: don't access userspace memory (`act`) under spinlock

proc/threads.c
```diff
@@ -1444,6 +1568,103 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
+
+	if (old != NULL) {
+		/* sigactions can be null if act.sa_handler == SIG_DFL */
+		if (process->sigactions == NULL) {
+			old->sa_handler = SIG_DFL;
+			old->sa_flags = 0;
+			old->sa_mask = 0;
+		}
+		else {
+			hal_memcpy(old, &process->sigactions[sig - 1], sizeof(struct sigaction));
```
adamgreloch, line 1746: don't access userspace memory (`old`) under spinlock

proc/threads.c
```diff
@@ -1444,6 +1568,103 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
+		}
+	}
+
+	/* sigactions can be null if act.sa_handler == SIG_DFL */
+	if ((act != NULL) && (process->sigactions != NULL)) {
+		hal_memcpy(&process->sigactions[sig - 1], act, sizeof(struct sigaction));
```
adamgreloch, line 1752: don't access userspace memory (`act`) under spinlock

proc/threads.c
```diff
@@ -1426,14 +1519,45 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
 #ifndef KERNEL_SIGNALS_DISABLE
 
 	unsigned int sig;
+	sighandler_t handler;
+	int defaultAction;
 
 	sig = (selected->sigpend | proc->sigpend) & ~selected->sigmask;
-	if ((sig != 0U) && (proc->sighandler != NULL)) {
+	while (sig != 0U) {
 		sig = hal_cpuGetLastBit(sig);
```
adamgreloch: `hal_cpuGetLastBit(hal_cpuGetLastBit(sig))` possible on second loop iteration, please split `sig` into a separate mask and index variables

proc/process.c
```diff
@@ -1696,9 +1706,26 @@ static int process_execve(thread_t *current)
 	current->parentkstack = NULL;
 	current->execdata = NULL;
 
-	current->process->sighandler = NULL;
 	current->process->sigpend = 0;
```
adamgreloch: reinitialize `sa_mask`, `sa_flags`, `sigtrampoline` as well?

BTW: `sigtrampoline` is never initialized

proc/threads.c
adamgreloch: note that on `SIGNAL_TERMINATE` (default action) this will still schedule the killed thread for the quantum. maybe we should just ghost it/recheck exit?

include/signal.h
```diff
@@ -15,8 +15,105 @@
...
+#define SIGUSR1      30
+#define SIGUSR2      31
+#define PH_SIGCANCEL 32
 
 #define NSIG 32
 
+/* clang-format off */
+#define SIG_DFL ((sighandler_t)0)
+#define SIG_IGN ((sighandler_t)-1)
+#define SIG_ERR ((sighandler_t)-2)
```
adamgreloch:
```suggestion
#define SIG_DFL ((sighandler_t)0)
#define SIG_IGN ((sighandler_t)1)
#define SIG_ERR ((sighandler_t)-1)
```

any reason to deviate from glibc/musl/bsd?

proc/process.c
etiaro: As of now, these targets ignore `hal_exceptionsSetHandler` for exceptions others than `EXC_DEFAULT`

main.c
etiaro: This line isn't really necessary and will be removed, as init thread is never outside of SupervisorMode anyway, so it won't be killed.

proc/process.c
etiaro: Please note that this is the only thread of this process (thanks to `while (proc_threadsOther(current) != 0)` above). Hence, there is noone to race with, as each process only uses its own sigactions (apart from vfork).

I don't consider this a "running process". Its maps were destroyed a few lines earlier, making it not runnable even in some corner-case bugged scenarios. 
This an exec transition state that rebuilds process_t struct, which is exactly what's happening here

proc/threads.c
etiaro: While I agree it might cause security issues, we do that constantly anyway when delivering signals inside scheduler

proc/threads.c
etiaro: While I agree it might cause security issues, we do that constantly anyway when delivering signals inside scheduler

proc/process.c
etiaro: sa_mask and sa_flags are ignored with SIG_DFL/SIG_IGN, also sigtrampoline won't be used with these values.
For cleaniness, I'll set sigtrampoline to NULL, but I'll skip modifying `sa_mask` or `sa_flags`, I see no point in doing that, especially 31 times with a `threads_common.spinlock` held.

proc/threads.c
etiaro: This requires larger rewrite of _threads_schedule that I would prefer to avoid in this PR (which already turned out harder to merge than expected as it is).
This is still POSIX-compliant behavior, standard signals do not need to guarantee termination on soonest reschedule or anything alike.

include/signal.h
etiaro: No, probably no good reason

main.c
julianuziemblo: If we don't already have it: could you add a test for this so that any future changes dont remove this required behavior?

proc/process.c
adamgreloch: Then I'd add a comment explaining that this doesn't break the invariant and that the access to sigactions is safe. This code is tricky in general.

proc/threads.c
adamgreloch: Your call, but IMHO we should fix the access in the scheduler some day and here just avoid repeating this bad pattern.

proc/threads.c
adamgreloch: Please add a TODO/FIXME then, to keep track of this

proc/process.c
adamgreloch: Please add a WARN about this, could be in these two hals. This can have drastic consequences once we stop ignoring the exceptions

proc/threads.c
ziemleszcz: Where do we do this constantly? Anyway, an existing bug is not a justification for making it worse. This allows a user-space app to deadlock the kernel at any time, which is a critical bug. This can be easily fixed, so I see no reason not to do it.

proc/process.c
ziemleszcz: Not true. Another process running on a different CPU can send a signal to this thread, so `thread_sigpost()` can access `process->sigaction` after it has already been freed but before it is set to NULL.

etiaro: Will add a threads.c helper with spinlock and adequate comment, sorry for misleading explanation.

proc/process.c
etiaro: Will add as a note for future implementation.

proc/threads.c
```diff
@@ -1415,7 +1496,19 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
 		}
 	}
 
-	(void)hal_cpuReschedule(&threads_common.spinlock, &sc);
+	if ((sigbit & ~thread->sigmask) != 0U) {
+		if (thread->interruptible != 0U) {
```
ziemleszcz: Interrupts unconditionally even when signal action is IGNORE. maybe nove this check after next `if` and use `_threads_sigdefault()` return value in this check to exclude ignored signals (`_threads_sigdefault() != SIGNAL_IGNORE`).

proc/threads.c
etiaro: We do this constantly when receiving a signal with user stack (that's location is purely based on user's SP pointer). Also, I just followed behavior of reducing copies already established in other syscalls like gettime and possibly more.

This is a simple fix that I will implement, but I still think this is larger issue to be addressed in PR like #664.

proc/threads.c
etiaro: I don't think this is really necessary unless we try to implement RealTime signals, which is what I'll put in this comment instead of TODO/FIXME (this is not a bug in context of standard, async signals)

adamgreloch: My concern comes solely from the RT perspective, regardless of POSIX. This impacts the general latency of the system (thread is scheduled on a core for one time quantum more than it should)

proc/threads.c
ziemleszcz: Please point me to the exact location. Reducing copies is fine if it's safe, but in this case, it isn't.

vm/map.c
```diff
@@ -827,7 +827,11 @@ static void map_pageFault(unsigned int n, exc_context_t *ctx)
 
 		LIB_ASSERT_ALWAYS(thread->process != NULL, "exception in kernel");
 
-		(void)threads_sigpost(thread->process, thread, signal_segv);
+		(void)threads_sigpost(thread->process, thread, SIGSEGV);
```
ziemleszcz, line 835: Consider forcing default action exit for hardware fault signals. If the sigal is ignored or blocked it will loop printing dump on each iteration.

syscalls.c
ziemleszcz: The same issue in `syscalls_sigreturn()`.

proc/threads.c
ziemleszcz: Also maybe use `hal_cpuGetFirstBit()` (deliver lowest signal first).

vm/map.c
etiaro: Forcing delivery when it's blocked or ignored probably always makes sense, but I'm not sure if it shouldn't be left as configured when a handler is installed?
Some programs could use that as a tool for greaceful shutdown or some ports as an obscure hack for stackoverflow handling?

EDIT: to keep the handler delivery, we would need a mechanism to force the delivery of this particular signal (in case other signals are pending). Will leave that as a FIXME.

proc/threads.c
etiaro: gettime: 
- https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3f86c550b7ef62626358f57d32baef389a7184ee/syscalls.c#L1118
- https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3f86c550b7ef62626358f57d32baef389a7184ee/proc/threads.c#L1305

scheduler:
- https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3f86c550b7ef62626358f57d32baef389a7184ee/proc/threads.c#L447 (with `signalCtx = (void *)((char *)hal_cpuGetUserSP(selCtx) - sizeof(cpu_context_t))`)
- https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3f86c550b7ef62626358f57d32baef389a7184ee/proc/threads.c#L1434
- hal_cpuPushSignal on each target [fills the `signalCtx` memory](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3f86c550b7ef62626358f57d32baef389a7184ee/hal/armv7a/cpu.c#L114) and [pushes some more things](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/3f86c550b7ef62626358f57d32baef389a7184ee/hal/armv7a/cpu.c#L126) on user stack

etiaro: Copy will be implemented in syscalls.c to keep user-memory handling closer to user entry point.

proc/process.c
etiaro: After taking a closer look, a pending terminating signal should not happen here.

sigpost does handle terminate instantly (by setting appropriate thread flags), while syscalls that could modify handling method also call `threads_setupUserReturn` with `_threads_checkSignal` and do check `thread->exit` before returning to userspace. Will be included inside a note here.

proc/threads.c
etiaro: Okay, that matches realtime signal ordering

proc/threads.c
etiaro: After taking a closer look, a delivery of pending terminating signal should not happen here.

sigpost does handle terminate instantly (by setting appropriate thread flags), while syscalls that could modify handling method/sigmask also call threads_setupUserReturn with _threads_checkSignal and do check thread->exit before returning to userspace. Will be included inside a note here.

EDIT: it could happen on race path in sigpost with unblocking sigmask update.  I'll make sigmask updates synchronized under spinlock to avoid that.

proc/threads.c
etiaro: Good point, thanks

proc/threads.c
```diff
@@ -275,6 +280,11 @@ static void thread_destroy(thread_t *thread)
 	while (thread->locks != NULL) {
 		proc_lockForceUnlock(thread->locks, UNLOCK_DO_YIELD);
 	}
+
+	if (thread->execdata != NULL) {
+		thread->kstack = thread->execkstack;
+		proc_vforkedDied(thread, FORKED);
+	}
```
adamgreloch, line 339: a funny one:
1. proc_resourcesCopy fails in process_vforkThread, so child does proc_spawnThreadEnd (L1429)
2. parent sees negative spawn->state, breaks out of while into `if (isparent != 0)` in L1517 and does vm_kfree(spawn), so thread->execdata is a dangling pointer
3. eventually reaper comes, and proc_vforkedDied does UAF on thread->execdata and double-free on thread->parentkstack

proc/threads.c
```diff
@@ -1425,17 +1520,59 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
+			selected->sigpend &= ~(u32)(1UL << curSig);
+			proc->sigpend &= ~(u32)(1UL << curSig);
+
+			/* Check for other signals */
+			deliveryMask &= ~(u32)(1UL << curSig);
+			continue;
+		}
+
+		/* POSIX: sa_mask should be ORed with current process signal mask */
+		selected->sigmask |= proc->sigactions[curSig - 1U].sa_mask;
```
adamgreloch: `sa_mask` is taken raw from user sigaction, so SIGKILL/SIGSTOP are still blockable. Sanitize it in syscalls_signalAction?

syscalls.c
```diff
@@ -1180,54 +1184,37 @@ addr_t syscalls_va2pa(u8 *ustack)
...
-	(void)proc_put(proc);
-	if (t != NULL) {
-		threads_put(t);
+	if (act != NULL) {
+		hal_memcpy(&kact, act, sizeof(kact));
+	}
+	err = threads_setSigaction(sig, trampoline, (act != NULL ? &kact : NULL), (old != NULL ? &kold : NULL));
+	if (old != NULL) {
+		hal_memcpy(old, &kold, sizeof(kold));
 	}
```
adamgreloch:
```suggestion
	if ((err == EOK) && (old != NULL)) {
		hal_memcpy(old, &kold, sizeof(kold));
}
```

kstack leak

proc/threads.c
```diff
@@ -1425,17 +1520,59 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
...
+		if (((unsigned int)proc->sigactions[curSig - 1U].sa_flags & SA_NODEFER) == 0U) {
+			selected->sigmask |= (u32)(1UL << curSig);
+		}
+		/* TODO: Handle other sa_flags */
+
+		if (hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sigtrampoline, handler, signalCtx, (int)curSig, oldmask, src) == 0) {
```
adamgreloch:
```suggestion
		if (((unsigned int)proc->sigactions[curSig - 1U].sa_flags & SA_NODEFER) == 0U) {
			newmask |= (u32)(1UL << curSig);
		}
		/* TODO: Handle other sa_flags */

		if (hal_cpuPushSignal(selected->kstack + selected->kstacksz, proc->sigtrampoline, handler, signalCtx, (int)curSig, oldmask, src) == 0) {
		_threads_setSigmask(selected, newmask);
```

using _threads_setSigmask is debatable, but why not update the sigmask once push succeeds? (theoretical, becuase push always succeeds now, but for some reason we allow it to fail)

proc/process.c
adamgreloch: > especially 31 times with a threads_common.spinlock held.

lol, fair point.

> I'll set sigtrampoline to NULL

Thanks. It shouldn't hurt much. I am just wary of leaving stuff uninitialized as I see this as a very easy thing to trip over

proc/threads.c
etiaro: yeaaaah

proc/threads.c
etiaro: Yeah, probably a good idea.

proc/threads.c
etiaro: Fixed by NULLing execdata and execstack. Leaving unresolved only to double-check if I didn't miss something.

proc/threads.c
```diff
@@ -1586,14 +1595,19 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
 }
 
 
-static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context_t *signalCtx, unsigned int oldmask, const int src)
+/* With installed handler returns signal number or -EINTR for a terminating signal */
+static int _threads_checkSignal(thread_t *selected, process_t *proc)
```
adamgreloch, line 1599: Maybe rename to `_threads_signalDispatch`? It does much more than just a check

etiaro: Well, it definitely does less than previous checkSignal did huh.
But yeah, I had some trouble to come up with a good names here.
I'd like to highlight the difference to trySignalDeliver, mainly that one only delivers termination/ignore while the other also does preparation for handler call whenever needed.

Decided to stay with `check` to at least keep it different enough from `deliver` to reduce confusion between two functions. I'm open to suggestions for better names for both functions though.

adamgreloch: ok, let's keep the `check`. Naming is the last concern for now

adamgreloch:
Seems sane. @ziemleszcz, do you spot anything else?

(we should still wait for nightly CI though)

proc/process.c
```diff
@@ -305,7 +295,11 @@ static void process_illegal(unsigned int n, exc_context_t *ctx)
 
 	LIB_ASSERT_ALWAYS(process != NULL, "exception in kernel");
 
-	(void)threads_sigpost(process, thread, signal_illegal);
+	(void)threads_sigpost(process, thread, SIGILL);
```
ziemleszcz, line 303: Please add FIXME comment just like for `SIGSEGV` in `map_pageFault()`.

include/signal.h
```diff
@@ -15,8 +15,105 @@
...
+#define SIGIO        23
+#define SIGXCPU      24
+#define SIGXFSZ      25
+#define SIGVTALRM    26
+#define SIGPROF      27
+#define SIGWINCH     28
+#define SIGINFO      29
+#define SIGUSR1      30
+#define SIGUSR2      31
+#define PH_SIGCANCEL 32
```
ziemleszcz: Nitpick: use `SIGCANCEL` and just add a comment that this is a custom Phoenix-RTOS signal. `PH_SIGCANCEL` looks out of place.

include/signal.h
```diff
@@ -15,8 +15,105 @@
...
 
 #define NSIG 32
```
ziemleszcz: This is misleading. There are 33 signals (numbers 0 to 32). It also makes signal number validation look wrong.

proc/threads.c
```diff
@@ -1478,35 +1577,265 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+}
+
+
+int threads_setSigaction(int sig, sigtrampolineFn_t trampoline, const struct sigaction *act, struct sigaction *old)
+{
+	process_t *process;
+	struct sigaction *sa = NULL;
+	spinlock_ctx_t sc;
+
+	if ((sig <= 0) || (sig >= NSIG)) {
```
ziemleszcz, line 1714: Add a short comment why SIGNULL and SICANCEL are not allowed.

include/signal.h
etiaro: I'd prefer to have two numbers then. 
`NSIG=32` for standard POSIX, handlable signals (well, handlable comes with some exceptions) and `NSIG_TOTAL=33` to keep it clean and avoid arbitrary -2 in some places.

proc/threads.c
etiaro: Not sure what you mean here. This Particular if allows both signull and sigcancel. As for the next condition, it call sigmutable function, I'm not sure what comment could explain `if(!sigmutable) return -EINVAL;` further.

I'll add some explanation comments inside `threads_sigmutable()`

ziemleszcz: This `if` returns -EINVAL  for 0 and 32.

etiaro: Oof. Yes it does.
Still, I'd rather change this if and leave the explanation  in `threads_sigmutable`.

proc/process.c
ziemleszcz: Pre-exisiting issue: negative exit status (see for example `syscalls_sys_exit()`).

proc/threads.c
```diff
@@ -1481,35 +1577,265 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+		hal_spinlockClear(&threads_common.spinlock, &sc);
+		sa = vm_kmalloc(sizeof(struct sigaction) * (size_t)(NSIG - 1));
+		if (sa == NULL) {
+			return -ENOMEM;
+		}
+
+		hal_spinlockSet(&threads_common.spinlock, &sc);
+		/* for a running process this array should never get freed, but allocation race can happen here */
+		if (process->sigactions == NULL) {
+			hal_memset(sa, 0, sizeof(struct sigaction) * (size_t)(NSIG - 1));
```
ziemleszcz: Could you zero `sa` before acquiring global spinlock?

proc/process.c
etiaro: This is kind of gray-area when it comes to POSIX (dying in kernelspace during exec()), but yeah, I guess we can make it `-err`

ziemleszcz: I don't think it makes sense to return errno here. Maybe EXIT_FAILURE or some other predefined code? Another option is SIGKILL signal.

proc/threads.c
```diff
@@ -1130,17 +1162,23 @@ int proc_threadNanoSleep(time_t *sec, long int *nsec, int absolute)
...
 	hal_spinlockSet(&threads_common.spinlock, &tsc);
 
-	if (((flags & THREAD_WAIT_INTERRUPTIBLE) != 0U) && (_proc_current()->exit != 0U)) {
-		/* Waiting in this state can lead to becoming a hanging zombie */
-		hal_spinlockClear(&threads_common.spinlock, &tsc);
-		return -EINTR;
+	thread = _proc_current();
+
+	if ((flags & THREAD_WAIT_INTERRUPTIBLE) != 0U) {
+		err = _threads_checkSignal(thread, thread->process);
```
etiaro, line 1176: This "race-catcher" causes problems - some places in our codebase do 
`while(err!=EINTR) { ... -> waitInterruptible(); }` (catched exactly such deadlock inside posix_fileDeref at phoenix-rtos-kernel/posix/posix.c:129)

I'm working on a better solution here, suggestions will be appricieted.

EDIT: I don't think that anything better than avoiding this looping pattern in kernel is possible (userspace is fine, as handler will be called on kernel->userspace boundary, but we can't break out of kernel to userspace handler this way).

proc/threads.c
```diff
@@ -1481,35 +1580,266 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+}
+
+
+int threads_setSigaction(int sig, sigtrampolineFn_t trampoline, const struct sigaction *act, struct sigaction *old)
+{
+	process_t *process;
+	struct sigaction *sa = NULL;
+	spinlock_ctx_t sc;
+
+	if ((sig < 0) || (sig >= NSIG_TOTAL)) {
```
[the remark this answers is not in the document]
etiaro: Correct, accidentaly introduced after last review fixes. Will revert to <= with appropriate comment this time.

proc/threads.c
```diff
@@ -1481,35 +1580,266 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+
+	/* allocate sigactions array if required */
+	if ((act != NULL) && (process->sigactions == NULL) && (act->sa_handler != SIG_DFL)) {
+		hal_spinlockClear(&threads_common.spinlock, &sc);
+		sa = vm_kmalloc(sizeof(struct sigaction) * (size_t)(NSIG - 1));
+		if (sa == NULL) {
+			return -ENOMEM;
+		}
+
+		hal_memset(sa, 0, sizeof(struct sigaction) * (size_t)(NSIG - 1));
```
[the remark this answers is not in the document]
etiaro: Fix by changing function entry to return EINVAL on SIGCANCEL

proc/threads.c
```diff
@@ -1481,35 +1580,266 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
+
+			hal_spinlockClear(&threads_common.spinlock, &sc);
+			child->sigactions = vm_kmalloc(sizeof(struct sigaction) * (size_t)(NSIG - 1));
+			if (child->sigactions == NULL) {
+				return -ENOMEM;
+			}
+
+			/* For a running process sigactions should never get freed */
+			hal_spinlockSet(&threads_common.spinlock, &sc);
+			hal_memcpy(child->sigactions, parent->sigactions, sizeof(struct sigaction) * (size_t)(NSIG - 1));
```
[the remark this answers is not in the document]
etiaro: SIGCANCEL is not expected to work with sigaction (EINVAL). It is unmodifiable, same as on master.

proc/threads.c
etiaro: fileDeref seems to be the only remaining loop of this kind. Changed proc_close to be uninterruptible, as the rest of its calls (apart fileDeref one) ignored the return value intirely.

proc/threads.c
```diff
@@ -1481,35 +1580,267 @@ int threads_sigpost(process_t *process, thread_t *thread, int sig)
...
```
ziemleszcz, line 1796: Copy before acquiring the spinlock?

proc/threads.c
```diff
@@ -1415,42 +1453,107 @@ static time_t _proc_nextWakeup(void)
...
+	int performedAction = -1;
 
-		default:
-			/* Handles any value of 'sig' not covered by the case labels. */
-			break;
+	LIB_ASSERT((thread == NULL) || (process == thread->process), "Sigpost to thread with mismatched process");
+
+	/* Disable "killing" kernel threads */
+	if (((thread != NULL) && (thread->process == NULL)) || (process->mapp == NULL)) {
+		return -ESRCH;
```
ziemleszcz: Use `-PERM`. Also I think this check should be done under `&threads_common.spinlock` (`process->mapp`).

include/signal.h
```diff
@@ -15,8 +15,106 @@
...
+/* clang-format off */
+#define SIG_DFL ((sighandler_t)0)
+#define SIG_IGN ((sighandler_t)1)
+#define SIG_ERR ((sighandler_t)-1)
+
+
+enum { SIG_BLOCK, SIG_SETMASK, SIG_UNBLOCK };
+/* clang-format on */
+
+#define SA_NOCLDSTOP (1U << 0)
```
ziemleszcz: Please add a TODO/FIXME comment to indicate which flags are currently not implemented.

proc/threads.c
etiaro: -EPERM it will be.

We already do this check without spinlock [here](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/b199fd6ff6db8e7e002f69847cc180faa079dc67/syscalls.c#L460).
I'm not sure why would it require spinlock? If process map is being changed from/to NULL, it still would race for spinlock acquisition with proc_changeMap (this particular function does not use target process memory, so it won't prevent faults with signal push anyway). 
Ofc This is a marginal change and I can change that, but if that's an issue then syscalls@targetGet should be fixed too.

proc/threads.c
etiaro: parent->sigactions coherency is protected with that spinlock

etiaro:
I'm removing commit enabling nightly.
[Here is last nightly run](https://github.com/phoenix-rtos/phoenix-rtos-kernel/actions/runs/35085728637?pr=670), currently only applying marginal fixes (comments and error code change).

proc/threads.c
ziemleszcz: But this is freshly allocated memory. Before you set `sigaction` pointer it should be safe to copy to it.

include/signal.h
etiaro: also removed non-POSIX unused SA_RESTORER flag.

proc/threads.c
etiaro: Freshly allocated for child, not for parent (source of memcpy).
`vfork()` only hangs the thread it was called from, so other parent threads might be calling sigactions while this copy is done. I know multithreaded vfork is sketchy, but it seems to work.

ziemleszcz: OK.

ziemleszcz (approved):

etiaro:
Waiting for nighly run to finish

PR #672, opened 2025-07-14
proc: fix threadCreate and kill race condition

etiaro:
There's a race condition that allows a process to create thread during its termination, which will continue to run when the process should be already dead.

JIRA: RTOS-1079

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


proc/threads.c
```diff
@@ -1009,6 +1013,8 @@ void proc_threadsDestroy(thread_t **threads)
 
 	hal_spinlockSet(&threads_common.spinlock, &sc);
 	if ((t = *threads) != NULL) {
+		t->process->dead = 1;
```
Darchiv: What if the last thread is destroyed by `proc_threadDestroy()` (singular)? `dead = 1` is done only here. Is the race possible there?

proc/process.h
```diff
@@ -45,6 +45,7 @@ typedef struct _process_t {
 	vm_map_t *imapp;
 	pmap_t *pmapp;
 	int exit;
+	unsigned dead : 1;
```
Darchiv: nitpick: Maybe stick this bitfield to other bitfields instead of separating it?

proc/threads.c
etiaro: IMO, if we are killing a single thread, then its our responsibility to keep in mind that there can be other ones spawned in the meantime, e.g. just before getting a lock inside proc_threadDestroy.
Attempt to make it look like it gives a guarantee of killing a process in certain circumstances can lead to errors.

proc/threads.c
```diff
@@ -821,8 +821,12 @@ int proc_threadCreate(process_t *process, void (*start)(void *), int *id, unsign
 	else {
 		hal_spinlockSet(&threads_common.spinlock, &sc);
 	}
-	/* Insert thread to scheduler queue */
 
+	if ((process != NULL) && (process->dead != 0)) {
+		t->exit = THREAD_END;
+	}
```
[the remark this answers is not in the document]
agkaminski: My man, there's spinlock just above

PR #675, opened 2025-07-28
posix/unix: add support for SOCK_DGRAM sockets in connect()

ziemleszcz:
This change introduces initial support for connecting SOCK_DGRAM sockets. The following limitations still apply:

 - disconnecting a SOCK_DGRAM socket using AF_UNSPEC is not yet implemented
 - reconnecting an already connected SOCK_DGRAM socket is not supported
 - a connected SOCK_DGRAM socket may still receive datagrams from other sockets, which is not POSIX compliant

Other changes:

 - minor MISRA compliance fixes
 - removed unix_lookSocket() function
 - replaced variable names for consistency (using r for the remote socket)
 - return EINVAL from bind(), connect() and send() when the address family is not AF_UNIX
 - return EPROTOTYPE from connect() and send() when the socket type does not match
 - clear US_PEER_CLOSED flag on the first send() for SOCK_DGRAM sockets
 - changed some error codes

JIRA: DTR-502

## Types of changes

## How Has This Been Tested?

- [x] New test added: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/407).
- [x] Tested by hand on: (imx6ull, ia32-generic-qemu).

## Checklist:

## Special treatment

### DIFF
--- a/posix/unix.c
+++ b/posix/unix.c
@@ -5,8 +5,8 @@
  *
  * POSIX-compatibility module, UNIX sockets
  *
- * Copyright 2018, 2020 Phoenix Systems
- * Author: Jan Sikorski, Pawel Pisarczyk
+ * Copyright 2018, 2020, 2025 Phoenix Systems
+ * Author: Jan Sikorski, Pawel Pisarczyk, Ziemowit Leszczynski
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -21,6 +21,11 @@
 #include "posix_private.h"
 #include "fdpass.h"
 
+
+/*
+ * FIXME: this module has multiple potential race conditions. For example,
+ * in unix_bind(), unix_connect(), and other related functions.
+ */
 
 #define US_DEF_BUFFER_SIZE SIZE_PAGE
 #define US_MIN_BUFFER_SIZE SIZE_PAGE
@@ -53,7 +58,15 @@ typedef struct _unixsock_t {
 
 	spinlock_t spinlock;
 
-	struct _unixsock_t *connect;
+	/* Socket to which this socket is connected to. */
+	struct _unixsock_t *remote;
+
+	/* For SOCK_DGRAM: list of sockets connected to this socket. */
+	struct _unixsock_t *connected;
+
+	/* For other types: list of sockets requesting a connection. */
+	struct _unixsock_t *connecting;
+
 	thread_t *queue;
 	thread_t *writeq;
 } unixsock_t;
@@ -190,7 +203,9 @@ static unixsock_t *unixsock_alloc(unsigned *id, int type, int nonblock)
 	r->nonblock = nonblock;
 	r->buffsz = US_DEF_BUFFER_SIZE;
 	r->fdpacks = NULL;
-	r->connect = NULL;
+	r->remote = NULL;
+	r->connected = NULL;
+	r->connecting = NULL;
 	r->queue = NULL;
 	r->writeq = NULL;
 	r->state = 0;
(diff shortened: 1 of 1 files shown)


posix/unix.c
```diff
@@ -762,34 +825,64 @@
...
 				}
+
+				r = unixsock_get(oid.id);
+				if (r == NULL) {
+					err = -ENOTSOCK;
+					break;
+				}
+
+				if (s->type != r->type) {
+					unixsock_put(r);
```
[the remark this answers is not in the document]
ziemleszcz: Why do you think it would be freed a second time? After the first release, the break exits the do-while loop.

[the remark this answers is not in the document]
nalajcie: @gemini-code-assist Why do you think it would be freed a second time? After the first release, the break exits the do-while loop.

posix/unix.c
```diff
@@ -53,7 +58,15 @@ typedef struct _unixsock_t {
 
 	spinlock_t spinlock;
 
+	/* Socket this one is connected to */
```
adamgreloch:
```suggestion
	/* Socket to which this socket is connected to */
```

posix/unix.c
```diff
@@ -53,7 +58,15 @@ typedef struct _unixsock_t {
...
+	struct _unixsock_t *remote;
+
+	/*
+	 * For SOCK_DGRAM: list of sockets connected to this socket.
+	 * For other types: list of sockets requesting a connection.
+	 */
 	struct _unixsock_t *connect;
```
adamgreloch: Wondering about a better name for this. It was already misleading before the change and now it is even worse as `connect` may wrongly suggest something about the the state of sockets stored while the state varies across socket types (connected vs. requesting a connection). Maybe some general name (`queue`, whatever) would be better, to emphasize that the list serves multiple purposes?

posix/unix.c
```diff
@@ -237,15 +252,34 @@ static unixsock_t *unixsock_get_connected(unixsock_t *s)
 
 static void unixsock_put(unixsock_t *s)
 {
+	unixsock_t *r;
+
 	proc_lockSet(&unix_common.lock);
 	if (--s->refs <= 0) {
```
julianuziemblo: As you're already on fixing MISRA: statements should not have side effects

```suggestion
        s->refs--;
	if (s->refs <= 0) {
```

posix/unix.c
```diff
@@ -750,10 +813,12 @@ static ssize_t recv(unsigned socket, void *buf, size_t len, int flags, struct so
...
+/* TODO: a connected SOCK_DGRAM socket should only receive data from its peer. */
 static ssize_t send(unsigned socket, const void *buf, size_t len, int flags, const struct sockaddr *dest_addr, socklen_t dest_len, fdpack_t *fdpack)
 {
-	unixsock_t *s, *conn;
+	unixsock_t *s, *r;
 	int err;
+	oid_t oid;
 	spinlock_ctx_t sc;
 
 	if ((s = unixsock_get(socket)) == NULL)
```
julianuziemblo: MISRA: assign then check

posix/unix.c
```diff
@@ -1064,20 +1159,20 @@ int unix_poll(unsigned socket, short events)
 		}
 
 		if (events & (POLLOUT | POLLWRNORM | POLLWRBAND)) {
-			if ((conn = unixsock_get_connected(s)) != NULL) {
-				proc_lockSet(&conn->lock);
-				if (conn->type == SOCK_STREAM) {
-					if (_cbuffer_free(&conn->buffer) > 0) {
+			if ((r = unixsock_get_remote(s)) != NULL) {
```
julianuziemblo: MISRA: assign then check

adamgreloch (approved):

PR #682, opened 2025-08-11
coredump: coredump server support

etiaro:
## Description

Introduces kernel thread that uses messaging to inform coredump server process about crashed processes and provide any data required for creating coredump for post-mortem analysis with GDB.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic-qemu, armv7r5f-zynqmp-qemu, sparcv8leon-generic-qemu

## Checklist:

## Special treatment

- #670 
- #685 

### DIFF
--- a/coredump/coredump.c
+++ b/coredump/coredump.c
@@ -1,0 +1,378 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * Coredump server communication
+ *
+ * Copyright 2025 Phoenix Systems
+ * Author: Jakub Klimek
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "coredump.h"
+
+#include "proc/threads.h"
+#include "proc/ports.h"
+#include "include/coredump.h"
+
+static struct {
+	spinlock_t spinlock;
+	thread_t *toDump;
+	process_t *current;
+	thread_t *dumperQ;
+	oid_t oid;
+} coredump_common;
+
+
+int coredump_enqueue(process_t *process)
+{
+	thread_t *thread, *ghost;
+	spinlock_ctx_t sc;
+
+	/* No need to protect process.ghosts, we have the last reference to a dying process */
+	thread = process->ghosts;
+
+	/* Cleanup threads that finished before crash */
+	do {
+		while ((process->ghosts != NULL) && (thread->kstack == NULL)) {
+			ghost = thread;
+			thread = thread->procnext;
+			LIST_REMOVE_EX(&process->ghosts, ghost, procnext, procprev);
+			vm_kfree(ghost);
+		}
+		if (process->ghosts == NULL) {
+			return -ENOENT;
+		}
+		thread = thread->procnext;
+	} while (thread != process->ghosts);
+
+	/* Find crashing thread */
+	do {
+		if ((thread->sigpend & (1U << SIGNULL)) != 0U) {
+			break;
+		}
+		thread = thread->procnext;
+	} while (thread != process->ghosts);
+
+	if ((thread->sigpend & (1U << SIGNULL)) == 0U) {
+		return -ENOENT;
+	}
+
+	hal_spinlockSet(&coredump_common.spinlock, &sc);
+	LIST_ADD(&coredump_common.toDump, thread);
+	(void)proc_threadWakeup(&coredump_common.dumperQ);
+	hal_spinlockClear(&coredump_common.spinlock, &sc);
+
+	return EOK;
+}
+
+
+static size_t coredump_memEntryList(coredump_memseg_t *list, size_t n)
+{
+	map_entry_t *e;
+	size_t segCnt = 0;
+	size_t i = 0;
+
+	(void)proc_lockSet(&coredump_common.current->mapp->lock);
+
+#ifdef NOMMU
+
+	e = coredump_common.current->entries;
+	do {
+		if (((e->prot & PROT_READ) != 0) && ((e->prot & PROT_WRITE) != 0)) {
+			segCnt++;
+			if (i < n) {
+				list[i].startAddr = e->vaddr;
+				list[i].endAddr = e->vaddr + e->size;
+				i++;
+			}
+		}
+		e = e->next;
+	} while (e != coredump_common.current->entries);
+
+#else
+
+	e = lib_treeof(map_entry_t, linkage, lib_rbMinimum(coredump_common.current->mapp->tree.root));
+	while (e != NULL) {
+		if (((e->prot & PROT_READ) != 0U) && ((e->prot & PROT_WRITE) != 0U)) {
+			segCnt++;
+			if (i < n) {
+				list[i].startAddr = e->vaddr;
+				list[i].endAddr = e->vaddr + e->size;
+				i++;
+			}
+		}
+		e = lib_treeof(map_entry_t, linkage, lib_rbNext(&e->linkage));
+	}
+
+#endif
+
+	(void)proc_lockClear(&coredump_common.current->mapp->lock);
+
+	return segCnt;
+}
+
+
+static void coredump_handleMemRead(msg_t *msg, msg_rid_t rid)
+{
+	const coredump_req_t *req = (const coredump_req_t *)msg->i.data;
+	u32 responsePort = req->mem.responsePort;
+	void *startAddr = req->mem.startAddr;
+	size_t len = req->mem.size;
+	msg_t memMsg;
+
+	if (vm_mapBelongs(coredump_common.current, req->mem.startAddr, len) != 0) {
+		msg->o.err = -EINVAL;
+		(void)proc_respond(coredump_common.oid.port, msg, rid);
+		return;
+	}
+	msg->o.err = EOK;
+	(void)proc_respond(coredump_common.oid.port, msg, rid);
+
+	memMsg.type = mtWrite;
+	memMsg.oid.id = 0;
+	memMsg.oid.port = responsePort;
+	memMsg.i.size = len;
+	memMsg.i.data = startAddr;
+	memMsg.o.size = 0;
+
+	(void)proc_sendFromMap(responsePort, &memMsg, coredump_common.current->mapp);
+}
+
+
+static void coredump_handleThreadRead(msg_t *msg, msg_rid_t rid)
+{
+	const coredump_req_t *req = (const coredump_req_t *)msg->i.data;
+	coredump_thread_t *threadResp;
+	thread_t *thread;
+
+	threadResp = (coredump_thread_t *)msg->o.data;
+	thread = coredump_common.current->ghosts;
+	if (thread != NULL) {
+		do {
+			if (proc_getTid(thread) == req->thread.tid) {
+				break;
+			}
+			thread = thread->procnext;
+		} while (thread != coredump_common.current->ghosts);
+	}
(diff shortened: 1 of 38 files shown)


coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+	hal_spinlockSet(&coredump_common.spinlock, &sc);
+	LIST_ADD(&coredump_common.toDump, thread);
+	proc_threadWakeup(&coredump_common.dumperQ);
+	hal_spinlockClear(&coredump_common.spinlock, &sc);
+
+	return EOK;
+}
+
+
+int coredump_memEntryList(coredump_memseg *list, size_t n)
```
julianuziemblo: `static`?

proc/msg-nommu.c
julianuziemblo, line 54: shouldn't this also be just `senderProc` instead of `sender->proc`? It seems that calling it for different process might give the msg wrong pid and priority

proc/msg.c
julianuziemblo, line 376: same here about `senderProc` and `sender->proc`

proc/process.h
```diff
@@ -27,6 +27,14 @@
 
 #define MAX_PID MAX_ID
 
+#ifdef NOMMU
+typedef struct _reloc {
```
julianuziemblo, line 40: `_reloc` name is not used anywhere now, so it doesn't need to exist

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+#else
+
+	msg->o.err = ENOSYS;
+	proc_respond(coredump_common.oid.port, msg, rid);
+	return ENOSYS;
+#endif
+}
+
+
+int coredump_handleRead(msg_t *msg, msg_rid_t rid)
```
julianuziemblo: `static`?

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+		default:
+			break;
+	}
+	msg->o.err = EINVAL;
+	proc_respond(coredump_common.oid.port, msg, rid);
+	return EINVAL;
+}
+
+
+int coredump_threadCnt(process_t *process)
```
julianuziemblo: `static`?

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+	do {
+		count++;
+		t = t->procnext;
+	} while (t != process->ghosts);
+
+	return count;
+}
+
+
+int coredump_isRunning(int pid)
```
julianuziemblo: `static`?

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+	process_t *p = proc_find(pid);
+	if (p == NULL) {
+		return 0;
+	}
+	proc_put(p);
+	return 1;
+}
+
+
+void coredump_dump(void)
```
julianuziemblo: `static`?

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+		}
+	} while (msg.type != mtClose);
+
+	coredump_common.current->coredump = 0;
+	proc_put(coredump_common.current);
+	coredump_common.current = NULL;
+}
+
+
+void coredump_msgthr(void *arg)
```
julianuziemblo: `static`?

include/coredump.h
```diff
@@ -0,0 +1,77 @@
...
+	int tid;
+	int signo;
+	enum {
+		COREDUMP_TYPE_32,
+		COREDUMP_TYPE_64
+	} type;
+	char path[64];
+	int memSegCnt;
+	int threadCnt;
+} coredump_general;
```
julianuziemblo: the coding standards don't say this, but wouldn't adding `*_t` to type names be more readable?

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+	map_entry_t *e;
+	size_t segCnt = 0;
+	size_t i = 0;
+
+	proc_lockSet(&coredump_common.current->mapp->lock);
+
+#ifdef NOMMU
+	e = coredump_common.current->entries;
+	do {
+		if ((e->prot & PROT_READ) && (e->prot & PROT_WRITE)) {
```
julianuziemblo: MISRA: `!= 0`

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+			}
+		}
+		e = e->next;
+	} while (e != coredump_common.current->entries);
+
+#else
+
+	e = lib_treeof(map_entry_t, linkage, lib_rbMinimum(coredump_common.current->mapp->tree.root));
+	while (e != NULL) {
+		if ((e->prot & PROT_READ) && (e->prot & PROT_WRITE)) {
```
julianuziemblo: MISRA: `!= 0`

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+	const coredump_req *req = (coredump_req *)msg->i.data;
+	u32 responsePort = req->mem.responsePort;
+	void *startAddr = req->mem.startAddr;
+	size_t len = req->mem.size;
+	int ret;
+
+	if (vm_mapBelongs(coredump_common.current, req->mem.startAddr, len) != 0) {
+		msg->o.err = EINVAL;
+		proc_respond(coredump_common.oid.port, msg, rid);
+		return EINVAL;
```
julianuziemblo: The convention seems to be that we return negative values as errors, so:

```suggestion
		msg->o.err = -EINVAL;
		proc_respond(coredump_common.oid.port, msg, rid);
		return -EINVAL;
```

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+
+		case COREDUMP_REQ_RELOC:
+			return coredump_handleRelocRead(msg, rid);
+
+		default:
+			break;
+	}
+	msg->o.err = EINVAL;
+	proc_respond(coredump_common.oid.port, msg, rid);
+	return EINVAL;
```
julianuziemblo: negative error values

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+			if (coredump_isRunning(srvPid)) {
+				continue;
+			}
+			else {
+				break;
+			}
+		}
+
+		if (msg.type == mtRead) {
+			coredump_handleRead(&msg, rid);
```
julianuziemblo: the return value of this function is ignored. If this is intentional - just make it `void` and make return values of internal functions also `void`, if they would also be ignored.

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+	memMsg.o.size = 0;
+
+	ret = proc_sendFromProcess(responsePort, &memMsg, coredump_common.current);
+
+	if (ret != 0) {
+		return ret;
+	}
+	if (memMsg.o.err != 0) {
+		return memMsg.o.err;
+	}
```
julianuziemblo: super nitpick: I like to not leave any empty space between the call and return value check, but that's just preference - I think I've seen you do similar in another place in the code, so I'd do the same here

tl;dr: I'd remove empty line 144, but that's just stylistic choice

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+
+	if (vm_mapBelongs(coredump_common.current, req->mem.startAddr, len) != 0) {
+		msg->o.err = EINVAL;
+		proc_respond(coredump_common.oid.port, msg, rid);
+		return EINVAL;
+	}
+	msg->o.err = EOK;
+	proc_respond(coredump_common.oid.port, msg, rid);
+
+	msg_t memMsg;
```
julianuziemblo: Coding convention in kernel says to define all variables at the top of the function. I'd also hesitate with this one as this is extra 200-or-so bytes to allocate not in every case. Maybe make an inline function that makes and sends the msg to call after the initial checks?

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+				break;
+			}
+			thread = thread->procnext;
+		} while (thread != coredump_common.current->ghosts);
+	}
+
+	if ((thread == NULL) || (proc_getTid(thread) != req->thread.tid)) {
+		msg->o.err = ENOENT;
+		proc_respond(coredump_common.oid.port, msg, rid);
+		return ENOENT;
```
julianuziemblo: negative return values

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+	msg->o.err = EOK;
+	proc_respond(coredump_common.oid.port, msg, rid);
+	return EOK;
+
+#else
+
+	msg->o.err = ENOSYS;
+	proc_respond(coredump_common.oid.port, msg, rid);
+	return ENOSYS;
+#endif
```
julianuziemblo, line 215: super nitpick: add empty line before `#endif`

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+
+static int coredump_handleRelocRead(msg_t *msg, msg_rid_t rid)
+{
+#ifdef NOMMU
+
+	int i;
+	process_t *process = coredump_common.current;
+
+	coredump_reloc *relocs = (coredump_reloc *)msg->o.data;
+	for (i = 0; i < process->relocsz && i < msg->o.size / sizeof(coredump_reloc); i++) {
```
julianuziemblo: MISRA: `size_t i;`

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+
+	hal_spinlockSet(&coredump_common.spinlock, &scp);
+
+	while (coredump_common.toDump == NULL) {
+		proc_threadWait(&coredump_common.dumperQ, &coredump_common.spinlock, 0, &scp);
+	}
+	crashed = coredump_common.toDump;
+	coredump_common.current = crashed->process;
+	LIST_REMOVE(&coredump_common.toDump, crashed);
+	hal_spinlockClear(&coredump_common.spinlock, &scp);
```
julianuziemblo, line 291: super nitpick: add empty line before `hal_spinlockClear`

coredump/coredump.c
```diff
@@ -0,0 +1,389 @@
...
+			coredump_handleRead(&msg, rid);
+		}
+		else if (msg.type == mtClose) {
+			msg.o.err = EOK;
+			proc_respond(coredump_common.oid.port, &msg, rid);
+		}
+		else {
+			msg.o.err = -EINVAL;
+			proc_respond(coredump_common.oid.port, &msg, rid);
+		}
```
julianuziemblo: nitpick: `break` is not called here, so I think `switch` would be easier to read - but it's just a preference

proc/msg-nommu.c
etiaro: I agree that the naming is really confusing here.
Primary objective of this modification was to send data from senderProc's address space (for MMU targets, here it doesn't have much effect).
The messaging itself is handled by sender (proc_current()), so in my opinion pid and priority should remain as it is.

That is probably not a great design choice, I'll try to change this function parameter to source map instead of process to show more directly what it does.

coredump/coredump.c
etiaro: Moved to the top.
This checks are safety-related and should pass if everything is working correctly, so there's no need for separate function here.

etiaro:
Converted to draft and put on hold until signal's rework is merged.
Looking back after a few months, I can see that the architecture of the coredump mechanism requires some additional work, especially around hal exception handling.

PR #685, opened 2025-08-12
hal/armvXm: use cpu_context_t as exc_context_t 

etiaro:
## Description

This way armv7m & armv8m targets follow convention of having cpu_context_t as substructure of exc_context_t and residing at the end of kernel stack when being in kernel space even during fault handling.

Also adjusted SIZE_FPUCTX to include fpscr and padding as described in [ARM docs](https://developer.arm.com/documentation/ddi0403/d/System-Level-Architecture/System-Level-Programmers--Model/ARMv7-M-exception-model/Exception-entry-behavior#:~:text=%271%27%20then-,framesize%20%3D%200x68%3B,-forcealign%20%3D%20%271).

Currently tested only on imxrt117x

## Motivation and Context

 This assumption simplifies producing coredump in unified way among all targets.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt117x-evk

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv8m/arch/exceptions.h
+++ b/hal/armv8m/arch/exceptions.h
@@ -25,21 +25,6 @@
 #define SIZE_CTXDUMP 512 /* Size of dumped context */
 
 
-typedef struct _exc_context_t {
-	/* Saved by ISR */
-	u32 psp;
-	u32 r4;
-	u32 r5;
-	u32 r6;
-	u32 r7;
-	u32 r8;
-	u32 r9;
-	u32 r10;
-	u32 r11;
-	u32 excret;
-
-	/* Saved by hardware */
-	cpu_hwContext_t mspctx;
-} exc_context_t;
+typedef cpu_context_t exc_context_t;
 
 #endif
--- a/hal/armv8m/exceptions.c
+++ b/hal/armv8m/exceptions.c
@@ -49,21 +49,18 @@ void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
 		"12 #Debug", "13 #", "14 #PendSV", "15 #SysTick"
 	};
 	size_t i = 0;
-	u32 msp = (u32)ctx + sizeof(*ctx);
-	const u32 fpu_hwctx_size = ((ctx->excret & EXC_RETURN_FTYPE) == 0) ? SIZE_FPUCTX : 0;
+	const u32 fpu_hwctx_size = ((ctx->irq_ret & EXC_RETURN_FTYPE) == 0) ? SIZE_FPUCTX : 0;
 	u32 psp = ctx->psp;
 	u32 cfsr, far;
 	cpu_hwContext_t *hwctx;
 
 	/* If we came from userspace HW ctx in on psp stack (according to EXC_RETURN) */
-	if ((ctx->excret & EXC_RETURN_SPSEL) != 0) {
+	if ((ctx->irq_ret & EXC_RETURN_SPSEL) != 0) {
 		hwctx = (void *)ctx->psp;
-		msp -= sizeof(cpu_hwContext_t);
 		psp += sizeof(cpu_hwContext_t) + fpu_hwctx_size;
 	}
 	else {
-		hwctx = &ctx->mspctx;
-		msp += fpu_hwctx_size;
+		hwctx = &ctx->hwctx;
 	}
 
 	n &= 0xf;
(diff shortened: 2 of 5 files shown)


hal/armv7m/imxrt/_init.S
```diff
@@ -192,12 +192,23 @@ _syscallend:
...
 _exceptions_dispatch:
 	cpsid if
+	isb
 
-	mrs r0, psp
-	stmdb sp!, {r0, r4-r11, lr}
+	/* if we came from userspace, make space for hw-saved regs */
+	tst lr, #(1 << 2)
+	it ne
+	subne sp, sp, #(26 * 4) /* space for hw-saved ctx */
```
agkaminski: I think the comment is somewhat misleading - we're aligning the `cpu_context_t` struct, not making space for the ctx - it's never put in there. I would like to also suggest writing the size out more explicitly - `#((8 + 18) * 4)`

agkaminski: Also, I think we can skip this aliment - part of the struct will point to the trash, but it's never accessed. On the other hand we'll save some stack space (that is used extensively in the exception handler)

etiaro: Sure, the comment could be better indeed.

The problem shows up when we start accesing this context as `kstack + kstacksz - sizeof(cpu_context_t)`:
Even if we don't want to deliver signals from exceptions on this target, it's still useful for coredump to have  single kind of context

etiaro: I'm also wondering if we shouldn't move s0-s15, fpscr, pad1 from cpu_context_t to cpu_hwContext_t for more clarity, as it seems to me like these are also hardware-saved.
Or am I missing something here?

hal/armv7m/exceptions.c
```diff
@@ -45,7 +45,7 @@ void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, int n)
 	cpu_hwContext_t *hwctx;
 
 	/* If we came from userspace HW ctx in on psp stack */
-	if (ctx->excret == RET_THREAD_PSP) {
+	if (ctx->irq_ret == RET_THREAD_PSP) {
 		hwctx = (void *)ctx->psp;
 		msp -= sizeof(cpu_hwContext_t);
```
jmaksymowicz: After the changes you made I think it would make more sense to take `msp` from the `ctx` structure instead of doing this calculation. However, the way you push `msp` on the stack is not correct, see comment in `imxrt/_init.S`. Then you can just
```c
u32 msp = ctx->msp;
```
witout the need to subtract or add to it in C code. `psp` will still need to be added to/subtracted from.

hal/armv7m/imxrt/_init.S
```diff
@@ -192,12 +192,23 @@ _syscallend:
...
-	mrs r0, psp
-	stmdb sp!, {r0, r4-r11, lr}
+	/* if we came from userspace, align to match whole cpu_context_t, as hw-saved regs are on user stack */
+	tst lr, #(1 << 2)
+	it ne
+	subne sp, sp, #((8 + 18) * 4) /* space corresponding to hw-saved ctx */
+
+	vstmdb sp!, {s16-s31}
+	mov r0, sp
+	str r0, [sp, #-8]!
```
jmaksymowicz: Here we want to push `msp` as it was before the exception occurred. 
* If coming from userspace, `msp` is already the value we want upon exception handler entry.
* If coming from kernel, hw-saved context was already pushed onto main stack as we enter the exception handler, so we need to add the size of that context to `msp` to get that value.
```suggestion
	mov r0, sp /* save MSP to be stored in cpu_context_t */
	tst lr, #(1 << 2)
	ite ne
	subne sp, sp, #((8 + 18) * 4) /* If coming from userspace, make room on the stack for hardware-stored context */
	addeq r0, r0, #((8 + 18) * 4) /* If coming from kernel, calculate msp as it was before the exception  */

	vstmdb sp!, {s16-s31}
	str r0, [sp, #-8]! /* Store msp, skip over pad0 */

```

Similar suggestion goes for other targets (with different hw-context size).

etiaro: Thanks, nice catch

hal/armv7m/exceptions.c
etiaro: Good point, let's make the context struct values accurate.

PR #687, opened 2025-08-14
syscalls: add offset handling to read/write

R4ken:
Modify read and write syscalls allowing passing offset as argument

JIRA: RTOS-1088

## Description
Add offset as an argument to read and write syscalls. This modification allows reading and writing at specified position as well as using file descriptor's current offset (value of argument < 0)

## Motivation and Context
This modification allows using these syscalls for both read/write and pread/pwrite POSIX functions

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/418
- [x] Tested by hand on: (ia32-generic-qemu, armv7a9-zynq7000-qemu)

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/libphoenix/pull/429 
    - https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/418

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -706,13 +706,13 @@ int posix_close(int fildes)
 }
 
 
-ssize_t posix_read(int fildes, void *buf, size_t nbyte)
+ssize_t posix_read(int fildes, void *buf, size_t nbyte, off_t offset)
 {
-	TRACE("read(%d, %p, %u)", fildes, buf, nbyte);
+	TRACE("read(%d, %p, %zu, %jd)", fildes, buf, nbyte, (intmax_t)offset);
 
 	open_file_t *f;
 	ssize_t rcnt;
-	off_t offs;
+	off_t offs = offset;
 	unsigned int status;
 	int err;
 
@@ -726,8 +726,16 @@ ssize_t posix_read(int fildes, void *buf, size_t nbyte)
 		return -EBADF;
 	}
 
+	if (offset >= 0 && !F_SEEKABLE(f->type)) {
+		posix_fileDeref(f);
+		return -ESPIPE;
+	}
+
 	proc_lockSet(&f->lock);
-	offs = f->offset;
+	/* offset < 0 means use current fd offset */
+	if (offset < 0) {
+		offs = f->offset;
+	}
 	status = f->status;
 	proc_lockClear(&f->lock);
 
@@ -738,7 +746,7 @@ ssize_t posix_read(int fildes, void *buf, size_t nbyte)
 		rcnt = proc_read(f->oid, offs, buf, nbyte, status);
 	}
 
-	if (rcnt > 0) {
+	if (rcnt > 0 && offset < 0) {
 		proc_lockSet(&f->lock);
 		f->offset += rcnt;
 		proc_lockClear(&f->lock);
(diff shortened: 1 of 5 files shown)


syscalls.c
```diff
@@ -1124,6 +1124,56 @@ ssize_t syscalls_sys_write(char *ustack)
...
+	GETFROMSTACK(ustack, int, fildes, 0);
+	GETFROMSTACK(ustack, void *, buf, 1);
+	GETFROMSTACK(ustack, size_t, nbyte, 2);
+	GETFROMSTACK(ustack, off_t, offset, 3);
+
+	if ((buf == NULL) && (nbyte != 0)) {
+		return -EFAULT;
+	}
+
+	if ((buf != NULL) && (nbyte != 0) && (vm_mapBelongs(proc, (char *)buf + offset, nbyte) < 0)) {
```
Darchiv: Why check for `buf + offset`? `offset` only applies to a seekable file and not the user-provided buffer. ALso applies to `syscalls_sys_pwrite()`

posix/posix.c
```diff
@@ -794,6 +794,78 @@ ssize_t posix_write(int fildes, void *buf, size_t nbyte)
 }
 
 
+ssize_t posix_pread(int fildes, void *buf, size_t nbyte, off_t offset)
```
Darchiv: `posix_pread()` and `posix_pwrite()` functions look very alike. We don't want to duplicate code (hurts maintainability and increases kernel's image size). I think you can commonize these functions by adding an `off_t offset` parameter and changing the base function's login inside to use it.

Fortunately, POSIX mandates `off_t` to be a signed type and `pread`/`pwrite` functions reject negative `offset` values. This way, you can use -1 as a default value which means "use implicit (file descriptor's) offset". You would need to make `read`/`write` (in libphoenix) wrappers which pass -1 to `sys_read`/`sys_write`. The `read`/`write` syscall will be changed (by adding `offset`), which makes this PR a breaking, inter-dependent change (both kernel and libphoenix must be merged in order for the system to work) - please create a PR in phoenix-rtos-project which moves kernel, libphoenix, and (hopefully) tests submodules (commit name: *update submodules*) - this way we can see if the combined change builds and passes CI. Foe example, see: https://github.com/phoenix-rtos/phoenix-rtos-project/pull/1317

Also, please add (or enable if they exist) test cases for `pread` and `pwrite` in phoenix-rtos-tests.

Darchiv (commented):
Can be commonized. See the comments.

Please add reviewers at the moment you make a PR ready for review.

syscalls.c
Darchiv: Gemini also told you about that, but you ignored it

posix/posix.c
```diff
@@ -706,13 +706,13 @@ int posix_close(int fildes)
 }
 
 
-ssize_t posix_read(int fildes, void *buf, size_t nbyte)
+ssize_t posix_read(int fildes, void *buf, size_t nbyte, off_t offset)
 {
-	TRACE("read(%d, %p, %u)", fildes, buf, nbyte);
+	TRACE("read(%d, %p, %u, %lld)", fildes, buf, nbyte, offset);
```
Darchiv: Actually, we should not assume the size of `off_t` type as it is architecture-specific, i.e. we shouldn't assume it is `long long int`. For the same reason we have `PRI*` macros in inttypes.h for fixed-width types - because we don't have to know whether `uint32_t` is implemented with `unsigned int` or `unsigned long int` etc.

For `off_t` I suggest casting to `intmax_t` and using `%jd` specifier, as there is no dedicated length sub-specifier for `off_t` like there is for `size_t` (`z`).

BTW: the format specifier for `nbyte` parameter is also wrong, as it misses the mentioned length sub-specifier - it should be `%zu`. You can fix it in this function if you want.

Also applies to `posix_write()`

posix/posix_private.h
```diff
@@ -73,6 +73,9 @@ enum { ftRegular, ftPipe, ftFifo, ftInetSocket, ftUnixSocket, ftTty };
 enum { pxBufferedPipe, pxPipe, pxPTY };
 
 
+#define F_SEEKABLE(type) (type == ftRegular)
```
Darchiv: `type` in macro body should also be parenthesized. See [SEI CERT Rec PRE01-C](https://wiki.sei.cmu.edu/confluence/display/c/PRE01-C.+Use+parentheses+within+macros+around+parameter+names) for rationale

PR #693, opened 2025-09-02
process: terminate threads on exec

etiaro:
## Description

Fixes a bug with multithreaded process call an exec leaving other threads running with new process memory map.

Fixes phoenix-rtos/phoenix-rtos-project#1393

JIRA: RTOS-1111

## Motivation and Context

Bug with kernel panic was detected when added alarm thread spawning in forked process to keep alarm working.

## Types of changes

## How Has This Been Tested?

- [x] New test added: phoenix-rtos/phoenix-rtos-tests#420

## Checklist:

## Special treatment

### DIFF
--- a/proc/process.c
+++ b/proc/process.c
@@ -234,7 +234,7 @@ int proc_start(void (*initthr)(void *), void *arg, const char *path)
 
 void proc_kill(process_t *proc)
 {
-	proc_threadsDestroy(&proc->threads);
+	proc_threadsDestroy(&proc->threads, NULL);
 }
 
 
@@ -1555,6 +1555,14 @@ static int process_execve(thread_t *current)
 
 	/* The old user stack is no longer valid */
 	current->ustack = NULL;
+
+	/* Terminate other threads */
+	proc_threadsDestroy(&current->process->threads, current);
+
+	/* Wait for other threads termination before changing memory maps */
+	while (proc_threadsOther(current) != 0) {
+		proc_join(-1, 0);
+	}
 
 	/* Restore kernel stack of parent thread */
 	if (parent != NULL) {
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -1002,16 +1002,18 @@ void proc_threadDestroy(thread_t *t)
 }
 
 
-void proc_threadsDestroy(thread_t **threads)
+void proc_threadsDestroy(thread_t **threads, const thread_t *except)
 {
 	thread_t *t;
 	spinlock_ctx_t sc;
 
 	hal_spinlockSet(&threads_common.spinlock, &sc);
 	if ((t = *threads) != NULL) {
-		do
-			_proc_threadExit(t);
-		while ((t = t->procnext) != *threads);
+		do {
+			if (t != except) {
+				_proc_threadExit(t);
+			}
+		} while ((t = t->procnext) != *threads);
 	}
 	hal_spinlockClear(&threads_common.spinlock, &sc);
 }
@@ -2092,6 +2094,20 @@ int proc_threadsList(int n, threadinfo_t *info)
 	proc_lockClear(&threads_common.lock);
 
 	return i;
+}
+
+
+int proc_threadsOther(thread_t *t)
+{
+	int ret;
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	/* Assumes t is not NULL and belongs to a process */
+	ret = (t->procnext != t) ? 1 : 0;
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+
+	return ret;
 }
 
 
--- a/proc/threads.h
+++ b/proc/threads.h
@@ -139,7 +139,7 @@ extern int proc_threadJoin(unsigned int id);
 extern void proc_threadDestroy(thread_t *t);
 
 
-extern void proc_threadsDestroy(thread_t **threads);
+extern void proc_threadsDestroy(thread_t **threads, const thread_t *except);
 
 
 extern int proc_waitpid(int pid, int *stat, int options);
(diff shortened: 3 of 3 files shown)


proc/process.c
```diff
@@ -1556,6 +1556,15 @@ static int process_execve(thread_t *current)
 	/* The old user stack is no longer valid */
 	current->ustack = NULL;
 
+	/* Terminate other threads */
+	proc_kill(current->process);
+	current->exit = 0;
```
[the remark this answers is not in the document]
etiaro: The scheduler does not ghostify threads in kernel space, so the code is fine.

proc/process.c
```diff
@@ -1556,6 +1556,14 @@ static int process_execve(thread_t *current)
 	/* The old user stack is no longer valid */
 	current->ustack = NULL;
 
+	/* Terminate other threads */
+	proc_threadsDestroyExcept(&current->process->threads, current);
+
+	/* Wait for other threads termination before changing memory maps */
+	while (current->procnext != current) {
```
nalajcie: I understand that we need to wait here - I'm not sure what would happen if one of the other threads is in uninterruptible sleep right now (eg. waits for msg response). 

Hopefully almost nobody tries to call `exec` in multithreaded programs and this won't introduce any deadlocks.

nalajcie (commented):
nice find, I'm a little bit afraid about potential deadlocks - but that should not happen in real-use scenarios

proc/process.c
etiaro: Yes, uninterruptible sleep will block the exec. I see a few options here:
1. Make uninterruptible threads processless - complicates logic and requires careful inspection of all returns from uninterruptible sleep 
2. Ignore uninterruptible threads here - can lead to unexpected interactions with exec'd process, complicates logic around exec
3. Introduce a way of killing uninterruptible threads - cleanest after done, but complicated to do and can introduce lots of other unexpected problems.

I would opt for 3rd one, but it's probably too big of a change to be done here.

nalajcie: 3rd option is probably a huge undertaking, for now we need to be just aware of the possible consequences :).

AFAIK we can't ignore threads / make them processless as those thread use process memory map and other resources.

Partial solution is to make more operations interruptible (eg. IO operations are uninterruptible right now)

etiaro: Yes, they would have to be gracefully killed soon after waking up and cleaning up.

proc/process.c
```diff
@@ -1556,6 +1556,14 @@ static int process_execve(thread_t *current)
...
+		proc_join(-1, 0);
+	}
```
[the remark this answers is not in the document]
etiaro: @gemini-code-assist we have the assumption of atomicity of pointer access.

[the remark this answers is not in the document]
etiaro: @gemini-code-assist The read itself is working without the lock. Looking at lib_listRemove shows that its value is
- NULL when element is removed from the list (not in this case), 
- old value that is being modified, but in that case we'll just wait in join for the next thread, or 
- the new modified value, which will happen only once all other threads died.
So it's not safe to modify or even explore the list without the lock, but this particular read is fine.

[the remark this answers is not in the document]
etiaro: The code first removes from the list, then adds ghost and wakes up the reaper. Above interleave can't happen.
That mentioned, proc_threadsOther function is added with spinlock to remove the assumptions about lib_list* functions simplifying further codebase maintenance.

proc/process.c
agkaminski, line 1563: Hmm, aside from the gemini rant - IMHO this condition is not ok. We can find `current->procnext != current` and it still won't be a last one:

list head: current
current -> thread1 -> thread2 -> current

proc/threads.c
```diff
@@ -2095,6 +2097,20 @@ int proc_threadsList(int n, threadinfo_t *info)
...
+{
+	int ret;
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	/* Assumes t is not NULL and belongs to a process */
+	ret = (t->procnext != t);
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+
+	return ret ? 1 : 0;
```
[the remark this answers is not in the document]
nalajcie: The coding style is MISRA, how to approach implicit cating here?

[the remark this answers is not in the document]
etiaro:
```C
	ret = (t->procnext != t) ? 1 : 0;
```
This way it should be MISRA compliant

proc/threads.c
```diff
@@ -1002,16 +1002,18 @@ void proc_threadDestroy(thread_t *t)
 }
 
 
-void proc_threadsDestroy(thread_t **threads)
+void proc_threadsDestroyExcept(thread_t **threads, thread_t *except)
```
Darchiv: I'm not sure the `Except` suffix in function's name is needed here. The second parameter's name already tells us what this function is supposed to do. Also, this suggests that a more generic form of this function (`proc_threadsDestroy()`) exists somewhere, which will no longer be true.

Darchiv: `thread_t *except` can be `const`

proc/threads.h
```diff
@@ -154,6 +154,9 @@ extern void proc_changeMap(process_t *proc, vm_map_t *map, vm_map_t *imap, pmap_
 extern int proc_threadsList(int n, threadinfo_t *info);
 
 
+extern int proc_threadsOther(thread_t *t);
```
Darchiv, line 157: Looking at the prototype alone, what does this function do?

I know not all `proc_*()` functions use verbs in names (like `proc_threadsList()`) and it can be hard to come up with a concise name for this function (`proc_threadsDoOtherThreadsExist()` doesn't cut it :D), especially since we avoid `bool` in p-r-kernel. This is a nitpick for addressing it in the future, if we will ever make kernel code more readable.

Darchiv (approved):

PR #701, opened 2025-09-26
hal/stm32n6: multiple fixes and preparation for DMA

jmaksymowicz:
## Description
Added `platformctl`s and configuration necessary to implement DMA (cache management, DMA permissions control, RISAF configuration).
Add option to disable the use of HSE oscillator in case it's not available on a given board.
Fix mistakes in functions for RIMC configuration.

## Motivation and Context
Allow DMA to be implemented on STM32N6 and run the system on boards without external oscillator.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv8m55-stm32n6-nucleo

## Checklist:

## Special treatment

### DIFF
--- a/hal/arm/scs.c
+++ b/hal/arm/scs.c
@@ -17,6 +17,9 @@
 #include <arch/cpu.h>
 #include "scs.h"
 
+#define CPUID_PARTNO_M7  0xc27UL
+#define CPUID_PARTNO_M55 0xd22UL
+
 
 struct scs_s {
 	u32 _res0[2];
@@ -211,8 +214,8 @@ static int _hal_scbCacheIsSupported(void)
 {
 	u32 partno = ((_hal_scsCpuID() >> 4) & 0xfff);
 
-	/* Only supported on Cortex-M7 for now */
-	if (partno == 0xc27) {
+	/* Only supported on Cortex-M7 and Cortex-M55 for now */
+	if ((partno == CPUID_PARTNO_M7) || (partno == CPUID_PARTNO_M55)) {
 		return 1;
 	}
 
@@ -281,7 +284,7 @@ void _hal_scsDCacheDisable(void)
 }
 
 
-void _hal_scsDCacheCleanInvalAddr(void *addr, u32 sz)
+static void _hal_scsDCacheOpAddr(void *addr, u32 sz, volatile u32 *reg)
 {
 	u32 daddr;
 	int dsize;
@@ -300,13 +303,31 @@ void _hal_scsDCacheCleanInvalAddr(void *addr, u32 sz)
 	hal_cpuDataSyncBarrier();
 
 	do {
-		scs_common.scs->dccimvac = daddr;
+		*reg = daddr;
 		daddr += 0x20u;
 		dsize -= 0x20;
 	} while (dsize > 0);
 
 	hal_cpuDataSyncBarrier();
 	hal_cpuInstrBarrier();
+}
+
+
+void _hal_scsDCacheCleanInvalAddr(void *addr, u32 sz)
+{
+	_hal_scsDCacheOpAddr(addr, sz, &scs_common.scs->dccimvac);
+}
+
+
+void _hal_scsDCacheCleanAddr(void *addr, u32 sz)
+{
+	_hal_scsDCacheOpAddr(addr, sz, &scs_common.scs->dccvac);
+}
+
+
+void _hal_scsDCacheInvalAddr(void *addr, u32 sz)
+{
+	_hal_scsDCacheOpAddr(addr, sz, &scs_common.scs->dcimvac);
 }
 
 
--- a/hal/armv8m/stm32/n6/risaf.c
+++ b/hal/armv8m/stm32/n6/risaf.c
@@ -1,0 +1,402 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * STM32N6 RISAF (Resource isolation slave unit for address space protection (full version)) configuration.
+ *
+ * Copyright 2025 Phoenix Systems
+ * Author: Jacek Maksymowicz
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "hal/armv8m/stm32/stm32.h"
+#include "hal/armv8m/stm32/n6/stm32n6_regs.h"
+#include "include/errno.h"
+#include "hal/cpu.h"
+
+
+#define RISAF1_BASE  ((void *)0x54026000)
+#define RISAF2_BASE  ((void *)0x54027000)
+#define RISAF3_BASE  ((void *)0x54028000)
+#define RISAF4_BASE  ((void *)0x54029000)
+#define RISAF5_BASE  ((void *)0x5402a000)
+#define RISAF6_BASE  ((void *)0x5402b000)
+#define RISAF7_BASE  ((void *)0x5402c000)
+#define RISAF8_BASE  ((void *)0x5402d000)
+#define RISAF9_BASE  ((void *)0x5402e000)
+#define RISAF11_BASE ((void *)0x54030000)
+#define RISAF12_BASE ((void *)0x54031000)
+#define RISAF13_BASE ((void *)0x54032000)
+#define RISAF14_BASE ((void *)0x54033000)
+#define RISAF15_BASE ((void *)0x54034000)
+#define RISAF21_BASE ((void *)0x54035000)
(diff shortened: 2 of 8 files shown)


hal/armv8m/stm32/n6/stm32n6.c
```diff
@@ -208,24 +244,74 @@ int _stm32_rifsc_risup_change(unsigned int index, int secure, int privileged, in
...
+	}
+
+	if (privileged > 0) {
+		*(stm32_common.rifsc + rifsc_rimc_attr0 + index) |= (1 << 9);
+	}
+	else if (privileged < 0) {
+		*(stm32_common.rifsc + rifsc_rimc_attr0 + index) &= ~(1 << 9);
+	}
+
+	if ((cid >= 0) && (cid < 0x7)) {
```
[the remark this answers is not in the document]
jmaksymowicz: CID == 0x7 is invalid:
> This bitfield cannot be written with a value of 0x7 (write to MCID[2:0] is ignored).

include/arch/armv8m/stm32/n6/stm32n6.h
```diff
@@ -607,6 +619,10 @@ typedef struct {
 			unsigned int magic;
 			unsigned int reason;
 		} reboot;
+		struct {
+			void *addr;
+			unsigned int sz;
+		} cleanInvalDCache;
```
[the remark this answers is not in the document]
jmaksymowicz: I used this name to be consistent with APIs on `imxrt10xx` and `imxrt1170`.

hal/armv8m/stm32/n6/risaf.c
```diff
@@ -0,0 +1,381 @@
...
+	},
+	[risaf_cpu_mst] = {
+		.base = RISAF6_BASE,
+		.start = 0x0,
+		.end = 0xffffffff,
+		.granularity = (1 << 12) - 1,
+		.n_regions = 11,
+		.isCIDAware = 1,
+	},
+	[risaf_flexram] {
```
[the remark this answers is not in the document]
jmaksymowicz: There is a mistake but somehow it doesn't cause a compilation error.

hal/arm/scs.c
```diff
@@ -211,8 +211,8 @@ static int _hal_scbCacheIsSupported(void)
 {
 	u32 partno = ((_hal_scsCpuID() >> 4) & 0xfff);
 
-	/* Only supported on Cortex-M7 for now */
-	if (partno == 0xc27) {
+	/* Only supported on Cortex-M7 and Cortex-M55 for now */
+	if ((partno == 0xc27) || (partno == 0xd22)) {
```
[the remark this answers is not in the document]
jmaksymowicz: I could do that, but I think we need a better way to decide when to enable cache and when not to. If we just want to enable it whenever it's available, on ARMv8-M we can use the Cache Type Register.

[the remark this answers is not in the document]
agkaminski: You're both right, but this code is still common between v7 and v8, so no dice, we need a common denominator here

hal/armv8m/stm32/n6/risaf.c
```diff
@@ -0,0 +1,411 @@
...
+int _stm32_risaf_configRegion(
+		unsigned int risaf,
+		u8 region,
+		u32 start,
+		u32 end,
+		u8 privCIDMask,
+		u8 readCIDMask,
+		u8 writeCIDMask,
+		char secure,
+		char enable)
```
agkaminski: This style is not present in the kernel, perhaps one line break?

hal/armv8m/stm32/n6/risaf.c
```diff
@@ -0,0 +1,411 @@
...
+ */
+int _stm32_risaf_configRegion(
+		unsigned int risaf,
+		u8 region,
+		u32 start,
+		u32 end,
+		u8 privCIDMask,
+		u8 readCIDMask,
+		u8 writeCIDMask,
+		char secure,
```
agkaminski: MISRA - try to avoid using `char` for anything but characters and strings. `u8` is a best general purpose byte in most cases

hal/armv8m/stm32/n6/risaf.c
```diff
@@ -0,0 +1,411 @@
...
+	*(risafs[risaf].base + risaf_reg1_cidcfgr + region_offs) = tmp;
+	/* Values in registers are not CPU addresses, but offsets within the module's own address space. */
+	start -= risafs[risaf].start;
+	end -= risafs[risaf].start;
+	*(risafs[risaf].base + risaf_reg1_startr + region_offs) = start;
+	*(risafs[risaf].base + risaf_reg1_endr + region_offs) = end;
+	hal_cpuDataMemoryBarrier();
+
+	tmp = *(risafs[risaf].base + risaf_reg1_cfgr + region_offs);
+	if (secure) {
```
agkaminski: MISRA
```suggestion
	if (secure != 0) {
```

hal/armv8m/stm32/n6/risaf.c
```diff
@@ -0,0 +1,411 @@
...
+}
+
+int _stm32_risaf_getFirstDisabledRegion(int risaf)
+{
+	u32 region, region_offs;
+	if (risaf >= (sizeof(risafs) / sizeof(risafs[0]))) {
+		return -EINVAL;
+	}
+
+	for (region = 1; region <= risafs[risaf].n_regions; region++) {
```
agkaminski, line 332: Nit: better to start with `region = 0` because:
- is more common and thus less confusing,
- right now we need to subtract one on every iteration, otherwise we would only need to add one once ;)

hal/armv8m/stm32/n6/stm32n6.c
```diff
@@ -208,24 +244,74 @@ int _stm32_rifsc_risup_change(unsigned int index, int secure, int privileged, in
...
 
 	if (secure > 0) {
-		*(stm32_common.rifsc + rifsc_risc_seccfgr0 + index) |= (1 << 8);
+		*(stm32_common.rifsc + rifsc_rimc_attr0 + index) |= (1 << 8);
+	}
+	else if (secure < 0) {
+		*(stm32_common.rifsc + rifsc_rimc_attr0 + index) &= ~(1 << 8);
+	}
+
+	if (privileged > 0) {
```
agkaminski, line 261: Nit: up to now we used convention to set a param to a negative value if we want the related attribute not to change:
- 0 disable,
- 1 enable,
- -1 no change.

I don't say that a particular way is superior, but inconsistency in API is error prone

include/arch/armv8m/stm32/n6/stm32n6.h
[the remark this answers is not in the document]
agkaminski: It's ok as is, but striving for compatibility with platformctl of an another target is not needed, might be even detrimental:

Lets say that we mistakenly include a header of imxrt instead of stm32n6. Then we might still successfully compile a code using this field, but it won't work - APIs match, but ABIs do not (e.g. difference in value of `pctl_cleanDCache` enum.

Just a rant, it's ok as is

hal/armv8m/stm32/n6/stm32n6.c
jmaksymowicz: The problem is that I already used this convention (-1 disable, 0 no change, 1 enable) for other `platformctl`s on the N6, so if we switch the convention now we'd need to change `phoenix-rtos-devices` in sync with that.

hal/armv8m/stm32/n6/risaf.c
jmaksymowicz: In STM32N6 documentation the regions are numbered from 1 to 15 and refer to default behavior of the firewall as "region 0". I wanted to keep our numbering the same as in documentation, all of this is confusing enough as it is.

agkaminski: That I do get, I mean just the loop. But ok, ack

hal/armv8m/stm32/n6/risaf.c
jmaksymowicz: I checked the coding convention guide and it doesn't mention multi-line function declarations, so I changed it to a single line.

hal/arm/scs.c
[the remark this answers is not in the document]
jmaksymowicz: Added `#define`s for the different CPU part numbers.

hal/armv8m/stm32/n6/stm32n6.c
agkaminski: Ack, at least it is coherent within the targer

PR #720, opened 2025-12-05
!syscalls/threadsinfo: return more granular thread information

oI0ck:
## Description
This commit means to make `threadinfo` more flexible, by being able to query a thread identified by `tid` and query for specific thread attributes as specified in `flags` bitmask.

Depends-On: phoenix-rtos-utils#251
Depends-On: libphoenix#454

## Motivation and Context
`threadsinfo` is not very flexible and causes unnecessary overhead in programs using it.

Moreover, previously there was no was of inexpensively checking amount of present threads in the kernel, therefore the buffer passed to `threadsinfo` may have been often reallocated to check all threads. Now `threadsinfo`, if passed a `PH_THREADINFO_OPT_THREADCOUNT` flag will yield no information in passed buffer, but will return amount of currently present threads.

JIRA: RTOS-1187

## Types of changes

## How Has This Been Tested?
- [ ] New test added: (test PR is WIP).
- [X] Tested by hand on: `ia32-generic-qemu`, `riscv64-generic-qemu`, `sparcv8leon-generic-qemu`, `armv7a9-zynq7000-qemu`, `armv7r5f-zynqmp-qemu`.

## Checklist:

## Special treatment

- [X] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
- https://github.com/phoenix-rtos/libphoenix/pull/454
- https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/251
- https://github.com/phoenix-rtos/phoenix-rtos-doc/pull/247


proc/threads.c
```diff
@@ -2086,120 +2088,197 @@ void proc_threadsDump(u8 priority)
...
+	if (thread->process != NULL) {
+		info->pid = process_getPid(thread->process);
+		info->ppid = 0;
+	}
+	else {
+		info->pid = 0;
+		info->ppid = 0;
+	}
+
+	if ((flags & PH_THREADINFO_TID) == PH_THREADINFO_TID) {
```
adamgreloch: why not just 

```suggestion
	if ((flags & PH_THREADINFO_TID) != 0) {
```

include/sysinfo.h
```diff
@@ -17,6 +17,33 @@
...
+
+#define PH_THREADINFO_TID     (1U << 1U)
+#define PH_THREADINFO_PRIO    (1U << 2U)
+#define PH_THREADINFO_STATE   (1U << 3U)
+#define PH_THREADINFO_LOAD    (1U << 4U)
+#define PH_THREADINFO_CPUTIME (1U << 5U)
+#define PH_THREADINFO_WAITING (1U << 6U)
+#define PH_THREADINFO_NAME    (1U << 7U)
+#define PH_THREADINFO_VMEM    (1U << 8U)
+#define PH_THREADINFO_PPID    (1U << 9U)
```
adamgreloch: MISRA considers `1U` to be essentially char type (yuuup...), meaning that `1U << 9` is a violation. These have to be `1UL`

include/sysinfo.h
```diff
@@ -17,6 +17,33 @@
...
+		PH_THREADINFO_PRIO | \
+		PH_THREADINFO_STATE | \
+		PH_THREADINFO_LOAD | \
+		PH_THREADINFO_CPUTIME | \
+		PH_THREADINFO_WAITING | \
+		PH_THREADINFO_NAME | \
+		PH_THREADINFO_VMEM | \
+		PH_THREADINFO_PPID)
+
+#define PH_THREADINFO_OPT_THREADCOUNT (1U << 10U)
```
adamgreloch: Same as [above](https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/720/files#r2601980657)

proc/threads.c
```diff
@@ -2086,120 +2088,197 @@ void proc_threadsDump(u8 priority)
...
+				entry = entry->next;
+			} while (entry != thread->process->entries);
+		}
+	}
+#else /* !NOMMU */
 	vm_map_t *map;
-	time_t now;
-	char *name;
-	spinlock_ctx_t sc;
+	spinlock_ctx_t ctx;
```
adamgreloch: Maybe keep the spinlock context name as `sc` for consistency with the rest of threads.c?

proc/threads.c
```diff
@@ -2086,120 +2088,197 @@ void proc_threadsDump(u8 priority)
...
 
-		now = _proc_gettimeRaw();
-		if (now != t->startTime) {
-			info[i].load = (int)((t->cpuTime * 1000) / (now - t->startTime));
+	return vmem;
+}
+
+
+/* call with threads_common.lock set */
+void proc_threadInfo(thread_t *thread, unsigned int flags, threadinfo_t *info)
```
adamgreloch: If expected to be called with a lock, name it `_proc_threadInfo` (`_` implies need for external synchronization)

proc/threads.c
```diff
@@ -2086,120 +2088,197 @@ void proc_threadsDump(u8 priority)
...
+		hal_memcpy(name, thread->process->path, len);
+		name[len - 1] = '\0';
+	}
+	else if (thread->process == NULL) {
+		hal_memcpy(name, "[idle]", sizeof("[idle]"));
+	}
+}
+
+
+static inline int _proc_calculateVmem(thread_t *thread)
```
adamgreloch, line 2311: prefixed with `_` - does it require any external sync? there is a spinlock inside, so I guess not?

proc/threads.c
```diff
@@ -2086,120 +2088,197 @@ void proc_threadsDump(u8 priority)
 }
 
 
-int proc_threadsList(int n, threadinfo_t *info)
+static inline void _proc_makeProcessName(thread_t *thread, char *name, int sz)
```
adamgreloch: prefixed with `_` - does it require any external sync?

proc/threads.c
```diff
@@ -2086,120 +2088,197 @@ void proc_threadsDump(u8 priority)
...
+		if (t == NULL) {
+			(void)proc_lockClear(&threads_common.lock);
+			return -ENOENT;
 		}
 
-		++i;
-		t = lib_idtreeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
+		proc_threadInfo(t, flags, info);
+		i++;
 	}
```
adamgreloch, line 2457: maybe wrap it in `do { ... } while (0);` and just break instead of returns inside to always converge into same branch with lockClear?

proc/threads.c
```diff
@@ -2086,120 +2088,197 @@ void proc_threadsDump(u8 priority)
...
+
+	(void)proc_lockSet(&threads_common.lock);
+
+	if (tid == PH_THREADINFO_THREADS_ALL) {
+		t = lib_treeof(thread_t, idlinkage, lib_rbMinimum(threads_common.id.root));
+
+		if ((flags & PH_THREADINFO_OPT_THREADCOUNT) == PH_THREADINFO_OPT_THREADCOUNT) {
+			do {
+				i++;
+				t = lib_idtreeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
```
adamgreloch: I wonder if we shouldn't just maintain a global thread count instead

proc/threads.c
oI0ck: It uses a `thread_t *`, so we have to guarantee that it won't be deallocated, therefore it should be called with `threads_common.lock` set, as all the other functions here.

proc/threads.c
oI0ck: The difference is syntactical, the compiler should emit the same code in both cases, I guess I just felt like writing it like that :D. I will change it though, for the sake of shorter line length.

proc/threads.c
oI0ck: Might be, though I'd say that's good enough for the amount of changes here. We'd have to guarantee that this counter will be kept in sync with node count in tree, which I'd say is besides the point in this PR.

proc/threads.c
oI0ck: Right, that is pretty ugly. I've rewritten `PH_THREADINFO_THREADS_ALL` as a `do-while` loop and it's better.

proc/threads.c
oI0ck: As above

proc/threads.c
oI0ck: Corrected

proc/threads.c
oI0ck: Corrected

include/sysinfo.h
oI0ck: Oof. Fixed this one

include/sysinfo.h
oI0ck: ^

proc/threads.c
oI0ck: Corrected

adamgreloch: Sure it is syntactical, it was just a style consistency nit

proc/threads.c
adamgreloch: I would make it explicit in the comment then, as the use of `thread_t *` alone doesn't imply that - usually you just up the refcnt on a thread, guaranteeing that it won't get deallocated when you need it and `threadsinfo` is an exception.

proc/threads.c
adamgreloch: Oh, what I meant was that since all your if-else returns were not nested in a loop, you could just wrap everything in the `do { ... } while (0);` and replace any branch like
```c
if (....) {
		...
} else {
		t = lib_idtreeof(thread_t, idlinkage, lib_idtreeFind(&threads_common.id, tid));
		if (t == NULL) {
			(void)proc_lockClear(&threads_common.lock);
			return -ENOENT;
		}
}
(void)proc_lockClear(&threads_common.lock);
return i;
```
with
```c
do {
if (....) {
		...
} else {
		t = lib_idtreeof(thread_t, idlinkage, lib_idtreeFind(&threads_common.id, tid));
		if (t == NULL) {
			i = -ENOENT;
			break;
		}
}
} while (0);
(void)proc_lockClear(&threads_common.lock);
return i;
```
so that the code doesn't diverge inside the if-else. This way there is always one returning path that clears the `threads_common.lock` and you don't need to remember about in other branches.

proc/threads.c
adamgreloch: Yeah, you're right. Maybe add a TODO/REVISIT then for now?

proc/threads.c
oI0ck: Yes, I know what you mean, but it still looks a bit better, since in `PH_THREADINFO_THREADS_ALL` case, we will have to use the loop either way, so making two loops inside of two branches (when they do mostly the same thing) is kinda ugly and the way I rewrote it also addresses having two return points.

adamgreloch: There is still one return in `else`, that's what confused me, but if you think that forcibly reducing the divergence would clutter the code more than solve anything here, then it's fine by me

include/sysinfo.h
```diff
@@ -17,6 +17,33 @@
...
+		PH_THREADINFO_PRIO | \
+		PH_THREADINFO_STATE | \
+		PH_THREADINFO_LOAD | \
+		PH_THREADINFO_CPUTIME | \
+		PH_THREADINFO_WAITING | \
+		PH_THREADINFO_NAME | \
+		PH_THREADINFO_VMEM | \
+		PH_THREADINFO_PPID)
+
+#define PH_THREADINFO_OPT_THREADCOUNT (1UL << 10U)
```
julianuziemblo: Nitpick: second `U` not needed e.g.:
```c
#define PH_THREADINFO_CPUTIME (1UL << 5)
```

proc/threads.c
```diff
@@ -2086,120 +2088,189 @@ void proc_threadsDump(u8 priority)
 }
 
 
-int proc_threadsList(int n, threadinfo_t *info)
+static inline void _proc_makeProcessName(thread_t *thread, char *name, int sz)
 {
-	int i = 0, argc;
-	unsigned int len, space;
-	thread_t *t;
+	int i, len, left = sz;
```
julianuziemblo: I'd leave `left` as int (or maybe long?) and change type of `i` and `len` and `sz` to `size_t` or `unsigned int`, as `sz` is always passed from `sizeof`. 

Then of course remember to cast to int when assiging to `left` as for MISRA...

proc/threads.c
```diff
@@ -2086,120 +2088,189 @@ void proc_threadsDump(u8 priority)
...
+		*(name - 1) = '\0';
+	}
+	else if (thread->process != NULL && thread->process->path != NULL) {
+		len = min(sz, hal_strlen(thread->process->path) + 1U);
+		hal_memcpy(name, thread->process->path, len);
+		name[len - 1] = '\0';
+	}
+	else if (thread->process == NULL) {
+		hal_memcpy(name, "[idle]", sizeof("[idle]"));
+	}
```
julianuziemblo: MISRA: add empty `else` branch (yeah ik...)

proc/threads.c
```diff
@@ -2086,120 +2088,189 @@ void proc_threadsDump(u8 priority)
...
+{
+	int vmem = 0;
 	map_entry_t *entry;
+
+#ifdef NOMMU
+	if (thread->process != NULL) {
+		entry = thread->process->entries;
+		if (entry != NULL) {
+			do {
+				vmem += (int)entry->size;
```
julianuziemblo, line 2322: ik this would break the interface, but is there a reason `[threadinfo_t]::vmem` is int? if not - could it be changed to unsigned (or size_t) to avoid these awkward casts? 

After a quick look: this shouldn't break much, at least in the main repo

proc/threads.c
```diff
@@ -2086,120 +2088,189 @@ void proc_threadsDump(u8 priority)
...
-		else
-#endif
-				if (map != NULL) {
-			(void)proc_lockSet(&map->lock);
-			entry = lib_treeof(map_entry_t, linkage, lib_rbMinimum(map->tree.root));
+	}
+}
+
+
+int proc_threadsList(int tid, unsigned int flags, int n, threadinfo_t *info)
```
julianuziemblo: why is this function named `threadsList`? I thought it provided thread **info** list, and not a thread lists

julianuziemblo (commented):
MISRA nits (for a better (worse) evening)

proc/threads.c
adamgreloch: +1 - these changes are already breaking, so IMO this is a good opportunity to tidy up types in `threadinfo_t` and potentially reduce casts

julianuziemblo: so another thing if this is breaking: signal it with `!` in PR title and checkbox in description

proc/threads.c
oI0ck: I left the name as it previously was, but good nitpick, something like `proc_threadsInfo` is probably better

include/sysinfo.h
oI0ck: Done

proc/threads.c
oI0ck: Changed all of them to `size_t` as well as parameter `sz`, since we always pass `sizeof`

proc/threads.c
oI0ck: done

proc/threads.c
oI0ck: done

proc/threads.c
julianuziemblo: Now the test in line 2103 (`left <= 0`) is pointless because `left` cannot be < 0. You should change back the type as this probably introduces an error

include/sysinfo.h
julianuziemblo, line 27: `tid` seems to sometimes be signed, and sometimes unsigned. Broader question here - do you happen to know if there's a reason for this? Even `proc_getTid()` returns `int`. @adamgreloch

include/sysinfo.h
```diff
@@ -33,7 +60,7 @@ typedef struct _threadinfo_t {
 	time_t cpuTime;
 	int priority;
 	int state;
```
julianuziemblo, line 33: `state` and `priority` seem to be bitmasks too, could probably be changed to `unsigned int`

include/sysinfo.h
julianuziemblo, line 30: Is `load` a time parameter? if so - could maybe be changed to `time_t`

julianuziemblo (commented):
More comments to `threadsinfo_t` data structure. Some of its members seem to have improper types with no explanation and we could fix it while at it.

include/sysinfo.h
adamgreloch: Internally tid is an idtree key which is signed to support POSIX handles

adamgreloch: No, it is a percentage of time spent on cpu. Current master code:
```c
info[i].load = (int)((t->cpuTime * 1000) / (now - t->startTime));
```

oI0ck: A signed thread identifier makes little sense as a key in a tree structure. By convention, values such as PID are unique ID based on a counter, so making it signed is seemingly pointless, unless to signal a possibility of returning an error within the negative domain.

include/sysinfo.h
oI0ck: `priority` is an index into `thread_common.ready`, which is an array of arrays containing threads divided by their priority. Signedness of this value is more engraved in the system, as apparent by `proc_threadPriority` and the `priority` syscall, changing this seems unnecessary, at least as a part of this PR.

`state` is an enum with 3 possible values, `READY (0U)`, `SLEEP (1U)`, `GHOST (2U)`, so while, sure, we could change this type here, it is not only breaking, but it might be, that in the future, someone may want to introduce a state signalling an error as a negative value, which would the require changing it back to `int`, so further breaking changes.

I'd personally stray from making further breaking changes within this PR, as it will be harder to merge it, reason about implications of it, and so on and so forth. The fact that it is already breaking should not be a signal to break even more, without a proper discussion. Review comments are a place to talk through changes made in a specific PR, to probably make it better, not introduce more changes unrelated to the PR.

This should be talked through in-person or on our issue tracker, so if you are keen making those changes, please create a new ticket, I'll be glad to participate in the discussion.

proc/threads.c
oI0ck: eh, right, was tired when commiting that, i'm changing `left` to `long`

proc/threads.c
oI0ck: I'm dropping the `vmem` type change, correcting those type may have bigger implications and should be considered on its own.

include/sysinfo.h
julianuziemblo: Valid. From our perspective however, there isn't really ever better time for this discussions than these PRs. But it's hard to tell sometimes how much the discussion is connected with the topic. The recent MISRA changes also affect this mindset somewhat.
These points were brought up here because you're already changing the interface and fixing it in other subrepos, so I wouldn't agree these changes are hard to reason about as there are little users to change. 
But, if you think that this is too much of a breaking change, for me it can stay as is for now - the types here seem really weird though, that's what my comments are about.

include/sysinfo.h
adamgreloch: Yes, I agree. My (vague) comment's intention was to explain why `tid` sometimes happens to be an `int` and AFAIK it's purely a consequence of idtree implementation.

We should probably clean this up, make it consistently `unsigned int` across the kernel or even introduce `tid_t` to control the types even further (outside of this PR).

include/sysinfo.h
adamgreloch: @oI0ck makes a good point. I invite you both to RTOS-1196 :)

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
 
-		cb(arg, i, &tinfo);
+			t = lib_idtreeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
+			i++;
+		} while (t != NULL);
+	}
+	else {
+		t = lib_idtreeof(thread_t, idlinkage, lib_idtreeFind(&threads_common.id, tid));
+		if (t == NULL) {
+			(void)proc_lockClear(&threads_common.lock);
```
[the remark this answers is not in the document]
oI0ck: `t` cannot be null, as there should be at least one thread in the tree at all times.

include/sysinfo.h
```diff
@@ -20,15 +20,15 @@
...
-#define PH_THREADINFO_PPID    (1UL << 9)
+#define PH_THREADINFO_TID     ((unsigned int)1UL << 1)
+#define PH_THREADINFO_PRIO    ((unsigned int)1UL << 2)
+#define PH_THREADINFO_STATE   ((unsigned int)1UL << 3)
+#define PH_THREADINFO_LOAD    ((unsigned int)1UL << 4)
+#define PH_THREADINFO_CPUTIME ((unsigned int)1UL << 5)
+#define PH_THREADINFO_WAITING ((unsigned int)1UL << 6)
+#define PH_THREADINFO_NAME    ((unsigned int)1UL << 7)
+#define PH_THREADINFO_VMEM    ((unsigned int)1UL << 8)
+#define PH_THREADINFO_PPID    ((unsigned int)1UL << 9)
```
julianuziemblo: `PS-C-REC-058`: when shifting `<=7` bits, mark the `1` with `U` instead of `UL`. This eliminates the need for casting, but looks inconsistent. It's only a recommendation and not a rule though.

oI0ck: I think we can omit this recommendation for the sake of code clarity. It just seems unnecessary and inconsistent when defining enum-like defines wider than 8 bits.

perf/trace.c
```diff
@@ -181,7 +181,7 @@
 
 static void _emitThreadinfo(void)
 {
-	(void)proc_threadsIter(0xffff, _emitThreadsCb, NULL);
+	proc_threadsIter(PH_THREADINFO_TID | PH_THREADINFO_PRIO | PH_THREADINFO_NAME, _emitThreadsCb, NULL);
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
+	}
+
+	if ((flags & PH_THREADINFO_TID) != 0) {
+		info->tid = (unsigned int)proc_getTid(thread);
+	}
+
+	if ((flags & PH_THREADINFO_PRIO) != 0) {
+		info->priority = (int)thread->priorityBase;
+	}
+
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
+	if ((flags & PH_THREADINFO_LOAD) != 0) {
+		if (now != thread->startTime) {
+			info->load = (int)((thread->cpuTime * 1000) / (now - thread->startTime));
 		}
 		else {
-			tinfo.load = 0;
+			info->load = 0;
 		}
-		tinfo.cpuTime = t->cpuTime;
+	}
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
+	if ((flags & PH_THREADINFO_WAITING) != 0) {
+		if (thread->state == READY && thread->maxWait < now - thread->readyTime) {
+			info->wait = now - thread->readyTime;
 		}
 		else {
-			tinfo.wait = t->maxWait;
+			info->wait = thread->maxWait;
 		}
-		hal_spinlockClear(&threads_common.spinlock, &sc);
+	}
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
+	if ((flags & PH_THREADINFO_NAME) != 0) {
+		if (thread->process != NULL) {
+			process_getName(thread->process, info->name, sizeof(info->name));
 		}
 		else {
-			map = threads_common.kmap;
-			hal_memcpy(tinfo.name, "[idle]", sizeof("[idle]"));
+			hal_memcpy(info->name, "[idle]", sizeof("[idle]"));
 		}
+	}
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
+			info->ppid = ppid;
 		}
-		else
-#endif
-				if (map != NULL) {
-			(void)proc_lockSet(&map->lock);
-			entry = lib_treeof(map_entry_t, linkage, lib_rbMinimum(map->tree.root));
+	}
+}
+
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1971,6 +2042,30 @@
...
+	(void)proc_lockSet(&threads_common.lock);
+
+	flags |= PH_THREADINFO_TID;
+
+	t = lib_treeof(thread_t, idlinkage, lib_rbMinimum(threads_common.id.root));
+	do {
+		_proc_threadInfo(t, flags, &info);
+
+		if (cb != NULL) {
+			cb(arg, &info);
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
+
+/* call with threads_common.lock set */
+void _proc_threadInfo(thread_t *thread, unsigned int flags, threadinfo_t *info)
+{
+	time_t now;
+	pid_t ppid;
+
+	if (thread == NULL || info == NULL) {
+		return;
+	}
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
+
+	if ((flags & PH_THREADINFO_PRIO) != 0) {
+		info->priority = (int)thread->priorityBase;
+	}
+
+	if ((flags & PH_THREADINFO_STATE) != 0) {
+		info->state = (int)thread->state;
+	}
+
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.h
```diff
@@ -117,13 +117,13 @@ int proc_join(int tid, time_t timeout);
 void proc_changeMap(process_t *proc, vm_map_t *map, vm_map_t *imap, pmap_t *pmap);
 
 
-typedef void (*proc_threadsListCb_t)(void *arg, int i, threadinfo_t *info);
+typedef void (*proc_threadsListCb_t)(void *arg, threadinfo_t *info);
```
julianuziemblo: ik this is not a change from this PR, but `PS-C-REC-015` states that function types should be suffixed with `Fn_t`

proc/threads.c
```diff
@@ -1874,95 +1875,165 @@
...
-			if (entry != NULL) {
-				do {
-					tinfo.vmem += (int)entry->size;
-					entry = entry->next;
-				} while (entry != t->process->entries);
-			}
+	if ((flags & PH_THREADINFO_PPID) != 0) {
+		ppid = posix_getppid(info->pid);
+		if (ppid > 0) {
+			info->ppid = ppid;
```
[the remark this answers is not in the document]
oI0ck: Corrected

proc/threads.c
```diff
@@ -1881,95 +1882,165 @@ void proc_threadsDump(u8 priority)
...
+
+	now = _proc_gettimeRaw();
+	if (thread->process != NULL) {
+		info->pid = process_getPid(thread->process);
+		info->ppid = 0;
+	}
+	else {
+		info->pid = 0;
+		info->ppid = 0;
+	}
```
julianuziemblo: there was a TODO here for `ppid`:

> /* TODO: tinfo.ppid = t->process->parent != NULL ? t->process->parent->id : 0; */

I'd say it should still be here as this wasn't fixed

proc/threads.c
```diff
@@ -1881,95 +1882,165 @@ void proc_threadsDump(u8 priority)
...
+	pid_t ppid;
+
+	if (thread == NULL || info == NULL) {
+		return;
+	}
+
+	now = _proc_gettimeRaw();
+	if (thread->process != NULL) {
+		info->pid = process_getPid(thread->process);
+		info->ppid = 0;
```
julianuziemblo: Nit: set `info->ppid` to 0 outside of `if - else` as it's always set

proc/threads.h
julianuziemblo: Nit: new coding standard recommends `Fn_t` suffix for function types

julianuziemblo (dismissed):
nits, otherwise LGTM

proc/threads.c
adamgreloch: Please add a comment similar to one preceeding `_proc_threadInfo`

proc/threads.h
oI0ck: done

proc/threads.c
oI0ck: done

proc/threads.c
```diff
@@ -1907,97 +1908,165 @@ void proc_threadsDump(u8 priority)
...
+static inline int _proc_calculateVmem(thread_t *thread)
 {
-	int i = 0;
-	thread_t *t;
+	int vmem = 0;
 	map_entry_t *entry;
+
+#ifdef NOMMU
+	if (thread->process != NULL) {
+		entry = thread->process->entries;
```
etiaro, line 2160: I see that this is also the case on master, but process->entries are not synchronized here in any way (map.c takes process->lock and map-> lock for every operation on these).

oI0ck:
Whoops, closed by mistake. TIL that github will automatically close the PR if it has nothing to merge (makes sense), I'll open a new one.

There is a problem with this PR on zynq7000-zedboard that I can't seem to figure out, seemingly commenting out all of this PR's content will still cause zynq to hang during tests. Will reopen when I get rid of those issues.

etiaro:
You should be able to reopen it after pushing some commits on this branch

etiaro:
NIT: Typo in commit message:
`Moreover, previously there was no was`

ziemleszcz:
Please remove the jmaksymowicz commit from this PR.

proc/threads.c
```diff
@@ -1907,97 +1908,164 @@ void proc_threadsDump(u8 priority)
...
+				_proc_threadInfo(t, flags, &info[i]);
 			}
-		}
 
-		cb(arg, i, &tinfo);
+			t = lib_idtreeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
+			i++;
+		} while (t != NULL);
+	}
+	else {
```
ziemleszcz, line 2304: Missing `n >= 1` check. Note that `vm_mapBelongs()` check passes on `n==0`.

proc/threads.c
```diff
@@ -1907,97 +1908,164 @@ void proc_threadsDump(u8 priority)
...
+	pid_t ppid;
+	spinlock_ctx_t sc;
+
+	if (thread == NULL || info == NULL) {
+		return;
+	}
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	now = _proc_gettimeRaw();
+	hal_spinlockClear(&threads_common.spinlock, &sc);
```
ziemleszcz: Other fields should also be read under spinlock - check old `procThreadIter()`.

proc/threads.c
```diff
@@ -1907,97 +1908,164 @@ void proc_threadsDump(u8 priority)
...
-				/* No action required */
+
+				_proc_threadInfo(t, flags, &info[i]);
 			}
-		}
 
-		cb(arg, i, &tinfo);
+			t = lib_idtreeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
+			i++;
+		} while (t != NULL);
```
ziemleszcz: Change to `while(t != NULL) {}`.

proc/threads.c
```diff
@@ -2006,11 +2074,37 @@ int proc_threadsIter(int n, proc_threadsListCb_t cb, void *arg)
...
+	t = lib_treeof(thread_t, idlinkage, lib_rbMinimum(threads_common.id.root));
+	do {
+		_proc_threadInfo(t, flags, &info);
+
+		if (cb != NULL) {
+			cb(arg, &info);
+		}
+
+		t = lib_idtreeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
+	} while (t != NULL);
```
ziemleszcz: Change to while(t != NULL) {}.

proc/threads.c
oI0ck: We don't access the `vm_map_t` here so `process->lock` may do here?

proc/threads.c
```diff
@@ -1920,97 +1921,166 @@
...
+	if (thread->process != NULL) {
+		entry = thread->process->entries;
+		if (entry != NULL) {
+			do {
+				vmem += (int)entry->size;
+				entry = entry->next;
+			} while (entry != thread->process->entries);
+		}
+	}
+	proc_lockClear(&thread->process->lock);
```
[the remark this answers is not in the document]
oI0ck: resolved

proc/threads.c
```diff
@@ -1920,97 +1921,166 @@
...
-int proc_threadsIter(int n, proc_threadsListCb_t cb, void *arg)
+static inline int _proc_calculateVmem(thread_t *thread)
 {
-	int i = 0;
-	thread_t *t;
+	int vmem = 0;
 	map_entry_t *entry;
+
+#ifdef NOMMU
+	proc_lockSet(&thread->process->lock);
```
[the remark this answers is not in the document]
oI0ck: resolved

proc/threads.c
adamgreloch: Bump

proc/threads.c
adamgreloch: This seems unresolved too

oI0ck: The check was added in `syscalls.c`

oI0ck: Right, added the check here, 0 should be allowed for getting the thread count

proc/threads.c
```diff
@@ -1937,97 +1938,171 @@ void proc_threadsDump(u8 priority)
...
+				_proc_threadInfo(t, flags, &info[i]);
 			}
+
+			t = lib_idtreeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
+			i++;
+		}
+	}
+	else {
+		if (n < 1) {
+			return -EINVAL;
```
ziemleszcz, line 2307: Return with lock taken.

proc/threads.c
```diff
@@ -1937,97 +1938,171 @@ void proc_threadsDump(u8 priority)
...
-		if (t->process != NULL) {
-			entry = t->process->entries;
-			if (entry != NULL) {
-				do {
-					tinfo.vmem += (int)entry->size;
-					entry = entry->next;
-				} while (entry != t->process->entries);
-			}
+	if ((flags & PH_THREADINFO_PPID) != 0U) {
+		ppid = posix_getppid(info->pid);
```
ziemleszcz, line 2272: `posix_gettpid()` takes `posix_common.lock`. Since this entire function runs with `threads_common.lock` already taken, we end up taking two global locks. This should be avoided - see RTOS-1095 on YT. The original implementation took these two locks separately.

ziemleszcz: Calling `posix_gettpid()` twice now.

proc/threads.c
```diff
@@ -1937,110 +1938,225 @@ void proc_threadsDump(u8 priority)
...
-		++i;
-		t = lib_idtreeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
+		_proc_threadInfo(t, flags, info);
+		i++;
 	}
 
 	(void)proc_lockClear(&threads_common.lock);
 
+	if ((flags & PH_THREADINFO_PPID) != 0U) {
+		for (int i = 0; i < n; i++) {
```
ziemleszcz: Shadowing first `i` variable (return count). Declaring variable in this place is forbidded in the kernel.

ziemleszcz:
`lib_idtreeof()` is no longer available. Please update this PR.

proc/threads.c
```diff
@@ -2221,110 +2222,217 @@ void proc_threadsDump(priority_t priority)
...
-		++i;
-		t = lib_treeof(thread_t, idlinkage, lib_idtreeNext(&t->idlinkage.linkage));
+		_proc_threadInfo(t, flags, info);
+		i++;
 	}
 
 	(void)proc_lockClear(&threads_common.lock);
 
+	if ((flags & PH_THREADINFO_PPID) != 0U) {
+		for (j = 0; j < n; j++) {
```
adamgreloch:
```suggestion
	if (((flags & PH_THREADINFO_PPID) != 0U) && ((flags & PH_THREADINFO_OPT_THREADCOUNT) == 0U)) {
		for (j = 0; j < i; j++) {
```
1. bound on number of threads existing, not by the caller-provided n
2. skip the iteration on PH_THREADINFO_OPT_THREADCOUNT

proc/threads.c
```diff
@@ -2221,110 +2222,217 @@ void proc_threadsDump(priority_t priority)
...
-		if (t->process != NULL) {
-			tinfo.pid = process_getPid(t->process);
-			// tinfo.ppid = t->process->parent != NULL ? t->process->parent->id : 0;
-			/* TODO: tinfo.ppid = t->process->parent != NULL ? t->process->parent->id : 0; */
-			tinfo.ppid = 0;
-		}
-		else {
-			tinfo.pid = 0;
-			tinfo.ppid = 0;
+	hal_spinlockClear(&threads_common.spinlock, &sc);
```
adamgreloch: I don't think this spinlock helps in any way. `proc_changeMap()` can swap the map after it gets cleared.
@etiaro?

syscalls.c
```diff
@@ -526,30 +526,23 @@ int syscalls_schedSet(u8 *ustack)
...
 	threadinfo_t *info;
 
-	GETFROMSTACK(ustack, int, n, 0U);
-	GETFROMSTACK(ustack, threadinfo_t *, info, 1U);
+	GETFROMSTACK(ustack, int, tid, 0U);
+	GETFROMSTACK(ustack, unsigned int, flags, 1U);
+	GETFROMSTACK(ustack, int, n, 2U);
+	GETFROMSTACK(ustack, threadinfo_t *, info, 3U);
 
 	if (vm_mapBelongs(proc, info, sizeof(*info) * (size_t)n) < 0) {
```
adamgreloch: Validate that `n >= 0` prior to size_t cast?

oI0ck: done

proc/threads.c
oI0ck: done

proc/threads.c
oI0ck: I guess if there would be `proc_changeMap()` on scheduled on other core, and here the spinlock was not set, the `map` read would not be atomic.

adamgreloch: Sure, but then you iterate over the old map. Ok, let's keep it as-is, but add a note/FIXME that this can be an outdated map

etiaro: I don't think this spinlock does anything here, I think we already make assumption of atomicity of word reads/writes in kernel.

Even if it made a difference, there is still no reason to use it around threads_common.kmap read.
This spinlock is heavily over-used, we should try to decrease its usage wherever its possible, not the other way around.

proc/threads.c
```diff
@@ -69,10 +69,11 @@ static struct {
 	intr_handler_t pendsvHandler;
 #endif
 
+	/* Synchronized by spinlock */
 	thread_t *ghosts;
 	thread_t *reaper;
 
-	/* Debug */
+	/* Debug, synchronized by spinlock */
```
etiaro: How is stackCanary synchronized by spinlock?

proc/threads.c
```diff
@@ -2221,110 +2222,218 @@ void proc_threadsDump(priority_t priority)
...
+		info->pid = process_getPid(thread->process);
+	}
+	else {
+		info->pid = 0;
+	}
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	now = _proc_gettimeRaw();
+
+	if ((flags & PH_THREADINFO_TID) != 0U) {
```
etiaro: Why do we need this level of granularity? 
Does setting to 0 of these simple flags like `PH_THREADINFO_TID` `PH_THREADINFO_PRIO` etc. increase speed in any measurable way?

Not a change request, just a question to get better image of this PRs goals.

adamgreloch: Could be that the current granularity in the case of simple things like `tid`, `priority`, `state`, etc. which are just reads from `thread_t` have actually worse performance due to branching than just always assigning them

adamgreloch: Maybe grouping the trivial thread metadata under one flag would be enough (up to you guys, not a blocker for me though at this point)

proc/threads.c
oI0ck: no idea honestly, wrote that a year ago almost :D

syscalls.c
```diff
@@ -526,30 +526,27 @@ int syscalls_schedSet(u8 *ustack)
...
 	}
 
-	n = proc_threadsList(n, info);
-
-	for (i = 0; i < n; ++i) {
-		ppid = posix_getppid(info[i].pid);
-		if (ppid > 0) {
-			info[i].ppid = ppid;
-		}
+	if (vm_mapBelongs(proc, info, sizeof(*info) * (size_t)n) < 0) {
```
ziemleszcz: Pre-exisiting issue: check that `sizeof(*info) * (size_t)n` doesn't overflow.

proc/threads.c
oI0ck: Yeah sure, I wanted to model this interface similarly to Linux's `statx()`, but stat returns a bit more info (querying the filesystem and so on), so that level of granularity might make sense there?

I think we can leave flags for specific parameters and group them into something like `PH_THREADINFO_BASIC`, so querying one flag from this group will just branch into taking all the basic info.

proc/threads.c
oI0ck: removed the comment

proc/threads.c
oI0ck: Moved spinlock to the `thread->process != NULL` branch

proc/threads.c
```diff
@@ -2221,110 +2222,206 @@ void proc_threadsDump(priority_t priority)
...
-	time_t now;
 	spinlock_ctx_t sc;
-	threadinfo_t tinfo;
 
-	(void)proc_lockSet(&threads_common.lock);
+	/* @FIXME(michal.lach): the map can be changed after clearing the spinlock */
 
-	t = lib_treeof(thread_t, idlinkage, lib_rbMinimum(threads_common.id.root));
+	if (thread->process != NULL) {
+		hal_spinlockSet(&threads_common.spinlock, &sc);
```
ziemleszcz: This spinlock is doing nothing.

syscalls.c
ziemleszcz: Interesting builtin. I would return EINVAL here.

include/sysinfo.h
```diff
@@ -15,6 +15,35 @@
 #define _PH_SYSINFO_H_
 
 
+/* threadsinfo attributes */
+#define PH_THREADINFO_THREADS_ALL (-1)
+
+#define PH_THREADINFO_TID     ((unsigned int)1UL << 1)
+#define PH_THREADINFO_PRIO    ((unsigned int)1UL << 2)
+#define PH_THREADINFO_STATE   ((unsigned int)1UL << 3)
+#define PH_THREADINFO_LOAD    ((unsigned int)1UL << 4)
```
ziemleszcz, line 24: `PH_THREADINFO_LOAD` is unused. Probably should be in `PH_THREADINFO_BASIC`.

syscalls.c
```diff
@@ -526,30 +526,34 @@ int syscalls_schedSet(u8 *ustack)
...
-	GETFROMSTACK(ustack, int, n, 0U);
-	GETFROMSTACK(ustack, threadinfo_t *, info, 1U);
+	GETFROMSTACK(ustack, int, tid, 0U);
+	GETFROMSTACK(ustack, unsigned int, flags, 1U);
+	GETFROMSTACK(ustack, int, n, 2U);
+	GETFROMSTACK(ustack, threadinfo_t *, info, 3U);
 
-	if (vm_mapBelongs(proc, info, sizeof(*info) * (size_t)n) < 0) {
-		return -EFAULT;
+	if (n < 0) {
```
ziemleszcz, line 543: According to the doc `n` and `info` should be ignored when `PH_THREADINFO_OPT_THREADCOUNT` is set in flags. Currently it's not true.

Also currently the flags is valid only when `tid == PH_THREADINFO_THREADS_ALL`. Please update the doc to make it clear.

proc/threads.h
```diff
@@ -137,13 +137,13 @@ int proc_join(int tid, time_t timeout);
 void proc_changeMap(process_t *proc, vm_map_t *map, vm_map_t *imap, pmap_t *pmap);
 
 
-typedef void (*proc_threadsListCb_t)(void *arg, int i, threadinfo_t *info);
+typedef void (*proc_threadsListFn_t)(void *arg, threadinfo_t *info);
 
 
-int proc_threadsIter(int n, proc_threadsListCb_t cb, void *arg);
+void proc_threadsIter(unsigned flags, proc_threadsListFn_t cb, void *arg);
```
ziemleszcz, line 143: Nitpick: `unsigned` -> `unsigned int`.

syscalls.c
```diff
@@ -526,30 +526,34 @@ int syscalls_schedSet(u8 *ustack)
  */
 
 
-int syscalls_threadsinfo(u8 *ustack)
+int syscalls_sys_threadsinfo(u8 *ustack)
 {
 	process_t *proc = proc_current()->process;
-	int n, i;
-	pid_t ppid;
+	_Bool overflow;
```
ziemleszcz, line 532: It seems that we don't use `bool` yet in the kernel. Use `int` for consistency.

proc/threads.c
```diff
@@ -2221,110 +2222,202 @@ void proc_threadsDump(priority_t priority)
...
+}
+
+
+/* call with threads_common.lock set */
+static void _proc_threadInfo(thread_t *thread, unsigned int flags, threadinfo_t *info)
+{
+	time_t now;
+	spinlock_ctx_t sc;
+
+	if (thread == NULL || info == NULL) {
```
ziemleszcz, line 2277: Nitpick: `((thread == NULL) || (info == NULL))`

proc/threads.c
```diff
@@ -2221,110 +2222,202 @@ void proc_threadsDump(priority_t priority)
...
-				(void)proc_lockSet(&map->lock);
-				entry = lib_treeof(map_entry_t, linkage, lib_rbMinimum(map->tree.root));
-
-				while (entry != NULL) {
-					tinfo.vmem += (int)entry->size;
-					entry = lib_treeof(map_entry_t, linkage, lib_rbNext(&entry->linkage));
+
+int proc_threadsInfo(int tid, unsigned int flags, int n, threadinfo_t *info)
+{
+	int i = 0, j = 0;
```
ziemleszcz, line 2333: No need to intialize `j`. `i` should be renamed to `cnt` or `n`.

include/sysinfo.h
```diff
@@ -15,6 +15,35 @@
 #define _PH_SYSINFO_H_
 
 
+/* threadsinfo attributes */
+#define PH_THREADINFO_THREADS_ALL (-1)
+
+#define PH_THREADINFO_TID     ((unsigned int)1UL << 1)
```
ziemleszcz, line 21: Why is bit 0 unused?

syscalls.c
oI0ck: It's hard to keep the docs PR up to date with changes to the code here.

The code PR's are a source-of-truth to the documentation, I'll update it after the code gets merged.

syscalls.c
oI0ck: `bool` is not in the kernel, because it is a define and we don't use standard headers.
Neither do we have any functions using `bool` as a variable type or a return type.

`_Bool` on the other hand, is a keyword, and the overflow multiplication builtin explicitly returns it.
Wouldn't it  be better to keep it as it is instead of performing a conversion to `int`?

oI0ck: Changed it to `int`

include/sysinfo.h
oI0ck: Made `_TID` use bit 0, shifted rest of the options accordingly

proc/threads.c
oI0ck: done

proc/threads.c
oI0ck: done

syscalls.c
```diff
@@ -526,30 +526,33 @@
...
+	GETFROMSTACK(ustack, threadinfo_t *, info, 3U);
 
-	if (vm_mapBelongs(proc, info, sizeof(*info) * (size_t)n) < 0) {
-		return -EFAULT;
+	if (n < 0) {
+		return -EINVAL;
 	}
 
-	n = proc_threadsList(n, info);
+	overflow = __builtin_mul_overflow(sizeof(*info), (size_t)n, &infosz);
```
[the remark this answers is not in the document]
ziemleszcz: Just add an explicit cast to int.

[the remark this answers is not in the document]
oI0ck: inlined the call in the if statement

syscalls.c
ziemleszcz:
```
if ((n > 0) && (info != NULL)) {
        if (__builtin_mul_overflow(sizeof(*info), (size_t)n, &infosz)) {
                return -EOVERFLOW;
        }

        if (vm_mapBelongs(proc, info, infosz) < 0) {
                return -EFAULT;
        }
}
```

oI0ck: done

syscalls.c
```diff
@@ -526,30 +526,32 @@ int syscalls_schedSet(u8 *ustack)
...
+	}
 
-	for (i = 0; i < n; ++i) {
-		ppid = posix_getppid(info[i].pid);
-		if (ppid > 0) {
-			info[i].ppid = ppid;
-		}
+	if (vm_mapBelongs(proc, info, infosz) < 0) {
+		return -EFAULT;
 	}
```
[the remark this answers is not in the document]
oI0ck: done

syscalls.c
```diff
@@ -526,30 +526,34 @@ int syscalls_schedSet(u8 *ustack)
...
+	if ((n > 0) && (info != NULL)) {
+		if (__builtin_mul_overflow(sizeof(*info), (size_t)n, &infosz)) {
+			return -EOVERFLOW;
+		}
 
-	for (i = 0; i < n; ++i) {
-		ppid = posix_getppid(info[i].pid);
-		if (ppid > 0) {
-			info[i].ppid = ppid;
+		if ((n > 0) && (vm_mapBelongs(proc, info, infosz) < 0)) {
```
ziemleszcz: `(n > 0)` must be true.

proc/threads.c
```diff
@@ -2306,102 +2307,192 @@ void proc_threadsDump(priority_t priority)
...
+
+		t = lib_treeof(thread_t, idlinkage, lib_idtreeFind(&threads_common.id, tid));
+		if (t == NULL) {
+			(void)proc_lockClear(&threads_common.lock);
+			return -ENOENT;
+		}
+
+		_proc_threadInfo(t, flags, info);
+		count++;
+	}
```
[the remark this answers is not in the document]
ziemleszcz: Accepting this ugly API for now.

proc/threads.c
```diff
@@ -2306,102 +2307,192 @@ void proc_threadsDump(priority_t priority)
...
+			entry = lib_treeof(map_entry_t, linkage, lib_rbNext(&entry->linkage));
 		}
 
-		hal_spinlockSet(&threads_common.spinlock, &sc);
-		tinfo.tid = (unsigned int)proc_getTid(t);
-		tinfo.priority = (int)t->priorityBase;
-		tinfo.state = (int)t->state;
+		(void)proc_lockClear(&map->lock);
+	}
+#endif
```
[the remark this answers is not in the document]
oI0ck: No, there is no regression, this is the current master code. What are you on about:
```c
#ifdef NOMMU
		if (t->process != NULL) {
			entry = t->process->entries;
			if (entry != NULL) {
				do {
					tinfo.vmem += (int)entry->size;
					entry = entry->next;
				} while (entry != t->process->entries);
			}
		}
		else
#endif
		{
			if (map != NULL) {
				(void)proc_lockSet(&map->lock);
				entry = lib_treeof(map_entry_t, linkage, lib_rbMinimum(map->tree.root));

				while (entry != NULL) {
					tinfo.vmem += (int)entry->size;
					entry = lib_treeof(map_entry_t, linkage, lib_rbNext(&entry->linkage));
				}
				(void)proc_lockClear(&map->lock);
			}
			else {
				/* No action required */
			}
		}
```

proc/threads.c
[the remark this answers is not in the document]
oI0ck: I mean, this requires a small change to work and the suggestion is pretty good:
```diff
diff --git a/proc/threads.c b/proc/threads.c
index 962173ca..708cc89e 100644
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -2438,18 +2438,21 @@ int proc_threadsInfo(int tid, unsigned int flags, int n, threadinfo_t *info)
                }
        }
        else {
-               if (n < 1) {
-                       (void)proc_lockClear(&threads_common.lock);
-                       return -EINVAL;
-               }
-
                t = lib_treeof(thread_t, idlinkage, lib_idtreeFind(&threads_common.id, tid));
                if (t == NULL) {
                        (void)proc_lockClear(&threads_common.lock);
                        return -ENOENT;
                }
 
-               _proc_threadInfo(t, flags, info);
+               if ((flags & PH_THREADINFO_OPT_THREADCOUNT) == 0U) {
+                       if (n < 1) {
+                               (void)proc_lockClear(&threads_common.lock);
+                               return -EINVAL;
+                       }
+
+                       _proc_threadInfo(t, flags, info);
+               }
+
                count++;
        }
 
```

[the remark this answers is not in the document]
oI0ck: applied

syscalls.c
```diff
@@ -526,30 +526,34 @@ int syscalls_schedSet(u8 *ustack)
...
+		}
 
-	for (i = 0; i < n; ++i) {
-		ppid = posix_getppid(info[i].pid);
-		if (ppid > 0) {
-			info[i].ppid = ppid;
+		if (vm_mapBelongs(proc, info, infosz) < 0) {
+			return -EFAULT;
 		}
 	}
```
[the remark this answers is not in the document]
oI0ck: fixd

ziemleszcz (approved):

PR #749, opened 2026-03-13
vm: improve error handling

etiaro:
## Description

Changes aiming at increasing system reliability when running out of memory.
After applying this changes, running a fork bomb did not cause kernel panics anymore.

## Motivation and Context

This is part of series of PRs that increase kernel stability when system is out of memory, inspired by work on reliability of separation given by partitioning mechanisms and related to fork bomb issue 
phoenix-rtos/phoenix-rtos-project#560

Also related to phoenix-rtos/phoenix-rtos-project#348

JIRA: RTOS-1235

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ia32-generic-qemu, aarch64a53-zynqmp-qemu

## Checklist:

## Special treatment

### DIFF
--- a/proc/process.c
+++ b/proc/process.c
@@ -1503,6 +1503,13 @@ int proc_vfork(void)
 		(void)vm_objectPut(spawn->object);
 		ret = spawn->state;
 		vm_kfree(spawn);
+		if ((ret < 0) && (posix_getppid(pid) == process_getPid(current->process))) {
+			/*
+			 * if the child managed to register itself within posix subsystem before
+			 * failure, wait for its complete death and cleanup its posix metadata.
+			 */
+			(void)posix_waitpid(pid, NULL, 0);
+		}
 		return (ret < 0) ? ret : pid;
 	}
 
--- a/proc/process.c
+++ b/proc/process.c
@@ -1178,23 +1178,30 @@ static void process_exec(thread_t *current, process_spawn_t *spawn)
 }
 
 
+__attribute__((noreturn)) static void proc_spawnThreadEnd(process_spawn_t *spawn, int state)
+{
+	spinlock_ctx_t sc;
+
+	hal_spinlockSet(&spawn->sl, &sc);
+	spawn->state = state;
+	(void)proc_threadWakeup(&spawn->wq);
+	hal_spinlockClear(&spawn->sl, &sc);
+
+	proc_threadEnd();
+}
+
+
 static void proc_spawnThread(void *arg)
 {
 	thread_t *current = proc_current();
 	process_spawn_t *spawn = arg;
-	spinlock_ctx_t sc;
 	int ret;
 
 	/* temporary: create new posix process */
 	if (spawn->parent != NULL) {
 		ret = posix_clone(process_getPid(spawn->parent->process));
 		if (ret < 0) {
-			hal_spinlockSet(&spawn->sl, &sc);
-			spawn->state = ret;
-			(void)proc_threadWakeup(&spawn->wq);
-			hal_spinlockClear(&spawn->sl, &sc);
-
-			proc_threadEnd();
+			proc_spawnThreadEnd(spawn, ret);
 		}
 		current->process->posix = 1U;
 	}
@@ -1337,29 +1344,25 @@ static void process_restoreParentKstack(thread_t *current, thread_t *parent)
 }
 
 
-static void proc_vforkedExit(thread_t *current, process_spawn_t *spawn, int state)
+__attribute__((noreturn)) static void proc_vforkedExit(thread_t *current, process_spawn_t *spawn, int state)
 {
-	spinlock_ctx_t sc;
-
 	current->ustack = NULL;
 	proc_changeMap(current->process, NULL, NULL, NULL);
+
+	proc_kill(current->process);
 
 	/* Only possible in the case of `initthread` exit or failure to fork. */
 	if (spawn->parent == NULL) {
 		hal_spinlockDestroy(&spawn->sl);
 		(void)vm_objectPut(spawn->object);
+
+		proc_threadEnd();
 	}
 	else {
 		process_restoreParentKstack(current, spawn->parent);
 
-		hal_spinlockSet(&spawn->sl, &sc);
-		spawn->state = state;
-		(void)proc_threadWakeup(&spawn->wq);
-		hal_spinlockClear(&spawn->sl, &sc);
+		proc_spawnThreadEnd(spawn, state);
 	}
-
-	proc_kill(current->process);
-	proc_threadEnd();
 }
 
 
@@ -1400,12 +1403,7 @@ static void process_vforkThread(void *arg)
 	parent = spawn->parent;
 	ret = posix_clone(process_getPid(spawn->parent->process));
 	if (ret < 0) {
-		hal_spinlockSet(&spawn->sl, &sc);
-		spawn->state = ret;
-		(void)proc_threadWakeup(&spawn->wq);
-		hal_spinlockClear(&spawn->sl, &sc);
-
-		proc_threadEnd();
+		proc_spawnThreadEnd(spawn, ret);
 	}
 	current->process->posix = 1U;
 
@@ -1421,12 +1419,7 @@ static void process_vforkThread(void *arg)
 	/* Copy parent kernel stack */
 	current->parentkstack = vm_kmalloc(process_parentKstacksz(parent));
 	if (current->parentkstack == NULL) {
-		hal_spinlockSet(&spawn->sl, &sc);
-		spawn->state = -ENOMEM;
-		(void)proc_threadWakeup(&spawn->wq);
-		hal_spinlockClear(&spawn->sl, &sc);
-
-		proc_threadEnd();
+		proc_spawnThreadEnd(spawn, -ENOMEM);
 	}
 
 	hal_memcpy(current->parentkstack, hal_cpuGetSP(parent->context), process_parentKstacksz(parent));
@@ -1441,12 +1434,7 @@ static void process_vforkThread(void *arg)
 	if (ret < 0) {
 		vm_kfree(current->parentkstack);
 
-		hal_spinlockSet(&spawn->sl, &sc);
-		spawn->state = ret;
-		(void)proc_threadWakeup(&spawn->wq);
-		hal_spinlockClear(&spawn->sl, &sc);
-
-		proc_threadEnd();
+		proc_spawnThreadEnd(spawn, ret);
 	}
 
 	proc_changeMap(current->process, parent->process->mapp, parent->process->imapp, parent->process->pmapp);
--- a/proc/process.c
+++ b/proc/process.c
@@ -88,7 +88,9 @@ static void process_destroy(process_t *p)
 
 	trace_eventProcessKill(p);
 
-	posix_died(process_getPid(p), p->exit);
+	if (p->posix != 0U) {
+		posix_died(process_getPid(p), p->exit);
+	}
 
 	/* Destroy resources (especially rtInth) before changing map to prevent race */
 	proc_resourcesDestroy(p);
@@ -222,6 +224,8 @@ int proc_start(startFn_t start, void *arg, const char *path)
 #else
 	process->lazy = 1;
 #endif
+
+	process->posix = 0;
 
 	proc_changeMap(process, NULL, NULL, NULL);
 
@@ -1178,10 +1182,21 @@ static void proc_spawnThread(void *arg)
 {
 	thread_t *current = proc_current();
 	process_spawn_t *spawn = arg;
+	spinlock_ctx_t sc;
+	int ret;
 
 	/* temporary: create new posix process */
 	if (spawn->parent != NULL) {
-		(void)posix_clone(process_getPid(spawn->parent->process));
+		ret = posix_clone(process_getPid(spawn->parent->process));
+		if (ret < 0) {
+			hal_spinlockSet(&spawn->sl, &sc);
+			spawn->state = ret;
+			(void)proc_threadWakeup(&spawn->wq);
+			hal_spinlockClear(&spawn->sl, &sc);
+
+			proc_threadEnd();
+		}
+		current->process->posix = 1U;
 	}
 
 	process_exec(current, spawn);
@@ -1383,7 +1398,16 @@ static void process_vforkThread(void *arg)
 
 	current = proc_current();
 	parent = spawn->parent;
-	(void)posix_clone(process_getPid(parent->process));
+	ret = posix_clone(process_getPid(spawn->parent->process));
+	if (ret < 0) {
+		hal_spinlockSet(&spawn->sl, &sc);
+		spawn->state = ret;
+		(void)proc_threadWakeup(&spawn->wq);
+		hal_spinlockClear(&spawn->sl, &sc);
+
+		proc_threadEnd();
+	}
+	current->process->posix = 1U;
 
 	current->process->sigmask = parent->process->sigmask;
 	current->process->sighandler = parent->process->sighandler;
--- a/proc/process.c
+++ b/proc/process.c
@@ -1385,11 +1385,8 @@ static void process_vforkThread(void *arg)
 	parent = spawn->parent;
 	(void)posix_clone(process_getPid(parent->process));
 
-	proc_changeMap(current->process, parent->process->mapp, parent->process->imapp, parent->process->pmapp);
-
 	current->process->sigmask = parent->process->sigmask;
 	current->process->sighandler = parent->process->sighandler;
-	pmap_switch(current->process->pmapp);
 
 	hal_spinlockSet(&spawn->sl, &sc);
 	while (spawn->state < FORKING) {
@@ -1427,6 +1424,9 @@ static void process_vforkThread(void *arg)
 
 		proc_threadEnd();
 	}
+
+	proc_changeMap(current->process, parent->process->mapp, parent->process->imapp, parent->process->pmapp);
+	pmap_switch(current->process->pmapp);
 
 	current->ustack = parent->ustack;
 
--- a/vm/amap.c
+++ b/vm/amap.c
@@ -118,8 +118,6 @@ amap_t *amap_create(amap_t *amap, size_t *offset, size_t size)
 			(void)proc_lockClear(&amap->lock);
 			return amap;
 		}
-
-		amap->refs--;
 	}
 
 	/* Allocate anon pointer arrays in chunks
@@ -151,6 +149,7 @@ amap_t *amap_create(amap_t *amap, size_t *offset, size_t size)
 	}
 
 	if (amap != NULL) {
+		amap->refs--;
 		(void)proc_lockClear(&amap->lock);
 	}
 
@@ -243,7 +242,6 @@ int amap_page(vm_map_t *map, amap_t *amap, vm_object_t *o, void *vaddr, size_t a
 			(void)proc_lockClear(&amap->lock);
 			return EOK;
 		}
-		a->refs--;
 	}
 	else {
 		err = vm_objectPage(map, &amap, o, vaddr, offs, page);
@@ -304,6 +302,7 @@ int amap_page(vm_map_t *map, amap_t *amap, vm_object_t *o, void *vaddr, size_t a
 	(void)amap_unmap(map, v);
 
 	if (a != NULL) {
+		a->refs--;
 		(void)proc_lockClear(&a->lock);
 	}
 
--- a/vm/amap.c
+++ b/vm/amap.c
@@ -226,9 +226,9 @@ static int amap_unmap(vm_map_t *map, void *v)
 }
 
 
-page_t *amap_page(vm_map_t *map, amap_t *amap, vm_object_t *o, void *vaddr, size_t aoffs, u64 offs, vm_prot_t prot)
+int amap_page(vm_map_t *map, amap_t *amap, vm_object_t *o, void *vaddr, size_t aoffs, u64 offs, vm_prot_t prot, page_t **page)
 {
-	page_t *p = NULL;
+	int err = EOK;
 	anon_t *a;
 	void *v, *w;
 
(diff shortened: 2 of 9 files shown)


vm/map.c
```diff
@@ -1153,18 +1157,21 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
...
+			if (err != EOK) {
+				vm_mapDestroy(proc, dst);
+			}
+			for (offs = 0; offs < e->size; offs += SIZE_PAGE) {
+				LIB_ASSERT_ALWAYS(_map_force(src, e, (void *)((ptr_t)e->vaddr + offs), e->prot) == EOK, "Broken src map during mapCopy");
+			}
+			if (err != EOK) {
+				(void)proc_lockClear(&src->lock);
+				return err;
+			}
```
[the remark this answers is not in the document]
etiaro: @gemini-code-assist This split error handling aims at making sure that src map is left unchanged to avoid putting process owning this map in invalid state, I included a new comment to explain this.
In properly working system with disabled lazy, this assertion should always pass. If it doesn't, failing assertion still looks better than running processes with their maps in incoherent state.

2. The loop does in fact break, it just breaks after cleaning up the source map changes.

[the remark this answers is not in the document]
etiaro: Can not immidiately exit as src entry is already remapped to readonly and has added NEEDSCOPY flag.
If  unused lazy mechanism will be removed in the future, the copy could be done without temporary  remapping to readonly

adamgreloch (dismissed):

vm/amap.c
```diff
@@ -303,17 +301,18 @@ page_t *amap_page(vm_map_t *map, amap_t *amap, vm_object_t *o, void *vaddr, size
...
 		(void)proc_lockClear(&a->lock);
 	}
 
-	amap->anons[aoffs / SIZE_PAGE] = anon_new(p);
+	amap->anons[aoffs / SIZE_PAGE] = anon_new(*res);
 	if (amap->anons[aoffs / SIZE_PAGE] == NULL) {
-		vm_pageFree(p);
-		p = NULL;
+		vm_pageFree(*res);
+		*res = NULL;
```
ziemleszcz: Previous version returns `-ENOMEM` in this case.

vm/map.c
```diff
@@ -1153,18 +1157,22 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
...
+				if (err != EOK) {
+					break;
 				}
 			}
+			if (err != EOK) {
+				/* Destroy map before _map_force on source map to ensure refcounts are cleared and avoid unnecessary copies in amap_page */
+				vm_mapDestroy(proc, dst);
+			}
+			for (offs = 0; offs < e->size; offs += SIZE_PAGE) {
+				LIB_ASSERT_ALWAYS(_map_force(src, e, (void *)((ptr_t)e->vaddr + offs), e->prot) == EOK, "Broken src map during mapCopy");
```
ziemleszcz: Why kernel panic on `ENOMEM`?

vm/map.c
```diff
@@ -1153,18 +1157,22 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
...
-					(void)proc_lockClear(&src->lock);
-					return -ENOMEM;
+				err = _map_force(dst, f, (void *)((ptr_t)f->vaddr + offs), f->prot);
+				if (err != EOK) {
+					break;
 				}
 			}
+			if (err != EOK) {
+				/* Destroy map before _map_force on source map to ensure refcounts are cleared and avoid unnecessary copies in amap_page */
+				vm_mapDestroy(proc, dst);
```
ziemleszcz: Note that `vm_mapDestroy()` takes `dst->lock` (MMU).

vm/map.c
```diff
@@ -1153,18 +1157,22 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
...
+			}
+			for (offs = 0; offs < e->size; offs += SIZE_PAGE) {
+				LIB_ASSERT_ALWAYS(_map_force(src, e, (void *)((ptr_t)e->vaddr + offs), e->prot) == EOK, "Broken src map during mapCopy");
+			}
+			if (err != EOK) {
+				(void)proc_lockClear(&src->lock);
+				return err;
```
ziemleszcz: `dst->lock` is not cleared.

proc/process.c
```diff
@@ -1503,6 +1503,9 @@ int proc_vfork(void)
 		(void)vm_objectPut(spawn->object);
 		ret = spawn->state;
 		vm_kfree(spawn);
+		if ((ret < 0) && (posix_getppid(pid) >= 0)) {
```
ziemleszcz: Please add some comment to this code.

vm/map.c
etiaro: ENOMEM should not happen for source map when lazy == 0. This loop should effectively clean NEEDSCOPY flags and remap pages with original attributes.

What better to do here if not kernel panic? 
source process memory is corrupted, so we shouldn't let it run. To me reboot with some error message seems more reasonable on RTOS system than killing a parent process that tried to fork(). Restarting the process could be a viable option, but we are running Out Of memory so it is likely to fail anyway.

Ideally, we should stop pretending that lazy mechanism is available and handle mapCopy without changing src map at any given moment, but IMO this is a larger rewrite that deserves separate PR.

vm/amap.c
etiaro: Nice catch, thanks

vm/object.c
```diff
@@ -207,33 +207,36 @@ static page_t *object_fetch(oid_t oid, u64 offs)
...
 
 	if (o == VM_OBJ_PHYSMEM) {
 		/* parasoft-suppress-next-line MISRAC2012-RULE_14_3 "Check is needed on targets where sizeof(offs) != sizeof(addr_t)" */
 		if (offs > (addr_t)-1) {
-			return NULL;
+			return -ERANGE;
 		}
-		return _page_get((addr_t)offs);
+		*res = _page_get((addr_t)offs);
+		/* res can be NULL, when address outside of defined physical maps is used */
```
ziemleszcz: Could you describe this in the header file?

vm/map.c
etiaro: From what I can see, it takes dst->lock only on NOMMU. proc->lazy is equal to 1 on NOMMU, so this isn't a problem. IMO whole vm_mapCopy should be only available on MMU (it's not used on NOMMU and does not make much sense there).

vm/map.c
etiaro: This lock does not exist here anymore (lockDone called in vm_mapDestroy)

vm/map.c
ziemleszcz: OK, I didn't notice it's MMU only code.

vm/map.c
ziemleszcz: Understood, but at L1139 the same lock is cleared before `vm_mapDestroy` call. Should we also clear this lock before L1167 for consistency?

etiaro: Sure, can do.

etiaro:
After this last batch of fixes, running a fork bomb did not cause kernel panics anymore (at least for a few minutes on ia32 and aarch64).

proc/process.c
```diff
@@ -1503,6 +1521,12 @@ int proc_vfork(void)
...
+			/* if child managed to register itself within posix subsystem before failure,
+			 * wait for its complete death and cleanup its posix metadata.
+			 */
+			(void)posix_waitpid(pid, NULL, 0);
+		}
```
adamgreloch:
```suggestion
			/*
			 * if child managed to register itself within posix subsystem before failure,
			 * wait for its complete death and cleanup its posix metadata.
			 */
			(void)posix_waitpid(pid, NULL, 0);
		}
```

vm/amap.c
```diff
@@ -237,61 +236,60 @@ page_t *amap_page(vm_map_t *map, amap_t *amap, vm_object_t *o, void *vaddr, size
...
 			}
 			(void)proc_lockClear(&amap->lock);
-			return NULL;
+			return -ENOMEM;
 		}
-		w = amap_map(map, p);
+		w = amap_map(map, *res);
 		if (w == NULL) {
-			vm_pageFree(p);
+			vm_pageFree(*res);
```
adamgreloch, line 286: maybe set *res to NULL after the free like in L311

vm/object.h
```diff
@@ -45,7 +45,11 @@ int vm_objectGet(vm_object_t **o, oid_t oid);
 int vm_objectPut(vm_object_t *o);
 
 
-page_t *vm_objectPage(struct _vm_map_t *map, amap_t **amap, vm_object_t *o, void *vaddr, u64 offs);
+/* Fills appropriate page_t struct into *res,
+ * allocates new page when o == NULL
+ * Note that when o == VM_OBJ_PHYSMEM and offs is outside of defined physical maps, then *res will be NULL even though mapping can still be made
+ */
```
adamgreloch:
```suggestion
/*
 * Fills appropriate page_t struct into *res, allocates new page when o == NULL.
 * Note that when o == VM_OBJ_PHYSMEM and offs is outside of defined physical maps,
 * then *res will be NULL even though mapping can still be made.
 */
```

adamgreloch (commented):
small things, apart from that LGTM

ziemleszcz:
Please resolve conflicts on this PR.

proc/process.c
```diff
@@ -1178,10 +1178,20 @@ static void proc_spawnThread(void *arg)
...
 	/* temporary: create new posix process */
 	if (spawn->parent != NULL) {
-		(void)posix_clone(process_getPid(spawn->parent->process));
+		ret = posix_clone(process_getPid(spawn->parent->process));
+		if (ret < 0) {
+			hal_spinlockSet(&spawn->sl, &sc);
+			spawn->state = ret;
+			(void)proc_threadWakeup(&spawn->wq);
+			hal_spinlockClear(&spawn->sl, &sc);
+
```
adamgreloch: kinda duplicated with L1396-1404, maybe extract this to some helper (`proc_spawnPosixClone`?) and do proc_threadEnd() when it returns err?

etiaro: Good idea

proc/process.c
```diff
@@ -1174,14 +1174,29 @@ static void process_exec(thread_t *current, process_spawn_t *spawn)
...
 static void proc_spawnThread(void *arg)
 {
 	thread_t *current = proc_current();
 	process_spawn_t *spawn = arg;
 
 	/* temporary: create new posix process */
 	if (spawn->parent != NULL) {
-		(void)posix_clone(process_getPid(spawn->parent->process));
+		proc_spawnPosixClone(spawn);
```
adamgreloch: A comment that this never returns on failure would be good to reduce surprises. Same below

This is why I was thinking about calling threadEnd after the clone helper not as part of it

etiaro: Oh, I misunderstood the suggestion. Maybe the other way around would be more convenient, to create a wrapper for pending-spawn process failure and threadEnd?

adamgreloch: Yeah, that's even better

adamgreloch (dismissed):

proc/process.c
```diff
@@ -1174,14 +1174,31 @@ static void process_exec(thread_t *current, process_spawn_t *spawn)
 }
 
 
+static void proc_spawnThreadEnd(process_spawn_t *spawn, int state)
```
ziemleszcz: Mark this function as noreturn.

vm/map.c
ziemleszcz: I’m not familiar with this code, but I don’t see why `ENOMEM` could not occur in `_map_force()`. How does `lazy == 0` guarantee that?

etiaro: When lazy is disabled, all pages in an entry are allocated/mapped/copied instantly (via _map_force) after each operation on a map (unless we changed the lazy flag in the middle of process execution).

When anons/object pages are already populated and there is only one ref to page's corresponding amap/anon, no allocation happens (only flags/mapping changes according to current config).

Here, no new entries are added or their size modified, the only thing that changed is the interfering map (the dst map) that is already destroyed (and thus its refs are cleared) and the reason to call _map_force is handling the NEEDSCOPY flag and temporary readonly mapping.
So failure of this assertion would mean there is something wrong with the state of this map entry.

etiaro: TL;DR the assumption here is that all src entries in process with lazy=0 are in coherent state (everything is allocated and there are no multi-refs) and will remain in that state after dst map is fully copied or destroyed.

ziemleszcz: It’s still not obvious to me why ENOMEM is not possible. If `_map_force()` is supposed to clear the NEEDSCOPY flag, it can still return ENOMEM a few lines earlier if `amap_create()` fails. Perhaps rewriting `_map_force()` for this particular case would make it clearer.

If there is even a slight probability of memory allocation failure here, the kernel should return an error instead of panicking.

etiaro: The idea is that amap_create won't return an error in this case (refs should be equal 1).
I agree that this is a complex subsystem, there is a chance I am missing something right now or someone else will in the future and cause panics.

I'm not sure what to do about possible errors in eg. amap_page in such rewritten _map_force?

Overall, this assertion exists because there is no one to return the error to: Error here would cause the current process's (fork parent) map to be left broken (it could eg. cause the caller stack to become readonly).
So, instead of systemwide panic, the only option is let the process crash or gracefully kill it.

Long term, I would opt for removal of whats left after lazy mechanism, which would allow for much cleaner mapCopy without such assertion. This would, among other changes, include rewriting _map_force.
For now, I think this assertion is better than alternatives and what was here before (memory leaks and vague process / kernel crashes during OOM conditions).

Maybe it would make sense to make this assertion DEBUG-only?

ziemleszcz: If you are certain that `_map_force()` cannot fail due to memory allocation in this case, please add a comment explaining why. If this guarantee depends on `refs == 1` or other hidden condition, add an explicit check and only trigger a kernel panic in that condition. Note that `_map_force()` or `amap_page()` may change in the future, and this assumption may no longer hold.

Alternatively, consider adding a simplified version of `_map_force()` for this specific case (with an internal variant of `amap_page()` under the assumption that `refs == 1`). Currently, `_map_force()` is called with the assumption that it cannot fail, which may not be obvious to others.

etiaro: Apologies, after sleeping through that, I figured there is a way to skip changes of source map altogether without larger code rewrite.
This removed this fragile, unhandleable breaking point from mapCopy.

I'll wait until master nightly work and test it on full test suite again before merging to ensure that I'm not missing something here.

proc/process.c
```diff
@@ -1503,6 +1503,13 @@ int proc_vfork(void)
 		(void)vm_objectPut(spawn->object);
 		ret = spawn->state;
 		vm_kfree(spawn);
+		if ((ret < 0) && (posix_getppid(pid) >= 0)) {
+			/*
+			 * if the child managed to register itself within posix subsystem before
+			 * failure, wait for its complete death and cleanup its posix metadata.
+			 */
+			(void)posix_waitpid(pid, NULL, 0);
```
ziemleszcz, line 1523: Why is this required? Wouldn’t the process be cleaned up automatically anyway?

vm/amap.c
```diff
@@ -226,9 +226,9 @@ static int amap_unmap(vm_map_t *map, void *v)
 }
 
 
-page_t *amap_page(vm_map_t *map, amap_t *amap, vm_object_t *o, void *vaddr, size_t aoffs, u64 offs, vm_prot_t prot)
+int amap_page(vm_map_t *map, amap_t *amap, vm_object_t *o, void *vaddr, size_t aoffs, u64 offs, vm_prot_t prot, page_t **res)
```
ziemleszcz: `res` is misleading for a page pointer, `p` or `page` would be clearer.

vm/map.c
```diff
@@ -778,7 +778,6 @@ static int _map_force(vm_map_t *map, map_entry_t *e, void *paddr, vm_prot_t prot
 		return -ENOMEM;
 	}
 	else if (page_map(&map->pmap, paddr, p->addr, attr) < 0) {
-		amap_putanons(e->amap, e->aoffs + offs, SIZE_PAGE);
```
ziemleszcz, line 778: Does this change rely on `vm_mapDestroy()` to release the resources allocated in this function?
What if `vm_mapDestroy()` is not called on` _map_force()` failure?

proc/process.c
```diff
@@ -1334,27 +1341,23 @@ static void process_restoreParentKstack(thread_t *current, thread_t *parent)
 
 static void proc_vforkedExit(thread_t *current, process_spawn_t *spawn, int state)
 {
-	spinlock_ctx_t sc;
-
 	current->ustack = NULL;
 	proc_changeMap(current->process, NULL, NULL, NULL);
 
+	proc_kill(current->process);
```
ziemleszcz, line 1352: This change reverses order of `proc_kill()` and `process_restoreParentKstack()`.

proc/process.c
etiaro: The cleanup procedure is shared with regular fully-started processes, posix_died() does not clean up process_info_t (unless it is adopted by init process) to keep it for parent's waitpid.
It would be cleaned up on parent's death/waitpid syscall, but until then it would hang in posix_common.pid tree.

Apart from hanging memory, wait/waitpid would return such "incomplete children", which doesn't seem like an [expected behavior in POSIX](https://pubs.opengroup.org/onlinepubs/009696799/functions/fork.html#:~:text=Otherwise%2C%20%2D1%20shall%20be%20returned%20to%20the%20parent%20process%2C%20no%20child%20process%20shall%20be%20created).

vm/map.c
etiaro: Generally yes, it depends on external code to release the potentially allocated `anon` and corresponding page (not necessarily mapDestroy).
If `amap_putanons` would not be called on _map_force failure, then this resources remain saved in map_entry_t.amap.anon, creating the incoherence between actually mapped memory and state of the map. Next usages of _map_force with these args, freeing on munmap or mapDestroy will behave as expected.
That being said, all of the functions handling _map_force error do end up calling amap_putanons, except for pageFault handler and mprotect*, that is allowed to only partially change protections ([POSIX](https://pubs.opengroup.org/onlinepubs/009696699/functions/mprotect.html#:~:text=may%20have%20been%20changed))

On the other hand, when it comes to objects, they are never freed granularly. So if a page was allocated here for object, it will remain allocated until whole object is destroyed.

*mprotect has another ENOMEM bug ([RTOS-1302](https://youtrack.phoenix-rtos.com/issue/RTOS-1302)) that I noticed but it's going to be included in separate PR.

etiaro: I could try to change the behavior of map_putanons to remove the references to freed anons, if you think that would simplify the logic here? 
Then this line could remain here, as the `anon` resources could be freed without the risk of a double-free in the future.

vm/amap.c
etiaro: I generally find output arguments misleading, hence the idea to at least indicate that in the variable name.
That being said, I'll go with `page` as it is more consistent with the existing codebase.

proc/process.c
etiaro: Yes, I noticed that, but it should not cause any problems, as proc_kill only sets up a flag for each of process's threads, that takes effect on first scheduling to userspace, which won't happen for this thread thanks to threadEnd() call.

vm/map.c
```diff
@@ -1146,23 +1149,22 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
...
+					vm_mapDestroy(proc, dst);
 
-				if (_map_force(src, e, (void *)((ptr_t)e->vaddr + offs), e->prot) != 0) {
-					(void)proc_lockClear(&dst->lock);
 					(void)proc_lockClear(&src->lock);
-					return -ENOMEM;
+					return err;
 				}
 			}
 		}
```
ziemleszcz, line 1169: Why not clear MAP_NEEDSCOPY flag on `e` here:

e->flags &- ~MAP_NEEDSCOPY;

vm/map.c
```diff
@@ -1146,23 +1149,22 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
 			f->flags |= MAP_NEEDSCOPY;
 
 			for (offs = 0; offs < f->size; offs += SIZE_PAGE) {
-				remap_readonly(src, e, offs);
+				if ((proc != NULL) && (proc->lazy != 0U)) {
+					remap_readonly(src, e, offs);
+				}
 				remap_readonly(dst, f, offs);
```
ziemleszcz: Why call `remap_readonly(dst, f, offs)` on freshly created pmap (no user space pages)? Isn't it no-op?

vm/map.c
```diff
@@ -1146,23 +1149,22 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
...
 		if ((proc == NULL) || (proc->lazy == 0U)) {
 			for (offs = 0; offs < f->size; offs += SIZE_PAGE) {
-				if (_map_force(dst, f, (void *)((ptr_t)f->vaddr + offs), f->prot) != 0) {
+				err = _map_force(dst, f, (void *)((ptr_t)f->vaddr + offs), f->prot);
+				if (err != EOK) {
 					(void)proc_lockClear(&dst->lock);
-					(void)proc_lockClear(&src->lock);
-					return -ENOMEM;
-				}
+					vm_mapDestroy(proc, dst);
```
ziemleszcz: Can you move it after `proc_lockClear(&src->lock)` so that both locks are released before this call as in L1140?

vm/map.c
etiaro: These are the remains of lazy mechanism. With lazy, this would be the only place this entry is added to pmap, to be later handled on write attempt inside pageFault handler.

vm/map.c
etiaro: I don't think we can do that - what if it had this flag set before this mapCopy call? 

Theoretically, we could save the flags and restore them for this failed entry, but to be completely fair all previous successful entries should be restored too.

proc/process.c
```diff
@@ -1174,14 +1174,31 @@ static void process_exec(thread_t *current, process_spawn_t *spawn)
...
 	thread_t *current = proc_current();
 	process_spawn_t *spawn = arg;
+	int ret;
 
 	/* temporary: create new posix process */
 	if (spawn->parent != NULL) {
-		(void)posix_clone(process_getPid(spawn->parent->process));
+		ret = posix_clone(process_getPid(spawn->parent->process));
+		if (ret < 0) {
+			proc_spawnThreadEnd(spawn, ret);
```
ziemleszcz, line 1204: Forcing `posix_clone()` to return ENOMEM triggers `kernel (posix_died:2754): pinfo not found` message.

This message appears intended for unexpected process death, not allocation failure. Can this case be handled more gracefully?

There also seems to be a memory leak when `posix_clone()` is forced to return with ENOMEM. Could you investigate this as well?

proc/process.c
```diff
@@ -1324,27 +1341,23 @@ static void process_restoreParentKstack(thread_t *current, thread_t *parent)
 
 static void proc_vforkedExit(thread_t *current, process_spawn_t *spawn, int state)
```
ziemleszcz: Could be marked with `__attribute__((noreturn))`.

proc/process.c
etiaro: Best I could come up is additional flag in process_t to note that the process is present in posix subsystem.
I also reverted 731c0fa that changed `LIB_ASSERT_ALWAYS` into `lib_printf`.

The memory leak turned out to be one-time allocation during lib_printf.

vm/map.c
ziemleszcz: OK, so why isn’t this inside inside `if ((proc != NULL) && (proc->lazy != 0U))`?

etiaro: Good point.

Coming back to the first one, `remap_readonly()` has no chance of working, as it resolves page using pmap_resolve instead of map entries, and indeed, for freshly created pmap, that will not be resolved.
But i leave it with details in YouTrack issue.

ziemleszcz (approved):

PR #753, opened 2026-03-26
posix: make posix access modes enum fields

oI0ck:
Most POSIX conformant/compilant systems treat O_ACCMODE fields as enum fields, not bitfields, which carries on to their usage semantics, making it impossible to repesent states when modes are not mutually exclusive.

JIRA: RTOS-1088

## Types of changes

## How Has This Been Tested?

- [X] Tested by hand on: `ia32-generic-qemu`, `armv7a9-zynq7000-qemu`.

## Checklist:

## Special treatment

### DIFF
--- a/include/posix-fcntl.h
+++ b/include/posix-fcntl.h
@@ -19,9 +19,18 @@
 
 #define FD_CLOEXEC 0x1U
 
-#define O_RDONLY   0x00001U
-#define O_WRONLY   0x00002U
-#define O_RDWR     0x00004U
+/*
+ * These are enum values, not a bitfields, because POSIX requires access modes
+ * to be mutually exclusive. It does not specify the format or values for those
+ * definitions, however, most POSIX conformant operating systems treat those
+ * values as enumerations.
+ */
+#define O_RDONLY 0x00000U
+#define O_WRONLY 0x00001U
+#define O_RDWR   0x00002U
+
+#define O_ACCMODE 0x00003U
+
 #define O_APPEND   0x00008U
 #define O_CREAT    0x00100U
 #define O_TRUNC    0x00200U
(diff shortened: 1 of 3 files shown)


include/posix-fcntl.h
```diff
@@ -19,9 +19,18 @@
 
 #define FD_CLOEXEC 0x1U
 
-#define O_RDONLY   0x00001U
-#define O_WRONLY   0x00002U
-#define O_RDWR     0x00004U
+/*
```
nalajcie, line 22: technically we could decide to not change the numeric values of these defines - that way we will keep binary compatibility (so the change would not be breaking)?

oI0ck: Technically, yes, though, having none of these bits set is an invalid state, which is why I marginally prefer starting with 0, although even with constants starting from 0, one could have a `11` state, which is also incorrect, so that might not matter as much as I'd like it to.

oI0ck: @Darchiv What do you think?

Darchiv: I don't think we have to avoid breaking changes since we already have broken multiple things on the master branch and recompilation is needed anyway. The "ABI stability" policy will be in effect since release 3.4. IMHO we can change the values.

oI0ck: IMHO we can change it later, but we should disassociate this change from WAMR patchset.
This one might require a wider evaluation of what might break.
For example, LWIP (because it is a GNU project) interprets access modes as bitfields, I don't have knowledge on our implementation, but intuition tells me that there might be places where this change will break other components.

For merging a port which is mostly done, this might be too much effort for now, we should postpone doing this.

PR #757, opened 2026-04-07
hal/armv8m: allow killing processes on exceptions

jakubsmolaga:
TASK: PP-462

## Description
This change allows programmers to control the behavior of exceptions through the `KERNEL_REBOOT_ON_EXCEPTION` macro. Setting it will cause a reboot/halt on an exception (current behavior), while not setting it will cause the offending process to be killed and reported in the console (new behavior).

## Motivation and Context

Currently all exceptions either halt or reboot the CPU and lack useful context for the crash (like PID and TID of the culprit). This is making certain issues harder to debug than they need to be, thus slowing down the work in our team.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `armv8m-stm32n6`

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv8m/exceptions.c
+++ b/hal/armv8m/exceptions.c
@@ -41,6 +41,14 @@ enum exceptions {
 };
 
 
+static struct {
+	excHandlerFn_t defaultHandler;
+} hal_exceptions_common;
+
+
+extern void hal_exceptionJump(unsigned int n, exc_context_t *ctx, void (*handler)(unsigned int, exc_context_t *));
+
+
 void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, unsigned int n)
 {
 	static const char *mnemonics[] = {
@@ -135,6 +143,12 @@ void exceptions_dispatch(unsigned int n, exc_context_t *ctx)
 {
 	char buff[SIZE_CTXDUMP];
 
+	/* Don't call the handler on kernel exceptions */
+	if ((ctx->excret == RET_THREAD_PSP) && hal_exceptions_common.defaultHandler != NULL) {
+		hal_exceptionJump(n, ctx, hal_exceptions_common.defaultHandler);
+		return;
+	}
+
 	hal_exceptionsDumpContext(buff, ctx, n);
 	hal_consolePrint(ATTR_BOLD, buff);
 
@@ -181,10 +195,17 @@ void *hal_exceptionsFaultAddr(unsigned int n, exc_context_t *ctx)
 
 int hal_exceptionsSetHandler(unsigned int n, excHandlerFn_t handler)
 {
+#ifndef KERNEL_REBOOT_ON_EXCEPTION
+	if (n == EXC_DEFAULT) {
+		hal_exceptions_common.defaultHandler = handler;
+	}
+#endif
+
 	return 0;
 }
 
 
 void _hal_exceptionsInit(void)
 {
+	hal_exceptions_common.defaultHandler = NULL;
 }


hal/armv8m/exceptions.c
```diff
@@ -135,6 +140,10 @@ void exceptions_dispatch(unsigned int n, exc_context_t *ctx)
 {
 	char buff[SIZE_CTXDUMP];
 
+	if (exceptions_common.defaultHandler != NULL) {
+		return exceptions_common.defaultHandler(n, ctx);
```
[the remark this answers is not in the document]
jakubsmolaga: done

agkaminski:
CPU reboot on exception is intentional on NOMMU - due to partial at best memory protection system is deemed damaged and not trustworthy anymore after exception.

STM32N6 hangs due to not yet implemented reboot.

But I'm ok with this being configurable behavior, would leave reboot default though.

Why is disabling the posix_write needed? What happens if you do not disable it?

jakubsmolaga:
> CPU reboot on exception is intentional on NOMMU - due to partial at best memory protection system is deemed damaged and not trustworthy anymore after exception.

Are the illegal operations still committed even if the exception occurs? If not, then I might lack understanding of why the memory protection is deemed damaged.

> STM32N6 hangs due to not yet implemented reboot.

Yes, this should probably be done at some point but during development it's not an issue.

> But I'm ok with this being configurable behavior, would leave reboot default though.

Maybe putting this behind some `#ifndef NDEBUG`?

> Why is disabling the posix_write needed? What happens if you do not disable it?

The CPU just dies in a silent way that I've not been able to debug.

agkaminski:
There might've been milion bytes of other process being altered before exception is triggered ;) E.g. stack overflow.

It would be best to get `posix_write` going, but putting it under `#ifndef NOMMU` would be ok for now (with some FIXME comment). Even better - with FIXME that explains **why** it's broken, if the fix is somehow difficult.

hal/armv8m/exceptions.c
```diff
@@ -41,6 +41,11 @@ enum exceptions {
 };
 
 
+static struct {
+	excHandlerFn_t defaultHandler;
+} exceptions_common = { 0 };
```
nalajcie: note: this probably doesn't work. In normal code it's a no-op (bss is zero-initialized by default so you shoudn't use a construction like that). In kernel code bss is not zero-initialized (as per https://docs.phoenix-rtos.com/latest/coding/index.html#ps-c-rec-019 note).

jakubsmolaga: Thank you, moved initialization to `_hal_exceptionsInit`

anglov:
Something was done for armv7m, see: 
https://github.com/phoenix-rtos/phoenix-rtos-kernel/commit/55705813be4a6e380a59141a4b32196aa0621120

This behavior is configured by variable KERNEL_REBOOT_ON_EXCEPTION

Maybe it will be good starting point for you.

agkaminski:
Lol, exception jump asm function was already present, nice. Perhaps you can make exception.c a direct copy of armv7 one to ease much needed unification. Otherwise LGTM (I assume it works. On a fatal path there's no interrupt disable, so were they already disabled? Unsettling difference with the armv7 implementation).

jakubsmolaga:
> Something was done for armv7m

This was indeed very helpful, thank you.

> On a fatal path there's no interrupt disable, so were they already disabled? Unsettling difference with the armv7 implementation

I might be wrong, but I think on armv7m there is just a superfluous interrupt disable since interrupts are already disabled before entering the exception handler (`cpsid if`)

There are still some things that don't make sense to me, namely:
- How did my previous code ever work in the first place (`hal_exceptionJump` seems to be required to return from handler mode to thread mode)
- Why did `posix_write` in particular break the previous code?

I'll try to investigate it today after work. I don't feel comfortable going forward with this until I feel like I have a good grasp of what's going on.

jakubsmolaga:
Ok, I can't seem to figure out what was going on with `posix_write`, and how did it work at all in the first place. In particular this is baffling to me:
```
void hal_consolePutch(char c)
{
	while (((*(console_common.base + isr)) & UART_ISR_TXE) == 0U) {
		/* Wait for transmit register empty */
	}

	*(console_common.base + tdr) = (u32)c;
}
```
How did this not deadlock if interrupts are disabled?

Either way I don't really have the time to investigate this further ( :cry: ) so I would opt for merging it as is.

hal/armv8m/exceptions.c
```diff
@@ -134,6 +142,13 @@ void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, unsigned int n)
 void exceptions_dispatch(unsigned int n, exc_context_t *ctx)
 {
 	char buff[SIZE_CTXDUMP];
+	int isUserspace = (ctx->excret == RET_THREAD_PSP);
+
+	/* Don't call the handler on kernel exceptions */
+	if (isUserspace && hal_exceptions_common.defaultHandler != NULL) {
```
nalajcie: MISRA: int-to-bool implicit conversion

hal/armv8m/exceptions.c
```diff
@@ -134,6 +142,13 @@ void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, unsigned int n)
 void exceptions_dispatch(unsigned int n, exc_context_t *ctx)
 {
 	char buff[SIZE_CTXDUMP];
+	int isUserspace = (ctx->excret == RET_THREAD_PSP);
```
nalajcie: MISRA: bool to int implicit conversion

jakubsmolaga: Fixed, I used `_Bool` because I wasn't sure if we can rely on `stdbool.h` in the kernel. Is that ok?

jmaksymowicz:
@jakubsmolaga 
> Ok, I can't seem to figure out what was going on with `posix_write`, and how did it work at all in the first place. In particular this is baffling to me:
> 
> ```
> void hal_consolePutch(char c)
> {
> 	while (((*(console_common.base + isr)) & UART_ISR_TXE) == 0U) {
> 		/* Wait for transmit register empty */
> 	}
> 
> 	*(console_common.base + tdr) = (u32)c;
> }
> ```
> 
> How did this not deadlock if interrupts are disabled?
> 
> Either way I don't really have the time to investigate this further ( 😢 ) so I would opt for merging it as is.

This will work regardless of if the interrupts are disabled because this code checks interrupt flags in the UART peripheral - these are set no matter if interrupts are masked in the CPU.

The way an interrupt happens is usually somewhat like this:
1. Something happens in the peripheral
2. Peripheral sets a bit in its status register
3. Peripheral checks if this interrupt is masked in its own mask register; if not, it activates interrupt line going to the CPU
4. When the interrupt line is active, CPU checks if interrupts are disabled; if they are, the interrupt is deferred until interrupts are enabled
5. Once interrupts are enabled in the CPU, the CPU handles the interrupt (on ARMv8-M it stacks some of the registers then jumps to the appropriate vector).

Masking interrupts in the CPU using `cpsid i` only affects point 4.

jakubsmolaga:
> @jmaksymowicz 
> This will work regardless of if the interrupts are disabled because this code checks interrupt flags in the UART peripheral - these are set no matter if interrupts are masked in the CPU.

That makes sense, I suppose the only mystery still remaining is what property of `posix_write` broke things when it worked otherwise

jmaksymowicz:
> STM32N6 hangs due to not yet implemented reboot.

Only on pilot boards, because Flash is not properly reset during reboot. As a result, after CPU reset the BootROM cannot communicate with Flash and enters serial boot mode (you can tell because DFU device appears on USB). On Nucleo reboot works correctly.

hal/armv8m/exceptions.c
```diff
@@ -134,6 +142,13 @@ void hal_exceptionsDumpContext(char *buff, exc_context_t *ctx, unsigned int n)
 void exceptions_dispatch(unsigned int n, exc_context_t *ctx)
 {
 	char buff[SIZE_CTXDUMP];
+	_Bool isUserspace = (ctx->excret == RET_THREAD_PSP);
+
+	/* Don't call the handler on kernel exceptions */
+	if (isUserspace && hal_exceptions_common.defaultHandler != NULL) {
```
agkaminski: `isUserspace` is kinda redundant as it's used only once
```suggestion
	if ((ctx->excret == RET_THREAD_PSP) && (hal_exceptions_common.defaultHandler != NULL)) {
```

jakubsmolaga: Done

jakubsmolaga:
Is there anything else I need to do before this can get merged?

PR #758, opened 2026-04-15
vm/zone: Fix bad arg check in _vm_zoneCreate

agkaminski:
Intend was to check if blocksz is a power of 2, one extra ~ caused the check to become not equal to zero instead.

TASK: RTOS-1288

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/vm/zone.c
+++ b/vm/zone.c
@@ -32,11 +32,8 @@ int _vm_zoneCreate(vm_zone_t *zone, size_t blocksz, unsigned int blocks)
 {
 	size_t i;
 
-	if ((blocksz == 0U) || (blocks == 0U)) {
-		return -EINVAL;
-	}
-
-	if ((blocksz & ~(blocksz - 1U)) == 0U) {
+	/* blocksz has to be a power of 2 */
+	if ((blocksz == 0UL) || (blocks == 0U) || ((blocksz & (blocksz - 1UL)) != 0UL)) {
 		return -EINVAL;
 	}
 
--- a/vm/zone.c
+++ b/vm/zone.c
@@ -114,10 +114,6 @@ void _vm_zfree(vm_zone_t *zone, void *block)
 		return;
 	}
 
-	if (((unsigned long)block & ~(zone->blocksz - 1U)) == 0U) {
-		return;
-	}
-
 	*((void **)block) = zone->first;
 	zone->first = block;
 	zone->used--;


vm/zone.c
```diff
@@ -32,11 +32,8 @@ int _vm_zoneCreate(vm_zone_t *zone, size_t blocksz, unsigned int blocks)
 {
 	size_t i;
 
-	if ((blocksz == 0U) || (blocks == 0U)) {
-		return -EINVAL;
-	}
-
-	if ((blocksz & ~(blocksz - 1U)) == 0U) {
+	/* blocksz has to be a power of 2 */
+	if ((blocksz == 0UL) || (blocks == 0U) || ((blocksz & (blocksz - 1UL)) != 0UL)) {
```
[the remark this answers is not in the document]
agkaminski: Sowwy, my IDE ate my `)`. It has already been... punished.

agkaminski:
Drafting for now, there seem to be an issue with the kmalloc itself.

CI is not passing because fixing the check causes memory to be not freed and out-of-memory on smaller targets

agkaminski:
Removed buggy check in `_vm_zfree` altogether.  It was testing if block address was within 0 to blocksz, which there's no reason to. Seem that the intention was to check if the block address is aligned to the blocksz, which breaks if blocksz is bigger than PAGE_SIZE.

vm/zone.c
ziemleszcz, line 36: Suggestion: add `|| (blocksz < sizeof(void*))` check.

vm/zone.c
```diff
@@ -32,11 +32,8 @@ int _vm_zoneCreate(vm_zone_t *zone, size_t blocksz, unsigned int blocks)
 {
 	size_t i;
 
-	if ((blocksz == 0U) || (blocks == 0U)) {
```
ziemleszcz, line 35: `blocksz == 0UL`?

vm/zone.c
ziemleszcz, line 40: It would nice to have `size_t` overflow check on `blocks * blocksz`.

vm/zone.c
agkaminski: Honestly, I would rather remove these checks altogether - these do nothing, only `kmalloc` uses this module and it does it in a predictable way

vm/zone.c
agkaminski: Same as above comment

vm/zone.c
adamgreloch: Then maybe demote them to DEBUG-level assertions? They still can be useful from self-testing perspective

vm/zone.c
agkaminski: Not possible, if blocksz > SIZE_PAGE then blocks = 1

https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/1e8be4ccf017f477a084dc8ea18515c9d4ac9653/vm/kmalloc.c#L131

vm/zone.c
agkaminski: Not possible
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/1e8be4ccf017f477a084dc8ea18515c9d4ac9653/vm/kmalloc.c#L156

agkaminski: > Then maybe demote them to DEBUG-level assertions? They still can be useful from self-testing perspective

For now I want to fix nonsensical checks, if I find time later I'd love to remove all runtime checks here and add some more asserts

vm/zone.c
ziemleszcz: Assuming external code doesn't change but OK.

vm/zone.c
ziemleszcz: Which comment? You are using `0UL`  at L36.

agkaminski: I've meant the previous conversation ;)

ziemleszcz (approved):

PR #759, opened 2026-04-15
!posix/ioctl: handle posix-defined requests

### LINKED ISSUE #1435: `ioctl()` on wrong device returns `EINVAL` instead of appropriate error (`ENOTTY`, `ENOTSOCK`, etc.)
I think [`posix_ioctl()`](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/master/posix/posix.c#L1718) should handle this by checking the file type against the ioctl request type and if it doesn't match just return error without sending a message. To do so, an additional macro is needed to extract the request type from the ioctl request.

One example of reproduction (ia32-generic-qemu):
```c
#include <stdio.h>
#include <unistd.h>
#include <fcntl.h>
#include <termios.h>

int main(void)
{
	int fd;
	struct termios t;
	char *name = "testfile";

	fd = creat(name, 0666);

	if (tcgetattr(fd, &t) < 0) {
		perror("tcgetattr");
		return 1;
	}

	unlink(name);
	close(fd);

	return 0;
}
```

Image

julianuziemblo:
Return ENOTTY or ENOTSOCK on file type mismatch
between the open file and the request.

Fixes: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1435

This is breaking as SOCK_IOC_TYPE was transferred to kernel and TTY_IOC_TYPE was defined in kernel.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [ ] This PR needs additional PRs to work:
    - https://github.com/phoenix-rtos/libphoenix/pull/468
    - https://github.com/phoenix-rtos/phoenix-rtos-build/pull/273 - *armv8r target doesn't build without this so all tests fail immediatelly* 

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -1755,6 +1755,28 @@ static int ioctl_processResponse(const msg_t *msg, unsigned long request, void *
 }
 
 
+static int ioctl_checkPosixDefinedRequests(open_file_t *f, unsigned long request)
+{
+	unsigned long group = IOCGROUP(request);
+
+	switch (group) {
+		case (unsigned long)TTY_IOC_TYPE:
+			/*
+			 * FIXME: for now, every character device is treated as a TTY.
+			 * Find a way to change this behavior.
+			 */
+			return (f->type == ftTty) ? EOK : -ENOTTY;
+
+		/* not required by POSIX, but nice to have */
+		case (unsigned long)SOCK_IOC_TYPE:
+			return ((f->type == ftUnixSocket) || (f->type == ftInetSocket)) ? EOK : -ENOTSOCK;
+
+		default:
+			return EOK;
+	}
+}
+
+
 int posix_ioctl(int fildes, unsigned long request, u8 *ustack)
 {
 	TRACE("ioctl(%d, %d)", fildes, request);
@@ -1767,21 +1789,16 @@ int posix_ioctl(int fildes, unsigned long request, u8 *ustack)
 
 	err = posix_getOpenFile(fildes, &f);
 	if (err == EOK) {
-		/* TODO: handle POSIX defined requests */
-		if (size > 0U) {
+		err = ioctl_checkPosixDefinedRequests(f, request);
+
+		if (err == EOK && size > 0U) {
 			GETFROMSTACK(ustack, void *, data, 2);
 			/* the actual size of the pointed-to structure: >= IOCPARM_LEN(request) */
 			GETFROMSTACK(ustack, size_t, size, 3);
 
 			if ((request & IOC_INOUT) != 0U) {
-				if (data == NULL) {
+				if (data == NULL || vm_mapBelongs(proc_current()->process, data, size) < 0) {
 					err = -EFAULT;
-				}
-				else if (vm_mapBelongs(proc_current()->process, data, size) < 0) {
-					err = -EFAULT;
-				}
-				else {
-					/* Nothing to do */
 				}
 			}
 		}
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -556,19 +556,42 @@ int posix_statvfs(const char *path, int fildes, struct statvfs *buf)
 }
 
 
+static int posix_getAttr(oid_t oid, int type, long long *attr)
+{
+	int err;
+	msg_t msg;
+
+	hal_memset(&msg, 0, sizeof(msg));
+	msg.type = mtGetAttr;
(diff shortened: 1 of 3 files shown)


nalajcie:
instead of `Fixes: #1435` you need to provide full `org/repo#issue` stem (as per commit message guidelines), otherwise it won't work (issue is in different repo than the change)

posix/posix.c
```diff
@@ -638,16 +661,38 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
...
+
+				if (posix_getAttr(f->ln, atMode, &attr) < 0) {
+					/* report files that don't report attr as regular files */
+					f->type = ftRegular;
+					break;
+				}
+				else {
+					mode = (mode_t)attr;
+					if (S_ISCHR(mode)) {
+						f->type = ftTty;
```
ziemleszcz, line 685: This is an oversimplification. Not all character devices are TTYs.
This breaks `posix_fstat()` and `ioctl_checkPosixDefinedRequests()` for devices that were previously classified as `ftRegular`.

posix/posix.c
```diff
@@ -638,16 +661,38 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
...
+					}
+					else if (S_ISFIFO(mode)) {
+						f->type = ftFifo;
+					}
+					else {
+						f->type = ftRegular;
```
ziemleszcz, line 691: File type enums should probably be extended.
At a minimum, add a comment noting that `ftRegular` currently also encompasses internet sockets, block devices, links, and directories.

posix/posix.c
```diff
@@ -638,16 +661,38 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
...
-			else {
-				f->type = ftRegular;
-			}
+			do {
+				if (oid.port == US_PORT) {
+					f->type = ftUnixSocket;
+					break;
+				}
+
+				if (proc_lookup("/dev/posix/pipes", NULL, &srv) >= 0) {
```
ziemleszcz, line 670: The commit message says `lookup pipesrv only if we're not US_PORT and handle inet socket server`. Where is the inet socker server part?

julianuziemblo: it was there but it made no sense as socksrv isn't a socket in itself but rather handles ctrl messages (sockmSocket, sockmGetNameInfo, sockmGetAddrInfo, sockmGetIfAddrs). forgot to remove from commit msg, deleted.

posix/posix.c
julianuziemblo: yes, but there's no other way for us to differentiate it here. This way at least it works for files that it should work for, e.g. /dev/console

julianuziemblo: I added a comment to the ftRegular enum value. Should I also add a FIXME comment here to somehow differentiate the types in the future, based on some other potential metadata that we don't have now?

ziemleszcz: Please add a TODO comment here.

This line still changes `posix_fstat()` behavior for character devices.
Also, `ioctl_checkPosixDefinedRequests()` accepts TTY ioctls for all character devices.

These issues should be fixed or documented with TODO comments in the relevant functions.

posix/posix.c
```diff
@@ -638,16 +661,41 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
...
+
+				if (posix_getAttr(f->ln, atMode, &attr) < 0) {
+					/* report files that don't report attr as regular files */
+					f->type = ftRegular;
+					break;
+				}
+				else {
+					mode = (mode_t)attr;
+					if (S_ISCHR(mode)) {
+						/* FIXME: for now, we treat every character device as
```
ziemleszcz: Fix all multi-line comments according to our coding style.

posix/posix.c
```diff
@@ -638,16 +661,41 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
...
+						   a TTY. This behavior should be fixed as it changes
+						   the behavior of e.g. posix_fstat() and ioctl(). */
+						f->type = ftTty;
+					}
+					else if (S_ISFIFO(mode)) {
+						f->type = ftFifo;
+					}
+					else {
+						f->type = ftRegular;
```
ziemleszcz, line 701: Add FIXME comment here as well (links, dir etc. are treated as regular files).

ziemleszcz (approved):

PR #760, opened 2026-04-21
vm: drop underscore from _page_get()

### LINKED ISSUE #1074: vm: `_page_get()` may cause unexpected behavior
Function that can cause a problem: `phoenix-rtos-kernel/vm/page.c:_page_get()`: https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/f29031f5ca6707f70e2720819f1255828a0c5cba/vm/page.c#L177

The problem: `_page_get()` searches through the binary tree without locking the pages lock. This function is used here: https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/f29031f5ca6707f70e2720819f1255828a0c5cba/vm/map.c#L949

There is possibility that one thread indirectly executes page deallocation (for example through msg_respond()->msg_release()->vm_pageFree()) while other thread indirectly seeks pages (syscalls_beginthreadex()->proc_put()->process_destroy()->vm_mapDestroy()-> _page_get()) then the binary tree might be changed while it is searched which leads to undefined behavior. I did not examine in detail this approach (maybe other lock prevents it) but there is also other use in https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/f29031f5ca6707f70e2720819f1255828a0c5cba/vm/object.c#L201 which is also unguarded.

adamgreloch:
Pages in pages_info.pages are never moved and their addresses don't change after _page_init(), only their state is updated. So page_get() is generally safe to be invoked in normal kernel code without external sync.

This PR adds invariant comments and replaces allocsz + freesz with totalsz to make the safety clearer.

Fixes: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1074 (though this is more of a refactor/issue dismiss)

JIRA: RTOS-1293

## Types of changes

- [X] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/vm/map.c
+++ b/vm/map.c
@@ -1050,7 +1050,7 @@ void vm_mapDestroy(process_t *p, vm_map_t *map)
 		if (a == 0U) {
 			break;
 		}
-		vm_pageFree(_page_get(a));
+		vm_pageFree(page_get(a));
 	}
 
 	(void)vm_munmap(map_common.kmap, map->pmap.pmapv, SIZE_PDIR);
--- a/vm/object.c
+++ b/vm/object.c
@@ -220,7 +220,7 @@ page_t *vm_objectPage(vm_map_t *map, amap_t **amap, vm_object_t *o, void *vaddr,
 		if (offs > (addr_t)-1) {
 			return NULL;
 		}
-		return _page_get((addr_t)offs);
+		return page_get((addr_t)offs);
 	}
 
 	(void)proc_lockSet(&object_common.lock);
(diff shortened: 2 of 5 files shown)


agkaminski (approved):
LGTM. Perhaps we could even keep number of pages in statistics instead of bytes, would save many `/ SIZE_PAGE`, but it might be too much at once

etiaro (approved):

PR #761, opened 2026-04-21
lib/cbuffer: assert size is a power of 2 in _cbuffer_init

julianuziemblo:
cbuffer's size is used as a bitmask in read and write, which only works if it's a power of 2.

YT: RTOS-1294

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/lib/cbuffer.c
+++ b/lib/cbuffer.c
@@ -20,6 +20,8 @@
 
 void _cbuffer_init(cbuffer_t *buf, void *data, size_t sz)
 {
+	LIB_ASSERT_ALWAYS(((sz == 0U) || ((sz & (sz - 1U)) == 0U)), "cbuffer's size has to be either 0 or a power of 2");
+
 	hal_memset(buf, 0, sizeof(cbuffer_t));
 	buf->sz = sz;
 	buf->data = data;
--- a/posix/unix.c
+++ b/posix/unix.c
@@ -1099,11 +1099,10 @@ int unix_shutdown(unsigned int socket, int how)
 static int unix_bufferSetSize(unixsock_t *s, int sz)
 {
 	void *v[2] = { NULL, NULL };
-	size_t size = (size_t)sz;
-
-	if (size < US_MIN_BUFFER_SIZE || size > US_MAX_BUFFER_SIZE) {
-		return -EINVAL;
-	}
+	/* cut requested buffer size to [US_MIN_BUFFER_SIZE, US_MAX_BUFFER_SIZE] range */
+	size_t size = min(max((size_t)sz, US_MIN_BUFFER_SIZE), US_MAX_BUFFER_SIZE);
+	/* round size to the next power of 2 */
+	size = (size_t)1U << (hal_cpuGetLastBit(size - 1U) + 1U);
 
 	(void)proc_lockSet(&s->lock);
 


lib/cbuffer.c
```diff
@@ -20,6 +20,8 @@
 
 void _cbuffer_init(cbuffer_t *buf, void *data, size_t sz)
 {
+	LIB_ASSERT_ALWAYS((sz & (sz - 1)) == 0, "cbuffer's size has to be a power of 2");
```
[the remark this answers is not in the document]
julianuziemblo: we allow 0-sized cbuffers for now. This is safe as `0xffffffff & 0x0` is 0

lib/cbuffer.c
```diff
@@ -20,6 +20,8 @@
 
 void _cbuffer_init(cbuffer_t *buf, void *data, size_t sz)
 {
+	LIB_ASSERT_ALWAYS((sz & (sz - 1U)) == 0U, "cbuffer's size has to be a power of 2");
```
adamgreloch: I know that this works for `sz=0` due to wrap-around but I'd not use such tricks and handle `sz=0` explicitly for clarity. GCC will reduce it to optimal expression anyway.

BTW: MISRA 12.4, while advisory, doesn't like wrap arounds too

ziemleszcz: Zero is not a power 2, please update the message.
Unix socket use size 0 (with NULL data) and and allow arbitrary buffer sizes via `SO_RCV_BUF` option.
So `cbuffer` should not be limited to power of two sizes, or `SO_RCV_BUF` handling code should be updated to reject sizes that are not powers of two.

julianuziemblo: For the 2nd thing: you think we should update SO_RCV_BUF, or rather use modulo operator here? Or just "grow" the requested buffer there to the next power of 2 if possible? I unfortunately didn't find any specification in POSIX for this

ziemleszcz: Rounding to the next power of two seems reasonable. If I remember correctly, Linux may also adjust the requested size, so the app should verify the actual value using `getsockopt()`.

lib/cbuffer.c
```diff
@@ -20,6 +20,8 @@
 
 void _cbuffer_init(cbuffer_t *buf, void *data, size_t sz)
 {
+	LIB_ASSERT_ALWAYS(((sz == 0U) || ((sz & (sz - 1U)) == 0U)), "cbuffer's size has to be a power of 2");
```
ziemleszcz: cbuffer size must be either zero or a power of two

posix/unix.c
```diff
@@ -1100,6 +1100,8 @@ static int unix_bufferSetSize(unixsock_t *s, int sz)
 {
 	void *v[2] = { NULL, NULL };
 	size_t size = (size_t)sz;
+	/* round size to the next power of 2 */
+	size = (size <= 1U) ? 1U : ((size_t)1U << (hal_cpuGetLastBit(size - 1U) + 1U));
```
ziemleszcz: I’m not sure we should adjust values where `sz < US_MIN_BUFFER_SIZE`. It might be better to return `EINVAL` in this case. The check should also be performed on `sz` to catch negative values (previously, `size_t` was signed). Could you verify how Linux handles this?

Additionally, I’m not sure whether overflow is handled correctly (when `sz` already has the highest bit set). Does this depend on the definition of size_t?

lib/cbuffer.c
julianuziemblo: Isn't this the current condition?

ziemleszcz: ASSERT message.

posix/unix.c
julianuziemblo: key takeaways from tests on my linux (my machine: `6.8.0-110-generic`): 
- setsockopt never fails for buffer values, just rounds to the next value
- the value doesn't have to be a power of 2
- the value set is exactly the value requested times 2
- if the requested value times 2 is too large, the value is max value
- if the requested value itself if 2 small but requested value times 2 is sufficient, it stays

```
buflen = 2147483647
SO_RCVBUF = 425984

buflen = 425983
SO_RCVBUF = 425984

buflen = 3000
SO_RCVBUF = 6000

buflen = 50000
SO_RCVBUF = 100000

buflen = 128
SO_RCVBUF = 2304

buflen = 2303
SO_RCVBUF = 4606

buflen = 2304
SO_RCVBUF = 4608

buflen = 1152
SO_RCVBUF = 2304

buflen = 1151
SO_RCVBUF = 2304

buflen = 1153
SO_RCVBUF = 2306

buflen = -1
SO_RCVBUF = 425984

buflen = 0
SO_RCVBUF = 2304
```

code used:
```c
#include <stdio.h>
#include <limits.h>
#include <sys/socket.h>
#include <sys/un.h>
#include <unistd.h>


static void test_sockopt(int fd, int buflen)
{
	printf("buflen = %d\n", buflen);
	socklen_t optlen = sizeof(buflen);
	int res = setsockopt(fd, SOL_SOCKET, SO_RCVBUF, &buflen, sizeof(buflen));
	if (res < 0) {
		perror("setsockopt(SO_RCVBUF, -1)");
	}
	else {
		res = getsockopt(fd, SOL_SOCKET, SO_RCVBUF, &buflen, &optlen);
		if (res < 0) {
			perror("getsockopt(SO_RCVBUF)");
		}
		else {
			printf("SO_RCVBUF = %d\n", buflen);
		}
	}
	printf("\n");
}


int main(void)
{
	int fd = socket(AF_UNIX, SOCK_STREAM, 0);
	if (fd < 0) {
		perror("socket(AF_UNIX, SOCK_STREAM, 0)");
		return 1;
	}

	test_sockopt(fd, INT_MAX);
	test_sockopt(fd, 425983);
	test_sockopt(fd, 3000);
	test_sockopt(fd, 50000);
	test_sockopt(fd, 128);
	test_sockopt(fd, 2303);
	test_sockopt(fd, 2304);
	test_sockopt(fd, 2304 / 2);
	test_sockopt(fd, 2304 / 2 - 1);
	test_sockopt(fd, 2304 / 2 + 1);
	test_sockopt(fd, -1);
	test_sockopt(fd, 0);

	return 0;
}
```

julianuziemblo: This seems in line with the POSIX specification as EINVAL is reported to be propagated only if "The specified option is invalid at the specified socket level or the socket has been shut down.".

https://pubs.opengroup.org/onlinepubs/009619199/setsockopt.htm

julianuziemblo: It seems that doubling the value is what the linux kernel does to keep more metadata in the buffer. I don't think we should double the data, but removing the `if` and just keeping the buffer size in to the correct range (and a power of 2) seems like a good idea. What do you think?

ziemleszcz (approved):

PR #762, opened 2026-04-28
posix/open: fix file mode checking

### LINKED ISSUE #1435: `ioctl()` on wrong device returns `EINVAL` instead of appropriate error (`ENOTTY`, `ENOTSOCK`, etc.)

julianuziemblo:
This PR fixes nightly CI with a workaround solution-based for now.

First, check the device itself if it reports its mode. If not, check the filesystem node.
Then, check if the fifo reported is a pipe or other FIFO file.

Fixes: phoenix-rtos/phoenix-rtos-project#1435

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/phoenix-rtos-posixsrv/pull/30
    - 

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -667,40 +667,46 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
 					break;
 				}
 
-				if (proc_lookup("/dev/posix/pipes", NULL, &srv) >= 0) {
-					if (oid.port == srv.port) {
-						f->type = ftPipe;
+				/*
+				 * first, check if the device reports its mode
+				 * if not: check the filesystem node
+				 * finally: report files that don't report attr at all as regular files
+				 */
+				if (posix_getAttr(f->oid, atMode, &attr) < 0) {
+					if (posix_getAttr(f->ln, atMode, &attr) < 0) {
+						f->type = ftRegular;
 						break;
 					}
 				}
 
-				if (posix_getAttr(f->ln, atMode, &attr) < 0) {
-					/* report files that don't report attr as regular files */
-					f->type = ftRegular;
-					break;
+				mode = (mode_t)attr;
+				if (S_ISCHR(mode)) {
+					/*
+					 * FIXME: for now, we treat every character device as
+					 * a TTY. This behavior should be fixed as it changes
+					 * the behavior of e.g. posix_fstat() and ioctl().
+					 */
+					f->type = ftTty;
 				}
-				else {
-					mode = (mode_t)attr;
-					if (S_ISCHR(mode)) {
-						/*
-						 * FIXME: for now, we treat every character device as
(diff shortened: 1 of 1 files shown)


posix/posix.c
```diff
@@ -667,40 +667,40 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
...
-						 * FIXME: for now, we treat every character device as
-						 * a TTY. This behavior should be fixed as it changes
-						 * the behavior of e.g. posix_fstat() and ioctl().
-						 */
-						f->type = ftTty;
-					}
-					else if (S_ISFIFO(mode)) {
-						f->type = ftFifo;
+				else if (S_ISFIFO(mode)) {
+					if (proc_lookup("/dev/posix/pipes", NULL, &srv) >= 0 && oid.port == srv.port) {
```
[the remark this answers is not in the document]
julianuziemblo: yeah, this kind of is outside the scope as we don't really check EINTR anywhere else in kernel - at least I don't see it

posix/posix.c
```diff
@@ -667,40 +667,40 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
 					break;
 				}
 
-				if (proc_lookup("/dev/posix/pipes", NULL, &srv) >= 0) {
-					if (oid.port == srv.port) {
-						f->type = ftPipe;
+				/* first, check if the device reports its mode */
```
etiaro: I would merge this comment into the one inside if

posix/posix.c
etiaro, line 696: Once TTY/CHR distinction is made, the same mechanism could be used to differenciate Pipes and Fifo and avoid this problematic oid compare. Comment about that here could be nice.

etiaro (commented):
Okay as a hotfix, some more general solution could be introduced here in the future.
Nice to check that no more nightly fails are happening before merging (either locally or here by temporarily changing a CI config on this PR)

julianuziemblo:
a proper solution for the kernel to be aware of driver types is needed - this only introduces further confusion

PR #763, opened 2026-04-29
hal: compile hal_timerSetWakeup only when implemented

julianuziemblo:
_threads_updateWakeup (the only user of hal_timerSetWakeup) does anything only if hal_timerSetWakeup is implemented to do anything. Else, it just queries the tree for no reason. This commit introduces CPU_SUPPORTS_TIMER_WAKEUP symbol to define _threads_updateWakeup only if the wakeup is suppored to eliminate unnecessary tree searches

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/aarch64/zynqmp/config.h
+++ b/hal/aarch64/zynqmp/config.h
@@ -26,6 +26,8 @@
 #define SIZE_INTERRUPTS 188U
 #define TIMER_IRQ_ID    68U
 
+#define CPU_SUPPORTS_TIMER_WAKEUP 0 /* if the function hal_timerSetWakeup is supported on this architecture */
+
 #ifndef __ASSEMBLY__
 
 #define HAL_NAME_PLATFORM "Xilinx Zynq Ultrascale+ "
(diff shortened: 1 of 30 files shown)


agkaminski:
This change is kinda against our generic guidelines, i.e. only introduce `ifdef`s if there's no other way

julianuziemblo:
> This change is kinda against our generic guidelines, i.e. only introduce `ifdef`s if there's no other way

Technically there is no other way to avoid "mindless" tree searches, but I get what you mean. Still, I'd argue we should keep this change or a similar one that removes the unneeded action.

nalajcie:
> > This change is kinda against our generic guidelines, i.e. only introduce `ifdef`s if there's no other way
> 
> Technically there is no other way to avoid "mindless" tree searches, but I get what you mean. Still, I'd argue we should keep this change or a similar one that removes the unneeded action.

`-flto` would take care of that (cross-compilation-unit dead code elimination) if the function really doesn't have side effects (sometimes gcc might get confused about some code, we might want to make it a little easier for him by using eg. `__attribute__((pure))` in the future).

PR #764, opened 2026-04-30
hal/arm: optimize and unify spinlock implementation

### LINKED ISSUE #567: kernel: arm: spinlock: r1 used in asm block even if a function has 2 arguments
- Applies to all supported arm architectures
- Example: https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/428d509d25d24967a4d88235effe0d7390a6677a/hal/armv7m/spinlock.c#L24-L60
 - As @kemonats suggested we can switch to using r2, r3.
 - Above that placing cc in clobbers for both functions should be applied.

### LINKED ISSUE #1311: kernel/hal/aarch64/arch/cpu.h  the wrong register was used
The `result` is of type u32 so should be:
```c
static inline u32 hal_cpuAtomicGet(volatile u32 *dst)
{
	u32 result;
	/* clang-format off */
	__asm__ volatile (
		"ldar %w0, [%1]\n"
		: "=r"(result)
		: "r"(dst)
	);
	/* clang-format on */
	return result;
}
```

jmaksymowicz:
## Description

Until now, implementations of spinlocks on different ARM platforms were copy-pasted from each other, sometimes with minor changes - a total of 5 implementations. In reality, only one is needed (with minor changes for Cortex-M targets). 

The old implementation was inefficient, especially when releasing a lock - it performed atomic add operation, when a store was sufficient (the only values the lock needs to take are 0 for "taken" and 1 for "free"). The `lock` status byte was moved to the start of the `spinlock_t` structure, so that `packed` attribute that was in armv7a, armv7r and armv8r implementations is no longer needed. The new implementation also reduces the number of registers that need to be used and fixes `"cc"` not being in clobber list in `hal_spinlockClear` (by not using status flags anymore).

The new spinlock implementation aligns with the example outlined in ARM Architecture Reference Manual (DDI 0406C.c, sections D7.3.1 "Acquiring a lock" and D7.3.2 "Releasing a lock") and should be safe for all ARM targets. 

On ARMv8 targets the implementation can be made more efficient by the use of load-acquire and store-release instructions - however, when tested on ARMv8-M it yielded no measurable improvement. Perhaps on multi-core targets the benefits would be larger, but currently we don't support any ARMv8-A targets in AArch32 mode. For this reason, I decided to not complicate this implementation more than necessary for now.

Unfortunately, I haven't solved the problem that `hal_cpuReschedule` asm implementations on different platforms still use copy-pasted spinlock release code with small modifications. Suggestions on how that could be improved are welcome.

I also fixed the minor issue in `hal_cpuAtomicGet` on AArch64 where a register of wrong width was used when reading.

## Motivation and Context
Closes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/567
Closes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1311
JIRA: RTOS-1304
JIRA: RTOS-1174

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv8m55-stm32n6-nucleo

## Checklist:

## Special treatment


hal/arm/spinlock.c
```diff
@@ -27,21 +32,26 @@ void hal_spinlockSet(spinlock_t *spinlock, spinlock_ctx_t *sc)
...
+#else
+		"mrs r2, cpsr\n"
+		"cpsid if\n"
+#endif
+		"str r2, [%0]\n"
+		"mov %0, #0\n"
+	"1:\n"
+		"ldrexb r2, [%1]\n"
+		"cmp r2, #0\n"
+		"beq 1b\n"
```
ziemleszcz, line 53: Any chance we could use WFE/SEV instead of a tight loop?

Darchiv (commented):
LGTM, deduplication is always welcome.

But, this change impacts all Arm platforms - did you test it exhaustively, e.g. with complete test suite (like nightly), some ports, products etc.?

hal/arm/arch/spinlock.h
```diff
@@ -0,0 +1,32 @@
...
+ * %LICENSE%
+ */
+
+#ifndef _PH_HAL_ARM_SPINLOCK_H_
+#define _PH_HAL_ARM_SPINLOCK_H_
+
+#include "hal/types.h"
+
+typedef struct _spinlock_t {
+	u8 lock; /* Must be first in the structure, it is used in assembly code */
```
ziemleszcz: Verify it using static assert in `spinlock.c` (`offsetof(spinlock_t, lock) == 0`)?

jmaksymowicz:
> LGTM, deduplication is always welcome.
> 
> But, this change impacts all Arm platforms - did you test it exhaustively, e.g. with complete test suite (like nightly), some ports, products etc.?

Unfortunately not, I've only run the default CI campaign. We need to test it thoroughly on SMP targets, because they have the highest likelihood of triggering errors. Then again, in my experience faulty spinlocks tend to crash the system very quickly - a regular campaign on Zynq7000 running to the end is already some evidence that it's stable.

On single-core targets these spinlocks only need to switch off interrupts and (maybe) perform memory barriers, writing to memory that they do is actually superfluous.

hal/arm/spinlock.c
jmaksymowicz: Yes, but it would only have any benefits in SMP systems. On single-core systems they will just be more pointless work.

Perhaps we could have another `#if` so that changing lock state and WFE/SEV is only done on multi-core systems.

ziemleszcz: Agreed.

hal/arm/arch/spinlock.h
jmaksymowicz: Good idea, added a static assert.

hal/arm/spinlock.c
jmaksymowicz: I think we can do it like this - we can merge these changes now to have a unified, basic implementation. In another PR I could try more changes:
1. Spinlocks with WFE/SEV
2. No memory operations if `NUM_CPUS <= 1`
3. Unify implementations of spinlock unlocking in assembly code

hal/arm/spinlock.c
```diff
@@ -3,19 +3,27 @@
...
 #include "hal/list.h"
+#include "lib/lib.h"
+#include "config.h"
 
+#ifndef KERNEL_SPINLOCK_ARM_M
+#error "KERNEL_SPINLOCK_ARM_M must be defined (1 for Cortex-M CPU, 0 for other ARM CPU)"
+#endif
+
+/* `lock` field must be first in the structure because it is used in platform-specific assembly code */
+_Static_assert(offsetof(spinlock_t, lock) == 0);
```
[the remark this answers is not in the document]
jmaksymowicz: Fixed

hal/aarch64/arch/cpu.h
```diff
@@ -187,7 +187,7 @@ static inline u32 hal_cpuAtomicGet(volatile u32 *dst)
 	u32 result;
 	/* clang-format off */
 	__asm__ volatile (
-		"ldar %0, [%1]\n"
+		"ldar %w0, [%1]\n"
 		: "=r"(result)
 		: "r"(dst)
```
ziemleszcz, line 192: Add memory clobber (similar to `hal_cpuAtomicInc`)?

hal/armv7m/imxrt/_init.S
```diff
@@ -299,18 +299,14 @@ hal_cpuReschedule:
 	eor r0, r0, r0 /* default return value */
 	bl _hal_invokePendSV
 	pop {r1-r3, lr}
+	/* If spinlock not NULL, clear it */
 	cmp r1, #0
 	beq hal_cpuReschedule0
 	push {r3-r4}
```
ziemleszcz: `r4` is no longer used.

hal/armv7m/stm32/_init.S
```diff
@@ -267,18 +267,13 @@ hal_cpuReschedule:
 	eor r0, r0, r0 /* default return value */
 	bl _hal_invokePendSV
 	pop {r1-r3, lr}
+	/* If spinlock not NULL, clear it */
 	cmp r1, #0
 	beq hal_cpuReschedule0
 	push {r3-r4}
```
ziemleszcz: `r4` is no longer used.

hal/armv8m/nrf/_init.S
```diff
@@ -265,18 +265,13 @@ hal_cpuReschedule:
 	eor r0, r0, r0 /* default return value */
 	bl _hal_invokePendSV
 	pop {r1-r3, lr}
+	/* If spinlock not NULL, clear it */
 	cmp r1, #0
 	beq hal_cpuReschedule0
 	push {r3-r4}
```
ziemleszcz: `r4` is no longer used.

hal/arm/spinlock.c
```diff
@@ -27,21 +35,26 @@ void hal_spinlockSet(spinlock_t *spinlock, spinlock_ctx_t *sc)
...
-	: "r2", "r3", "memory", "cc");
+#if KERNEL_SPINLOCK_ARM_M
+		"mrs r2, primask\n"
+		"cpsid i\n"
+#else
+		"mrs r2, cpsr\n"
+		"cpsid if\n"
+#endif
+		"str r2, [%0]\n"
+		"mov %0, #0\n"
```
ziemleszcz: Why not `r3` as in the old version?

hal/arm/spinlock.c
```diff
@@ -27,21 +35,26 @@ void hal_spinlockSet(spinlock_t *spinlock, spinlock_ctx_t *sc)
...
+	"1:\n"
+		"ldrexb r2, [%1]\n"
+		"cmp r2, #0\n"
+		"beq 1b\n"
+		"strexb r2, %0, [%1]\n"
+		"cmp r2, #0\n"
+		"bne 1b\n"
+		"dmb\n"
+	: "+r" (sc)
```
ziemleszcz: `sc` will be set to final register value 0.

hal/arm/spinlock.c
```diff
@@ -3,19 +3,27 @@
...
 #include "hal/list.h"
+#include "lib/lib.h"
+#include "config.h"
 
+#ifndef KERNEL_SPINLOCK_ARM_M
+#error "KERNEL_SPINLOCK_ARM_M must be defined (1 for Cortex-M CPU, 0 for other ARM CPU)"
+#endif
+
+/* `lock` field must be first in the structure because it is used in platform-specific assembly code */
+_Static_assert(offsetof(spinlock_t, lock) == 0U);
```
ziemleszcz: Please add `lock field must be first in the structure` message to this assert.

hal/aarch64/arch/cpu.h
jmaksymowicz: Added

hal/armv7m/stm32/_init.S
jmaksymowicz: Removed `r4` from push and pop operations. In the future I'll check if we have some unused register so we can get rid of the push and pop completely.

hal/armv8m/nrf/_init.S
jmaksymowicz: Removed `r4`.

hal/armv7m/imxrt/_init.S
jmaksymowicz: Removed `r4`.

hal/arm/spinlock.c
jmaksymowicz: I decided to rewrite the inline assembly so that it no longer uses fixed register numbers. This gives the register allocator more freedom to pick whatever registers are available.

hal/arm/spinlock.c
jmaksymowicz: Added.

hal/arm/spinlock.c
jmaksymowicz: This wasn't much of a problem, as `sc` was a local variable and wasn't used afterwards. After the rewrite, this is no longer an issue.

Darchiv (approved):

ziemleszcz (approved):

PR #771, opened 2026-05-08
posix: destroy pipe on error

julianuziemblo:
## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: `ia32-generic-qemu`

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/phoenix-rtos-posixsrv/pull/31 
@

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -559,7 +559,7 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
 {
 	TRACE("open(%s, %d, %d)", filename, oflag);
 	oid_t ln, oid, dev, pipesrv;
-	int fd = 0, err = 0;
+	int fd = 0, err = 0, created = 0;
 	process_info_t *p;
 	open_file_t *f;
 	mode_t mode;
@@ -603,6 +603,7 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
 				if (err < 0) {
 					break;
 				}
+				created = 1;
 				hal_memcpy(&ln, &oid, sizeof(oid_t));
 			}
 			else if (err < 0) {
@@ -663,6 +664,12 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
 			pinfo_put(p);
 			return fd;
 		} while (0);
+
+		if (created != 0) {
+			/* file was created in the filesystem - we should unlink it now */
+			(void)posix_unlink(filename);
+			(void)proc_destroy(oid.port, oid);
+		}
 
 		(void)proc_lockSet(&p->lock);
 		p->fds[fd].file = NULL;
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -984,16 +984,16 @@ int posix_pipe(int fildes[2])
 
 	fo = vm_kmalloc(sizeof(open_file_t));
 	if (fo == NULL) {
+		(void)proc_destroy(oid.port, oid);
 		pinfo_put(p);
-		/* FIXME: destroy pipe */
 		return -ENOMEM;
 	}
 
 	fi = vm_kmalloc(sizeof(open_file_t));
 	if (fi == NULL) {
 		vm_kfree(fo);
+		(void)proc_destroy(oid.port, oid);
 		pinfo_put(p);
-		/* FIXME: destroy pipe */
 		return -ENOMEM;
 	}
 
@@ -1008,6 +1008,8 @@ int posix_pipe(int fildes[2])
 
 		vm_kfree(fo);
 		vm_kfree(fi);
+
+		(void)proc_destroy(oid.port, oid);
 
 		pinfo_put(p);
 		return -EMFILE;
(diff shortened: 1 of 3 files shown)


posix/posix.c
```diff
@@ -988,15 +987,15 @@ int posix_pipe(int fildes[2])
 	fo = vm_kmalloc(sizeof(open_file_t));
 	if (fo == NULL) {
 		pinfo_put(p);
-		/* FIXME: destroy pipe */
+		(void)proc_destroy(oid.port, oid);
```
ziemleszcz, line 994: This leaks pipes on EINTR or ENOMEM here and elsewhere. Add a FIXME/TODO comment at minimum.

julianuziemblo: you mean this should be done in `while (proc_destroy(oid.port, oid) == -EINTR);`? or is it a greater issue in your opinion so for now a TODO suffices?

ziemleszcz: `EINTR` handling is a wider kernel problem, so a TODO is acceptable. For `ENOMEM`, I’m not sure we can do much. Still, these cases should be marked with FIXME/TODO comments.

ziemleszcz (dismissed):

oI0ck (dismissed):
If tested then LGTM

posix/posix.c
ziemleszcz, line 994: This does not work for dummyfs. If I force an error at L617, it calls `proc_destroy()`, but the file is not removed. I suspect that unlink is required prior to destruction (unclear API). I have not checked other filesystems.

oI0ck: Shouldn't we move handling `proc_destroy()` to another issue?

ziemleszcz: The "posix/open: destroy file on error if created" commit is supposed to fix this issue. The file is not destroyed. We can remove it from this PR and address pipe/fifo issues only. Still testing it.

posix/posix.c
```diff
@@ -664,6 +665,10 @@ int posix_open(const char *filename, int oflag, u8 *ustack)
 			return fd;
 		} while (0);
 
+		if (created != 0) {
+			(void)proc_destroy(oid.port, oid);
```
ziemleszcz, line 671: This does not work for dummyfs. If I force an error at L617, it calls proc_destroy(), but the file is not removed. I suspect that unlink is required prior to destruction (unclear API). I have not checked other filesystems.

posix/posix.c
ziemleszcz: Sorry, wrong location. The comment above refers to `posix_open()`.

ziemleszcz: I see 4 pages leak on this error path.

julianuziemblo: is this because of the lack of unlink, or something else?

julianuziemblo: about the unlink:
Unlinking it there seems to work (tested on dummyfs and ext2). I'm not sure if I should add this functionality to the proc_destroy itself (if path != NULL then proc_unlink) or should it be done separately here. My intuition is that proc_destroy seems itself as a counterpart to proc_create. 

Another issue is that mtDestroy IMO should undo everything done in mtCreate, but this is a separate implementation and interface definition issue.

For now I propose the simple pushed solution. What do you think?

ziemleszcz, line 1014: filedes[0] remains reserved here if `_posix_allocfd()` fails on filedes[1].

julianuziemblo: hm, is it a leak though? we still have that memory and will use it when we want to open another file if I understand the code correctly?

posix/posix.c
```diff
@@ -1059,12 +1068,15 @@ int posix_mkfifo(const char *pathname, mode_t mode)
 	/* link pipe in posix server */
 	ret = proc_link(oid, oid, pathname);
 	if (ret < 0) {
+		(void)proc_destroy(oid.port, oid);
```
ziemleszcz, line 1071: Same 4 pages leak on this error path as in `posix_pipe()`.

ziemleszcz: FYI: I don't see this leak on next error path at L1078 so it might be connected to `unlink`.

posix/posix.c
ziemleszcz: It is not a memory leak but the fd is reserved and cannot be assigned.

posix/posix.c
oI0ck: How did you manage to manage to trigger this error path, could you please send a repro?

julianuziemblo: strcmp on some path in proc_link will work :D

posix/posix.c
ziemleszcz: You can just set p->fds[filedes[]] to NULL to free it (under the lock).

ziemleszcz: Closing this comment in `posix_open()`. Leaving this thread open for memory leak.

ziemleszcz: Bug in posixsrv.

posix/posix.c
ziemleszcz: Bug in posixsrv.

posix/posix.c
julianuziemblo: are you sure it remains reserved? _posix_allocfd() does not reserve anything, just returns the first empty fd (or allocates new one), and that is zeroed out (hal_memset). I don't see any more memory usage in the tests either (which would be visible if the fds are not marked as empty).

ziemleszcz: Closing, descriptors are reserved in different location.

ziemleszcz (approved):

PR #774, opened 2026-05-11
Makefile: modify the linker script for .bss clearing

rmikielis:
JIRA: RTOS-1051

Different approach for clearing kernel's .bss - instead of creating custom linker scripts
for all targets, modifying default GCC linker scripts.

## Description

CONTEXT: Clearing .bss section requires certain modifications of kernels linker script:
1. moving syspage out of .bss
2. unifying "bss_start" and "bss_end" symbols across platforms, so we know .bss limits.

This PR is sed'ing default linker scripts instead of creating them from scratch. All required var's are in Depends-On PR. This solution sounds safer since in worst-case we just won't add symbols and won't clear bss. Possibility of missing sections in final .ELF seems harder to debug.

After first checks this solution works as expected. Linker script is edited => .elf modified
Screenshot from 2026-05-11 13-10-34
Screenshot from 2026-05-11 13-13-26

Please share your thouhts. Is it way to go or we stick to rewriting ld's from scratch.

Depends-On: phoenix-rtos-build:rmikielis/clearing_bss

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,10 @@ DEPS := $(patsubst %.o, %.c.d, $(OBJS))
 # 4) Remove "+ SIZEOF_HEADERS", which causes inclusion of program headers.
 $(PREFIX_O)$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt:
 	$(SIL)$(LD) $(LDFLAGS_PREFIX)--verbose 2>/dev/null | sed -n '/^==*$$/,/^==*$$/p' | sed '1,1d; $$d' | sed s/"\s*+\s*SIZEOF_HEADERS"// > "$@"
-
+	@printf '%s\n' "$$SYSPAGE_SECTION" > temp_file.txt
+	@sed -i "/^[[:space:]]*$$SYSPAGE_STARTING_POINT*/r temp_file.txt" "$@" && rm temp_file.txt
+	@sed -i "/^[[:space:]]*$$DEFAULT_BSS_START*/i\\  $$CUSTOM_BSS_START" "$@"
+	@sed -i "/^[[:space:]]*$$DEFAULT_BSS_END*/i\\  $$CUSTOM_BSS_END" "$@"
 
 $(PREFIX_PROG)phoenix-$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).elf: $(OBJS) $(PREFIX_O)$(TARGET_FAMILY)-$(TARGET_SUBFAMILY).ldt
 	@mkdir -p $(@D)


nalajcie:
IMHO clearing from `__bss_start` to `_end` is architecture-agnostic, no need to introduce special per-family define's. 

Not sure how syspage is currently passed, so I don have an input on that.

Note that you should not write any temp files in the source repo, only in build folder or some other external place, but I assume this is just a PoC.

rmikielis:
On most architectures ```_end``` is equivalent to end of bss section, but it generally marks end of writable data area, not .bss explicitly. Some architectures can squeeze something in between.
Screenshot from 2026-05-11 16-26-25

In general plo allocates syspage structure in its memory space and passes pointer to it in CPU register. First instruction in kernel's _start() is usually to save this value from register to ```syspage_t *syspage;```, which is ofc in kernel's ```.bss```. So specify'ing ```.syspage``` section is not necessary if we do ```.bss``` clearing before it. Its just a safer approach to save value from register first, than to execute code with it hanging in GPR, or if we decide to move ```.bss``` clearing further than first line of _start().

Moving temp file to different location will be fixed ofc.

rmikielis:
Some targets are initializing syspage or page tables in ```_init.S``` so clearing .bss in ```main.c``` crashes system. Moving these structures out of .bss also causes problems during vm init.
So clearing .bss must be done in ```_init.S``` anyway. Hence no point in defining global bss limits or .syspage.
Closing.

nalajcie:
ACK, no need to unify the symbol names. The per-target symbols provided by linker could still be used in per-target assembly code. 
If beneficial - we could have a (naked) common C function called from assembly to zero-out the area (and provide the args directly in per-target asm code). 

Regarding my other comment - AFAIK right now only arm adds `.noload`  section before `_end` symbol -  which should not be zeroed-out, buuut we're not using this section currently, so probably not a blocker (if we would like to use it, than we would need similar sections for other architectures also).

I'm not sure what to do with syspage, I don't know if it could (accidentally) overlap with kernel bss or not.

rmikielis:
Whole syspage is in kernel's .bss and kernel is copying it from plo address space, so without moving syspage out of kernel's .bss we must clear it before syspage is initialized.
Calling C function requires setting up small stack (what is usually done near the end of ```_start()```) so its just a (small piece of) redundant code. Also pointer to plo's syspage structure from which we copy syspage to kernel is passed in GPR. But since it is ```r9``` on ARM targets, we should be able to clear .bss w/o preserving it earlier.
So modifying ```_init.S``` of every target is inevitable, no point in playing with linker script for now (but it can be done w/o writing it from scratch if suddenly needed)

PR #776, opened 2026-05-12
process: copy argv and envp to newly created process

### LINKED ISSUE #1399: `ps` CMD column not printed properly
The issue first appeared on the 5af8b11 project commit when `dummyfs` CMD appeared with arguments, and the 13th entry from `ps` had no CMD name.

Image

Up to that day, only the command name was always printed, without any arguments:

Image

Looking into the `ps` code, it shows that it should always print up to 16 characters, so it seems like the `threadsinfo` syscall doesn't copy the full command with arguments into the `info` structure - only the command name. And this one time, it worked correctly for `dummyfs`, but for one of the entries in `info`, the command was empty.

The strange thing is that it happened only once and it seems to work randomly whereas one would expect `threadsinfo` syscall to always produce the same result.

oI0ck:
JIRA: RTOS-1319
Fixes: phoenix-rtos-project#1399

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
- [X] Already covered by automatic testing.

## Checklist:

## Special treatment

- [X] This PR needs additional PRs to work
- https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/480

### DIFF
--- a/proc/process.c
+++ b/proc/process.c
@@ -1518,9 +1518,24 @@ static int process_copy(void)
 	process_t *process = current->process;
 	parent = spawn->parent;
 
+	/* these resources should get cleaned up by process_destroy */
 	if (parent->process->path != NULL) {
 		process->path = lib_strdup(parent->process->path);
 		if (process->path == NULL) {
+			return -ENOMEM;
+		}
+	}
+
+	if (parent->process->argv != NULL) {
+		process->argv = proc_copyargs(parent->process->argv);
+		if (process->argv == NULL) {
+			return -ENOMEM;
+		}
+	}
+
+	if (parent->process->envp != NULL) {
+		process->envp = proc_copyargs(parent->process->envp);
+		if (process->envp == NULL) {
 			return -ENOMEM;
 		}
 	}


proc/process.c
```diff
@@ -1525,6 +1525,13 @@ static int process_copy(void)
 		}
 	}
 
+	if (parent->process->argv != NULL) {
+		process->argv = proc_copyargs(parent->process->argv);
+		if (process->argv == NULL) {
+			return -ENOMEM;
+		}
+	}
```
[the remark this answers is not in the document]
julianuziemblo: I see you added this to the commit. Please also add it to commit message that you copy env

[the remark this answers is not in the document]
oI0ck: done

proc/process.c
```diff
@@ -1544,6 +1525,20 @@ static int process_copy(void)
 		}
 	}
 
+	if (parent->process->argv != NULL) {
+		process->argv = proc_copyargs(parent->process->argv);
+		if (process->argv == NULL) {
+			return -ENOMEM;
```
ziemleszcz, line 1532: Memory allocated earlier in this function is not freed on error paths. Please free `process->path`, `process->argv`, `process->envp`, and `process->map`.

oI0ck: `process->map` is not a pointer, but the rest I will add to error paths

ziemleszcz: `vm_mapDestroy()` destroys the map. Please check if it should be destroyed here.

oI0ck: `vm_mapCopy()` seems to destory the map at least in one error path.

oI0ck: Even so, this is not the `if` statement for destoying the map as it precedes `vm_mapCreate()`.

ziemleszcz: This comment applies to all error paths in this function. I just didn’t want to copy and paste it in three different places.

Please test these error paths and verify that they do not leak memory.

oI0ck: The memory map seems to get destroyed properly in this function

ziemleszcz: It turns out that `process_destroy()` already frees path, argv, and envp, so we now have a double-free issue.

This is not obvious from `process_copy()` implementation, so either:
 - set these pointers to NULL after freeing them, or
 - leave cleanup to `process_destroy()` and add a comment here explaining that it releases these resources.

ziemleszcz (approved):

PR #777, opened 2026-05-13
proc/name: fix race in proc_portRegister()

ziemleszcz:
YT: RTOS-1321

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/proc/name.c
+++ b/proc/name.c
@@ -29,11 +29,11 @@ typedef struct _dcache_entry_t {
 
 
 static struct {
-	int root_registered;
-	oid_t root_oid;
+	int rootRegistered;
+	oid_t rootOid;
 
 	dcache_entry_t *dcache[0x1U << HASH_LEN];
-	lock_t dcache_lock;
+	lock_t lock;
 } name_common;
 
 
@@ -76,19 +76,19 @@ int proc_portRegister(u32 port, const char *name, oid_t *oid)
 	unsigned int hash = dcache_strHash(name);
 
 	/* Check if entry already exists */
-	(void)proc_lockSet(&name_common.dcache_lock);
+	(void)proc_lockSet(&name_common.lock);
 	if (_dcache_entryLookup(hash, name) != NULL) {
-		(void)proc_lockClear(&name_common.dcache_lock);
+		(void)proc_lockClear(&name_common.lock);
 		return -EEXIST;
 	}
-	(void)proc_lockClear(&name_common.dcache_lock);
+	(void)proc_lockClear(&name_common.lock);
 
 	if (name[0] == '/' && name[1] == '\0') {
-		name_common.root_oid.port = port;
+		name_common.rootOid.port = port;
 		if (oid != NULL) {
-			name_common.root_oid.id = oid->id;
+			name_common.rootOid.id = oid->id;
 		}
-		name_common.root_registered = 1;
+		name_common.rootRegistered = 1;
 		return EOK;
 	}
 
@@ -104,10 +104,10 @@ int proc_portRegister(u32 port, const char *name, oid_t *oid)
 
 	(void)hal_strcpy(entry->name, name);
 
-	(void)proc_lockSet(&name_common.dcache_lock);
+	(void)proc_lockSet(&name_common.lock);
 	entry->next = name_common.dcache[hash];
 	name_common.dcache[hash] = entry;
-	(void)proc_lockClear(&name_common.dcache_lock);
+	(void)proc_lockClear(&name_common.lock);
 
 	return EOK;
 }
@@ -118,7 +118,7 @@ int proc_portUnregister(const char *name)
 	dcache_entry_t *entry, *prev = NULL;
 	unsigned int hash = dcache_strHash(name);
 
-	(void)proc_lockSet(&name_common.dcache_lock);
+	(void)proc_lockSet(&name_common.lock);
 	entry = name_common.dcache[hash];
 
 	while (entry != NULL && hal_strcmp(entry->name, name) != 0) {
@@ -129,7 +129,7 @@ int proc_portUnregister(const char *name)
 
 	if (entry == NULL) {
 		/* There is no such entry, nothing to do */
-		(void)proc_lockClear(&name_common.dcache_lock);
+		(void)proc_lockClear(&name_common.lock);
 		return -ENOENT;
 	}
 
@@ -139,7 +139,7 @@ int proc_portUnregister(const char *name)
 	else {
 		name_common.dcache[hash] = entry->next;
 	}
-	(void)proc_lockClear(&name_common.dcache_lock);
+	(void)proc_lockClear(&name_common.lock);
 
 	vm_kfree(entry);
 
@@ -161,13 +161,13 @@ int proc_portLookup(const char *name, oid_t *file, oid_t *dev)
 	}
 
 	if (name[0] == '/' && name[1] == '\0') {
-		if (name_common.root_registered != 0) {
+		if (name_common.rootRegistered != 0) {
 			if (file != NULL) {
-				*file = name_common.root_oid;
+				*file = name_common.rootOid;
 			}
 
 			if (dev != NULL) {
-				*dev = name_common.root_oid;
+				*dev = name_common.rootOid;
 			}
 			return EOK;
 		}
@@ -176,7 +176,7 @@ int proc_portLookup(const char *name, oid_t *file, oid_t *dev)
 	}
 
 	/* Search cache for full path */
-	(void)proc_lockSet(&name_common.dcache_lock);
+	(void)proc_lockSet(&name_common.lock);
 	entry = _dcache_entryLookup(dcache_strHash(name), name);
 	if (entry != NULL) {
 		if (file != NULL) {
@@ -186,12 +186,12 @@ int proc_portLookup(const char *name, oid_t *file, oid_t *dev)
 		if (dev != NULL) {
 			*dev = entry->oid;
 		}
-		(void)proc_lockClear(&name_common.dcache_lock);
+		(void)proc_lockClear(&name_common.lock);
 		return EOK;
 	}
-	(void)proc_lockClear(&name_common.dcache_lock);
+	(void)proc_lockClear(&name_common.lock);
 
-	srv = name_common.root_oid;
+	srv = name_common.rootOid;
 
 	/* Search cache for starting point */
 	len = hal_strlen(name);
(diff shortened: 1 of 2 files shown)


proc/name.c
```diff
@@ -72,18 +72,28 @@ static dcache_entry_t *_dcache_entryLookup(unsigned int hash, const char *name)
 
 int proc_portRegister(u32 port, const char *name, oid_t *oid)
 {
-	dcache_entry_t *entry;
+	dcache_entry_t *entry = NULL;
+	int root = (name[0] == '/' && name[1] == '\0') ? 1 : 0;
 	unsigned int hash = dcache_strHash(name);
```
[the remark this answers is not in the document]
ziemleszcz: The NULL check is already handled by the caller (see `syscalls_sys_portRegister()`).

proc/name.c
```diff
@@ -92,8 +102,8 @@ int proc_portRegister(u32 port, const char *name, oid_t *oid)
 		return EOK;
 	}
 
-	entry = vm_kmalloc(sizeof(dcache_entry_t) + hal_strlen(name) + 1U);
 	if (entry == NULL) {
+		(void)proc_lockClear(&name_common.dcache_lock);
```
agkaminski: This if may be moved to the start

proc/name.c
```diff
@@ -72,18 +72,28 @@ static dcache_entry_t *_dcache_entryLookup(unsigned int hash, const char *name)
 
 int proc_portRegister(u32 port, const char *name, oid_t *oid)
 {
-	dcache_entry_t *entry;
+	dcache_entry_t *entry = NULL;
+	int root = (name[0] == '/' && name[1] == '\0') ? 1 : 0;
```
agkaminski: If you couldn't move the if because of warning (?)
```suggestion
	const int root = (name[0] == '/' && name[1] == '\0') ? 1 : 0;
```

proc/name.c
ziemleszcz: The intention here was to preserve the original EEXIST error.

proc/name.c
```diff
@@ -72,18 +72,28 @@ static dcache_entry_t *_dcache_entryLookup(unsigned int hash, const char *name)
...
 		(void)proc_lockClear(&name_common.dcache_lock);
+		if (entry != NULL) {
+			vm_kfree(entry);
+		}
 		return -EEXIST;
 	}
-	(void)proc_lockClear(&name_common.dcache_lock);
 
-	if (name[0] == '/' && name[1] == '\0') {
+	if (root != 0) {
```
julianuziemblo: can we not:
1) check if root - if yes, register as root
2) else: preallocate entry, calculate hash, set lock

ziemleszcz: No, the root check and registration must be performed under the lock.

julianuziemblo: why should root check be performed under lock? you check it in L76, and then here you clear the lock in the next line anyway. Additionally, you allocate, calculate hash and find entry always, even though you know from there that name is root. The EEXIST check can still be done here by just checking if root_registered == 1.

ziemleszcz: By “root check” I mean verifying if the root entry already exists (`_dcache_entryLookup`). This and the root registration must be done under the lock to avoid races.

julianuziemblo: okay, maybe I cant describe it. Why would this (pseudo) code be wrong in your opinion?

```c
if (name[0] == '/' && name[1] == '\0') {
   if (name_common.root_registered = 1) {
       return -EEXIST;
   }
	name_common.root_oid.port = port;
	if (oid != NULL) {
		name_common.root_oid.id = oid->id;
	}
	name_common.root_registered = 1;
	return EOK;
}

entry = vm_kmalloc(...);
hash = dcache_strHash(name);

(void)proc_lockSet(&name_common.dcache_lock);
if (_dcache_entryLookup(hash, name) != NULL) {
	(void)proc_lockClear(&name_common.dcache_lock);
	return -EEXIST;
}
// ...
```

ziemleszcz: There is a race condition in root registration. However, you are right that this is the same as in the current code. I missed this issue.

julianuziemblo: The other issue then is in `proc_portLookup`, as we don't do it under a lock either

proc/name.c
```diff
@@ -72,28 +72,38 @@ static dcache_entry_t *_dcache_entryLookup(unsigned int hash, const char *name)
...
 		name_common.root_oid.port = port;
 		if (oid != NULL) {
 			name_common.root_oid.id = oid->id;
 		}
 		name_common.root_registered = 1;
+		(void)proc_lockClear(&name_common.dcache_lock);
```
julianuziemblo: this is unfortunately still not safe as `proc_portLookup` checks `name_common.root_registered` without a lock

proc/name.c
```diff
@@ -160,23 +184,26 @@ int proc_portLookup(const char *name, oid_t *file, oid_t *dev)
...
 			if (dev != NULL) {
-				*dev = name_common.root_oid;
+				*dev = name_common.rootOid;
 			}
+			(void)proc_lockClear(&name_common.lock);
 			return EOK;
 		}
 
+		(void)proc_lockClear(&name_common.lock);
 		return -EINVAL;
```
julianuziemblo: `ENOENT`?

proc/name.c
```diff
@@ -29,11 +29,12 @@ typedef struct _dcache_entry_t {
...
 static struct {
-	int root_registered;
-	oid_t root_oid;
+	int rootRegistered;
+	oid_t rootOid;
 
 	dcache_entry_t *dcache[0x1U << HASH_LEN];
-	lock_t dcache_lock;
+
+	lock_t lock;
```
julianuziemblo, line 37: potentially these refactors should be made in separate commit (seems to me they should as there are multiple lines where the only change is the name change `dcache_lock` -> `lock`)

proc/name.c
```diff
@@ -73,41 +74,55 @@ static dcache_entry_t *_dcache_entryLookup(unsigned int hash, const char *name)
...
 		return -ENOMEM;
 	}
 
 	entry->oid.port = port;
-	if (oid != NULL) {
-		entry->oid.id = oid->id;
-	}
+	entry->oid.id = (oid != NULL) ? oid->id : 0U;
 
 	(void)hal_strcpy(entry->name, name);
```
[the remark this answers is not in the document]
ziemleszcz: This issue is already marked with a FIXME in `syscalls.c`.

proc/name.c
```diff
@@ -222,23 +252,23 @@ int proc_portLookup(const char *name, oid_t *file, oid_t *dev)
...
-		(void)proc_lockClear(&name_common.dcache_lock);
 	}
 
-	if (name_common.root_registered == 0 && i == 0U) {
+	if (name_common.rootRegistered == 0 && i == 0U) {
+		(void)proc_lockClear(&name_common.lock);
 		if (pheap != NULL) {
 			vm_kfree(pheap);
 		}
 		return -EINVAL;
```
julianuziemblo: ENOENT? or does this mean that "it's invalid to call lookup when root is not registered"?

proc/name.c
```diff
@@ -222,23 +252,23 @@ int proc_portLookup(const char *name, oid_t *file, oid_t *dev)
...
 
-		(void)proc_lockSet(&name_common.dcache_lock);
 		entry = _dcache_entryLookup(dcache_strHash(pptr), pptr);
 		if (entry != NULL) {
 			srv = entry->oid;
-			(void)proc_lockClear(&name_common.dcache_lock);
 			break;
 		}
-		(void)proc_lockClear(&name_common.dcache_lock);
 	}
```
julianuziemblo, line 284: maybe not in this commit, but this seems like it could be replaced with `lib_splitname` or at least `lib_strrchr`

proc/name.c
julianuziemblo: ... and while at it, maybe change the name "dcache" to something else, like "dhashmap"? "dcache" suggests that this is used as a cache, which it is not. @agkaminski what would you suggest?

proc/name.c
```diff
@@ -208,6 +234,10 @@ int proc_portLookup(const char *name, oid_t *file, oid_t *dev)
 		pptr = pheap;
 	}
 
+	(void)proc_lockSet(&name_common.lock);
```
julianuziemblo, line 243: I'd state in a comment why unlocking and locking again here is safe

syscalls.c
```diff
@@ -775,8 +775,10 @@ int syscalls_sys_portRegister(u8 *ustack)
 		return -EINVAL;
 	}
 
-	if (vm_mapBelongs(proc, oid, sizeof(*oid)) < 0) {
-		return -EFAULT;
+	if (oid != NULL) {
+		if (vm_mapBelongs(proc, oid, sizeof(*oid)) < 0) {
+			return -EFAULT;
+		}
 	}
```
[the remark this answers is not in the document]
jmaksymowicz: This is pointless. There is no TOCTOU possibility here because the `oid` pointer is a local variable (copied from the user stack) so there is no way for the userspace process to manipulate it after calling the syscall. Perhaps there could be some TOCTOU if there was a check for the contents of `oid->id`, but there isn't. 

One bit of weirdness that may happen is that in `proc_portRegister` we read `oid->id` twice, so on a multi-processor system this could be modified in the middle of the call so potentially root OID wouldn't match the OID in the entry. I don't know how much this could break, though.

proc/name.c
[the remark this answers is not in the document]
jmaksymowicz: This is true - also, since we already have the length of the string passed as argument to the syscall we should use it. Copy with `hal_memcpy` and remember to null-terminate it after copying, just to be sure.

Also, there's a FIXME about this already.

proc/name.c
ziemleszcz: No, these functions are not useful in this case.

proc/name.c
agkaminski: It's still a cache, crippled and effectively turned off, but still. I you want just a register functionality then it can be implemented way simpler than this.

I'm all in for separating these functionalities, using side effect of permanent cache for root registering is not pretty. But it's outside of the scope of this pr

syscalls.c
```diff
@@ -775,8 +775,10 @@ int syscalls_sys_portRegister(u8 *ustack)
 		return -EINVAL;
 	}
 
-	if (vm_mapBelongs(proc, oid, sizeof(*oid)) < 0) {
-		return -EFAULT;
+	if (oid != NULL) {
+		if (vm_mapBelongs(proc, oid, sizeof(*oid)) < 0) {
```
adamgreloch:
```suggestion
	if ((oid != NULL) && (vm_mapBelongs(proc, oid, sizeof(*oid)) < 0)) {
```
for consistency with the rest of similar conditions?

proc/name.c
```diff
@@ -73,40 +74,54 @@ static dcache_entry_t *_dcache_entryLookup(unsigned int hash, const char *name)
...
+		if (entry != NULL) {
+			vm_kfree(entry);
+		}
+		return -EEXIST;
+	}
+
 	if (entry == NULL) {
+		(void)proc_lockClear(&name_common.lock);
 		return -ENOMEM;
 	}
```
adamgreloch, line 115: in case allocation fails but the entry exists, is it really that important that we return `-EEXIST` instead of `-ENOMEM` just after the failed kmalloc? if not, moving this  just below the kmalloc would simplify things a bit

ziemleszcz: I see no reason to change the existing error return behavior. There’s little to gain.

adamgreloch: IMO change is justified - currently you need to care about freeing `entry` in L107 and clearing the lock in L113. Exiting just after L99 frees you of both of these concerns

ziemleszcz: IMHO, my last comment still stands.

adamgreloch: Since this was more of a nit and seems you have a strong preference, do as you wish, however I don't quite see on what basis.

You can drop one code path (false branch of `if (entry != NULL)`) and make the already touched code easier to comprehend pretty much for free... Also I'd prefer userspace to know about kernel ENOMEM-ing as soon as possible, especially not covering it up with EEXIST.

julianuziemblo (approved):

PR #778, opened 2026-05-13
proc/threads: implement force unlocking of locks

### LINKED ISSUE #1535: Deadlock occurs if kernel forcibly unlocks recursive mutex not released by user program
Reproduction (ia32-generic-qemu):
```c
#include <pthread.h>

int main(void) {
	pthread_mutex_t mutex;
	pthread_mutexattr_t attr;

	pthread_mutexattr_init(&attr);
	pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);

	pthread_mutex_init(&mutex, &attr);

	pthread_mutexattr_destroy(&attr);

	pthread_mutex_lock(&mutex);

	return 0;
}
```

It looks like [`_proc_lockUnlock()`](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/cbfa646aa36db35946d37f02320be426222126c3/proc/threads.c#L1833) keeps failing with `-EPERM` because `lock->owner` is a ghost, while `current` has already switched to a different thread.

```c
current = _proc_current();

if ((lock->attr.type == PH_LOCK_ERRORCHECK) || (lock->attr.type == PH_LOCK_RECURSIVE)) {
	if (lock->owner != current) {
		return -EPERM;
	}
}
```

As a result, the thread’s lock list is never cleared. This causes an infinite loop in [`thread_destroy()`.](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/cbfa646aa36db35946d37f02320be426222126c3/proc/threads.c#L495) 
```c
while (thread->locks != NULL) {
	proc_lockUnlock(thread->locks);
}
```

jmaksymowicz:
TODO: build fails on sparcv8leon-gr716-mimas because kernel is now too big
```
/opt/phoenix/sparc-phoenix/bin/../lib/gcc/sparc-phoenix/14.2.0/../../../../sparc-phoenix/bin/ld: section .got VMA [40001a08,40001a0f] overlaps section .bss VMA [40001a00,40002bef]
```

Fix infinite loop that could have occurred in `thread_destroy()` function if trying to unlock a lock held by a thread failed. 
Fix possible use-after-free after `proc_lockDone()` if unlocking a lock failed.

## Description
Unlocking of locks held by a process is subject to error checking, e.g. if the process unlocking the lock actually held it. This is only for locks of type `PH_LOCK_ERRORCHECK` and `PH_LOCK_RECURSIVE`, which is perhaps the reason why the issue below went unnoticed for some time now.

Problems arise when you consider the `init` process which reaps threads that have ended. If a userspace thread held any such locks, they wouldn't be closed because of error checking (`init` process doesn't actually own any of the userspace locks). The error wasn't propagated and the loop in `thread_destroy()` would spin forever - preventing any other threads from being reaped and possibly deadlocking the system.

Another subtle problem was in `proc_lockDone()`. Code in that function assumed that after calling `_proc_lockUnlock` the lock would be unlocked, removed from the list of locks held by a thread and could be freed. To that end it called `hal_spinlockDestroy()`. However, this could fail not only because of error checking mentioned above, but also when unlocking a recursive mutex held with a depth of > 1. When destroying such a mutex, a use-after-free could occur with hard to predict consequences. 

To fix this, `_proc_lockUnlock` was modified to allow force unlocking a lock, which omits error checks and always succeeds in removing the lock from the list of held locks. To protect against future mistakes, an assertion was added to make it clear the function must succeed at force unlocking.

`thread_destroy()` now force-unlocks all locks that a thread holds and so does `proc_lockDone()`. The latter function does return `int`, so in theory it could try unlocking and return a retcode if failed. However, every single call of that function ignores the retcode, so I chose force unlocking as safer.

## Motivation and Context
Closes https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1535
JIRA: RTOS-1322

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv8m55-stm32n6-nucleo, ia32-generic-qemu

## Checklist:

## Special treatment

### DIFF
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -30,6 +30,12 @@
 enum { event_scheduling, event_enqueued, event_waking, event_preempted };
 /* clang-format on */
 
+#define UNLOCK_DONT_YIELD 0
+#define UNLOCK_DO_YIELD   1
+#define UNLOCK_TRY        0
+#define UNLOCK_FORCE      1
+
+
 const struct lockAttr proc_lockAttrDefault = { .type = PH_LOCK_NORMAL };
 
 /* Special empty queue value used to wakeup next enqueued thread. This is used to implement sticky conditions */
@@ -239,7 +245,7 @@ static int threads_timeintr(unsigned int n, cpu_context_t *context, void *arg)
  */
 
 
-static void proc_lockUnlock(lock_t *lock);
+static void proc_lockForceUnlock(lock_t *lock, int doYield);
 
 
 static void thread_destroy(thread_t *thread)
@@ -252,7 +258,7 @@ static void thread_destroy(thread_t *thread)
 	/* No need to protect thread->locks access with threads_common.spinlock */
 	/* The destroyed thread is a ghost and no thread (except for the current one) can access it */
 	while (thread->locks != NULL) {
-		proc_lockUnlock(thread->locks);
+		proc_lockForceUnlock(thread->locks, UNLOCK_DO_YIELD);
 	}
 	vm_kfree(thread->kstack);
 
@@ -1610,7 +1616,7 @@ int proc_lockSetInterruptible(lock_t *lock)
 }
 
 
-static int _proc_lockUnlock(lock_t *lock)
+static int _proc_lockUnlock(lock_t *lock, int doForceUnlock)
 {
 	thread_t *owner = lock->owner, *current;
 	spinlock_ctx_t sc;
@@ -1626,18 +1632,25 @@ static int _proc_lockUnlock(lock_t *lock)
 	LIB_ASSERT(LIST_BELONGS(&owner->locks, lock) != 0, "lock: %s, owner pid: %d, owner tid: %d, lock is not on the list",
 			lock->name, (owner->process != NULL) ? process_getPid(owner->process) : 0, proc_getTid(owner));
 
-	if ((lock->attr.type == PH_LOCK_ERRORCHECK) || (lock->attr.type == PH_LOCK_RECURSIVE)) {
-		if (lock->owner != current) {
-			hal_spinlockClear(&threads_common.spinlock, &sc);
-			return -EPERM;
+	if (doForceUnlock == UNLOCK_TRY) {
+		if ((lock->attr.type == PH_LOCK_ERRORCHECK) || (lock->attr.type == PH_LOCK_RECURSIVE)) {
+			if (lock->owner != current) {
+				hal_spinlockClear(&threads_common.spinlock, &sc);
+				return -EPERM;
+			}
 		}
 	}
 
 	if ((lock->attr.type == PH_LOCK_RECURSIVE) && (lock->depth > 0U)) {
-		lock->depth--;
-		if (lock->depth != 0U) {
-			hal_spinlockClear(&threads_common.spinlock, &sc);
-			return 0;
+		if (doForceUnlock == UNLOCK_TRY) {
+			lock->depth--;
+			if (lock->depth != 0U) {
+				hal_spinlockClear(&threads_common.spinlock, &sc);
+				return 0;
+			}
+		}
+		else {
+			lock->depth = 0U;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


proc/threads.c
```diff
@@ -1660,29 +1666,34 @@ static int _proc_lockUnlock(lock_t *lock)
...
-		hal_spinlockClear(&lock->spinlock, &sc);
+	hal_spinlockClear(&lock->spinlock, &sc);
+	if ((ret > 0) && (doReschedule != 0)) {
 		(void)hal_cpuReschedule(NULL, NULL);
 	}
-	else {
-		hal_spinlockClear(&lock->spinlock, &sc);
-	}
+
+	LIB_ASSERT(ret >= 0, "lock: %s, force unlocking failed with code %d", lock->name, ret);
```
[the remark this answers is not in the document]
jmaksymowicz: This is really a check against kernel programmer's error, so really this should never be triggered. I made it an assertion because IMO it doesn't need a runtime check for non-debug builds.

proc/threads.c
```diff
@@ -239,7 +239,7 @@ static int threads_timeintr(unsigned int n, cpu_context_t *context, void *arg)
  */
 
 
-static void proc_lockUnlock(lock_t *lock);
+static void proc_lockForceUnlock(lock_t *lock, int doReschedule);
```
agkaminski: Just a personal opinion - I'm not a fan of modifier flags as their purpose is not apparent at all at usage.
```
proc_lockForceUnlock(lock, 0);
proc_lockForceUnlock(lock, 1);
```
does not convey that we want reschedule or not. Perhaps
```
proc_lockForceUnlock(lock);
proc_lockForceUnlockYield(lock);
```
It's not ideal too, as it is not apparent that `proc_lockForceUnlock` does not yield in this case, and symbol length becomes way too long.

So the current version is ok with me, just a rant, that perhaps something may be done in this matter. Maybe give names to the nameless values (e.g. `#define LOCK_YIELD 1`)?

proc/threads.c
```diff
@@ -1627,14 +1627,20 @@ static int _proc_lockUnlock(lock_t *lock)
 			lock->name, (owner->process != NULL) ? process_getPid(owner->process) : 0, proc_getTid(owner));
 
 	if ((lock->attr.type == PH_LOCK_ERRORCHECK) || (lock->attr.type == PH_LOCK_RECURSIVE)) {
-		if (lock->owner != current) {
+		if ((doForceUnlock == 0) && (lock->owner != current)) {
```
agkaminski: Perhaps move `(doForceUnlock == 0)` check one or two levels higher

proc/threads.c
```diff
@@ -1627,14 +1627,20 @@ static int _proc_lockUnlock(lock_t *lock)
...
 
 	if ((lock->attr.type == PH_LOCK_RECURSIVE) && (lock->depth > 0U)) {
-		lock->depth--;
+		if (doForceUnlock == 0) {
+			lock->depth--;
+		}
+		else {
+			lock->depth = 0U;
+		}
+
```
agkaminski: Lets integrate the flag more, no need to check if below if doForceUnlock
```
if (doForceUnlock != 0) {
  lock->depth = 0U;
}
else {
  lock->depth--;
  if (lock->depth != 0U) {
```

proc/threads.c
```diff
@@ -1660,29 +1666,34 @@ static int _proc_lockUnlock(lock_t *lock)
...
 	_proc_threadSetPriority(owner, _proc_threadGetPriority(owner));
 
-	LIB_ASSERT(current->priority <= current->priorityBase, "pid: %d, tid: %d, basePrio: %d, priority degraded (%d)",
-			(current->process != NULL) ? process_getPid(current->process) : 0, proc_getTid(current), current->priorityBase,
-			current->priority);
+	if (doForceUnlock == 0) {
+		LIB_ASSERT(current->priority <= current->priorityBase, "pid: %d, tid: %d, basePrio: %d, priority degraded (%d)",
+				(current->process != NULL) ? process_getPid(current->process) : 0, proc_getTid(current), current->priorityBase,
+				current->priority);
+	}
```
agkaminski:
```suggestion
	LIB_ASSERT((doForceUnlock == 0) || (current->priority <= current->priorityBase), "pid: %d, tid: %d, basePrio: %d, priority degraded (%d)",
				(current->process != NULL) ? process_getPid(current->process) : 0, proc_getTid(current), current->priorityBase,
				current->priority);
```

agkaminski (commented):
There're no errors, but introduction of these two flags makes the code look like it needs refactoring. Force unlock is it's own operation that can be performed simpler than normal unlock - we just need to extract priority adjustment instead. If you go this way instead then you no longer need `doReschedule` flag anymore

proc/threads.c
jmaksymowicz: I agree, I don't like these either. Would be nice if C had type checking for enums or named parameters or something. Out of the options I think `#define`s for values look best.

proc/threads.c
jmaksymowicz: I decided to restore the assertion to how it was before - force unlocking shouldn't cause priority degradation.

proc/threads.c
jmaksymowicz: Done

proc/threads.c
jmaksymowicz: Done

jmaksymowicz:
To fix building on `sparcv8leon-gr716-mimas` I decided to make the assertion message a bit shorter... yes I know I'm kicking the can down the road but it would be nice to get these bug fixes merged today.

Darchiv (approved):

agkaminski (approved):

PR #779, opened 2026-05-14
posix: file descriptor flags are not shared with the receiver

ziemleszcz:
YT: RTOS-1323

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/posix/fdpass.c
+++ b/posix/fdpass.c
@@ -121,7 +121,8 @@ int fdpass_pack(fdpack_t **packs, const void *control, socklen_t controllen)
 				return err;
 			}
 
-			FDPACK_PUSH(pack, file, 0); /* FIXME: copy flags? */
+			/* file descriptor flags are not shared with the receiver */
+			FDPACK_PUSH(pack, file, 0);
 
 			cmsg_data += sizeof(int);
 			--cnt;


posix/fdpass.c
```diff
@@ -121,7 +121,8 @@ int fdpass_pack(fdpack_t **packs, const void *control, socklen_t controllen)
 				return err;
 			}
 
-			FDPACK_PUSH(pack, file, 0); /* FIXME: copy flags? */
+			/* file descriptor flags are not shared with the receiver */
```
adamgreloch, line 124: Could you explain why the FIXME can be dropped?

ziemleszcz: We set flags to 0 here (these are file descriptor flags). I have confirmed that it's OK. File descriptor flags (currently FD_CLOEXEC only) are not shared through sendmsg().

adamgreloch: Ok, I confirm that not sharing the fd flags in sendmsg() is a consequence of SCM_RIGHTS behaving like dup() which also does not share the fd flags per POSIX

adamgreloch (approved):

PR #785, opened 2026-05-22
proc/threads: let _proc_lockTry relock a recursive mutex

julianuziemblo:
YT: RTOS-1334

## Description

related: https://github.com/phoenix-rtos/libphoenix/pull/477

Tests:
Depends-On: phoenix-rtos-tests:julianuziemblo/CI-676

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/489

## Checklist:

## Special treatment

### DIFF
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -1466,12 +1466,26 @@ int threads_sigsuspend(unsigned int mask)
 static int _proc_lockTry(thread_t *current, lock_t *lock)
 {
 	if (lock->owner != NULL) {
-		return -EBUSY;
+		if ((lock->attr.type == PH_LOCK_RECURSIVE) && (lock->owner == current)) {
+			/* parasoft-suppress-next-line MISRAC2012-RULE_14_3-ac "False-positive - checking for overflow this way is defined in the standard" */
+			if (lock->depth + 1U == 0U) {
+				/* recursive lock locked too many times */
+				return -EAGAIN;
+			}
+			else {
+				lock->depth++;
+				return EOK;
+			}
+		}
+		else {
+			return -EBUSY;
+		}
 	}
 
 	LIST_ADD(&current->locks, lock);
 
 	lock->owner = current;
+	lock->depth = 1;
 
 	return EOK;
 }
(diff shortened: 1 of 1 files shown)


etiaro (commented):
You could try to share some of this code with _proc_lockSet, to reduce chance of such oversights in the future.

proc/threads.c
```diff
@@ -1522,27 +1534,21 @@ static int _proc_lockSet(lock_t *lock, u8 interruptible, spinlock_ctx_t *scp)
...
+			return ret;
+		}
 
-	if (_proc_lockTry(current, lock) < 0) {
-		/* Lock owner might inherit our priority */
+		LIB_ASSERT(lock->owner != current, "lock: %s, pid: %d, tid: %d, deadlock on itself",
+				lock->name, (current->process != NULL) ? process_getPid(current->process) : 0, proc_getTid(current));
 
 		if (current->priority < lock->owner->priority) {
+			/* Lock owner might inherit our priority */
```
etiaro: this comment syntax ("might") makes more sense in its original position, before the if statement

proc/threads.c
```diff
@@ -1522,27 +1534,21 @@ static int _proc_lockSet(lock_t *lock, u8 interruptible, spinlock_ctx_t *scp)
...
 
-		hal_spinlockClear(&threads_common.spinlock, &sc);
-		_trace_eventLockSetExit(lock, tid, ret);
-		return ret;
-	}
-
-	LIB_ASSERT(lock->owner != current, "lock: %s, pid: %d, tid: %d, deadlock on itself",
-			lock->name, (current->process != NULL) ? process_getPid(current->process) : 0, proc_getTid(current));
+	if (ret < 0) {
+		if (ret == -EAGAIN) {
```
etiaro: You can avoid this `EAGAIN` if statement entirely, if the upper one is changed to 
`ret == -EBUSY`.

etiaro (changes requested):

proc/threads.c
```diff
@@ -1466,12 +1466,25 @@ int threads_sigsuspend(unsigned int mask)
 static int _proc_lockTry(thread_t *current, lock_t *lock)
 {
 	if (lock->owner != NULL) {
-		return -EBUSY;
+		if ((lock->attr.type == PH_LOCK_RECURSIVE) && (lock->owner == current)) {
+			if (((int)lock->depth + 1) == 0) {
```
etiaro: Shouldn't this condition look like this?
```
if ((lock->depth + 1U) == 0U)
```
I suspect that `(int)` got added during misrification, causing overflow check to be at least misleading or even becoming Undefined Behavior

I know this is not your code, but could be fixed while you're at it

proc/threads.c
```diff
@@ -1466,12 +1466,26 @@ int threads_sigsuspend(unsigned int mask)
 static int _proc_lockTry(thread_t *current, lock_t *lock)
 {
 	if (lock->owner != NULL) {
-		return -EBUSY;
+		if ((lock->attr.type == PH_LOCK_RECURSIVE) && (lock->owner == current)) {
```
etiaro, line 1469: Is there a reason for nesting this condition inside of `lock->owner != NULL`?

etiaro (changes requested):

proc/threads.c
julianuziemblo: This way we don't check it everytime we call mutexTry - just if we don't have the lock. Other way could be wasteful as we'd check it every time. If you have a better sugestion here let me know.

etiaro (approved):

PR #787, opened 2026-05-29
hal/armv7a: Adding IO mapping functionality in HAL initialization

jmaksymowicz:
Add IO memory mapping functionality similar to aarch64
Adjust code for imx6ull, zynq7000 targets to use this mapping functionality
Copy syspage to named object in `.bss` section instead of memory mapped past-the-end of `.bss`

## Description
Previously the way memory-mapped IO devices were made available to the kernel on armv7a platforms was very unintuitive. Code within `_init.S` mapped them past-the-end of `.bss` section at arbitrarily chosen offsets, which made it difficult for someone reading the code to understand what was going on.

To solve this, memory-mapped IO is now mapped near the top of the address space (range 0xFFC00000 - 0xFFFF0000) using unused entries in the page table `excptab`. It is done using `_pmap_halMapDevice` function with the same API as `aarch64`, `riscv64` and `sparcv8leon` targets. The physical addresses of peripherals are now declared in C code close to the driver where they are actually used, which helps readability and maintainability.

There are almost 4 MB of address space available through `excptab` which should be more than enough considering our OS has a microkernel design with few built-in drivers.

Additionally, memory for the syspage was also mapped past-the-end of `.bss` - which was not necessary considering that syspage could simply be stored to a named object within `.bss`. After changing this, there is nothing more mapped past `.bss` on armv7a and this unintuitive mechanism could be removed.

Other changes:
* On zynq7000 the entry for exception vectors in `excptab` was removed - it was unused because this target supports Security extensions and thus exception vector address is set using the VBAR register. This doesn't change the functioning of the system.
* The console driver for IMX6ULL was changed so that now it respects the `UART_CONSOLE_KERNEL` definition from `board_config.h`.

## Motivation and Context
Part of the work for https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1500

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a-imx6ull, armv7a-zynq7000-qemu

## Checklist:

## Special treatment

### DIFF
--- a/hal/aarch64/pmap.c
+++ b/hal/aarch64/pmap.c
@@ -890,21 +890,22 @@ void _pmap_preinit(addr_t dtbStart, addr_t dtbEnd)
 /* parasoft-begin-suppress MISRAC2012-RULE_17_1 "stdarg.h required for custom functions that are like printf" */
 void *_pmap_halMapDevice(addr_t paddr, size_t pageOffs, size_t size)
 {
+	size_t maxIdx = TTL_IDX(3U, VADDR_DTB);
 	descr_t attrs = DESCR_VALID | DESCR_TABLE | DESCR_AF | DESCR_ISH | DESCR_PXN | DESCR_UXN | DESCR_ATTR(MAIR_IDX_DEVICE);
-	ptr_t va_start = ((VADDR_MAX - (SIZE_PAGE << 9)) + 1U) + (pmap_common.dev_i * SIZE_PAGE);
+	ptr_t vaStart = ((VADDR_MAX - (SIZE_PAGE << 9)) + 1U) + (pmap_common.dev_i * SIZE_PAGE);
+	size_t span = pageOffs + size;
 	size_t offs;
 
-	if ((pmap_common.dev_i + (size / SIZE_PAGE)) > TTL_IDX(3U, VADDR_DTB)) {
-		return NULL;
-	}
+	LIB_ASSERT(size > 0U, "Attempted to map device of size 0");
+	LIB_ASSERT_ALWAYS((pmap_common.dev_i + ((span + SIZE_PAGE - 1U) / SIZE_PAGE)) < maxIdx, "Out of device address space");
 
-	for (offs = 0; offs < size; offs += SIZE_PAGE) {
-		pmap_common.devices_ttl3[TTL_IDX(3U, va_start + offs)] = DESCR_PA(paddr + offs) | attrs;
+	for (offs = 0U; offs < span; offs += SIZE_PAGE) {
+		pmap_common.devices_ttl3[pmap_common.dev_i] = DESCR_PA(paddr + offs) | attrs;
 		pmap_common.dev_i++;
 	}
 
 	hal_cpuDataSyncBarrier();
-	return (void *)va_start + pageOffs;
+	return (void *)(vaStart + pageOffs);
 }
 
 /* parasoft-end-suppress MISRAC2012-RULE_17_1 */
--- a/hal/armv7a/imx6ull/console.c
+++ b/hal/armv7a/imx6ull/console.c
@@ -15,15 +15,31 @@
 
 #include "hal/console.h"
 #include "hal/cpu.h"
+#include "hal/pmap.h"
 
-#define UART uart1
+#include <board_config.h>
+
+#ifndef UART_CONSOLE_KERNEL
+#define UART_CONSOLE_KERNEL 1
+#endif
+
 
 static struct {
-	volatile u32 *uart1;
-	volatile u32 *uart2;
+	volatile u32 *base;
 	u8 type;
 	u32 speed;
 } console_common;
+
+
+/* TODO: add more UARTs - they may require extra configuration, e.g. IOMUX or clock initialization */
+static const addr_t console_baseAddr[] = {
+	0x02020000U,
+	0x021e8000U,
+};
+
+#define N_UARTS (sizeof(console_baseAddr) / sizeof(console_baseAddr[0]))
+
+_Static_assert((UART_CONSOLE_KERNEL > 0) && ((size_t)UART_CONSOLE_KERNEL <= N_UARTS), "Invalid UART number selected");
 
 
 /* clang-format off */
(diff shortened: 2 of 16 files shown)


hal/armv7a/imx6ull/console.c
```diff
@@ -68,32 +80,31 @@
...
 __attribute__((section(".init"))) void _hal_consoleInit(void)
 {
-	console_common.uart1 = (void *)(((u32)&_end + (3U * SIZE_PAGE) - 1U) & ~(SIZE_PAGE - 1U));
-	console_common.uart2 = (void *)(((u32)&_end + (4U * SIZE_PAGE) - 1U) & ~(SIZE_PAGE - 1U));
+	console_common.base = _pmap_halMapDevice(console_baseAddr[UART_CONSOLE_KERNEL - 1], 0, SIZE_PAGE);
 	console_common.speed = 115200;
 
-	*(console_common.UART + ucr2) &= ~0x1U;
-	while ((*(console_common.UART + uts) & 0x1U) != 0U) {
+	*(console_common.base + ucr2) &= ~0x1U;
```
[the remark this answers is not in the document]
jmaksymowicz: Fixed, `_pmap_halMapDevice` can no longer return NULL.

hal/armv7a/pmap.c
```diff
@@ -583,47 +586,75 @@ int pmap_segment(unsigned int i, void **vaddr, size_t *size, vm_prot_t *prot, vo
...
+	if ((pmap_common.dev_i + (size / SIZE_PAGE)) >= maxIdx) {
+		return NULL;
+	}
 
-	/* Create initial heap */
-	LIB_ASSERT_ALWAYS(pmap_enter(pmap, pmap_common.start, (*vstart), PGHD_WRITE | PGHD_READ | PGHD_PRESENT, NULL) == EOK, "failed to create initial heap");
+	for (offs = 0; offs < size; offs += SIZE_PAGE) {
+		pmap_common.excptab[pmap_common.dev_i] = ((paddr + offs) & 0xfffff000UL) | attrs;
+		pmap_common.dev_i++;
+	}
```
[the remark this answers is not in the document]
jmaksymowicz: It's true, fixed the AArch64 implementation that has the same bug.

hal/armv7a/imx6ull/interrupts.c
ziemleszcz, line 20: Add `"hal/pmap.h"` include for `_pmap_halMapDevice`?

hal/armv7a/zynq7000/interrupts.c
ziemleszcz, line 24: Add `"hal/pmap.h"` include for `_pmap_halMapDevice`?

hal/aarch64/pmap.c
```diff
@@ -890,21 +890,22 @@ void _pmap_preinit(addr_t dtbStart, addr_t dtbEnd)
...
-	for (offs = 0; offs < size; offs += SIZE_PAGE) {
-		pmap_common.devices_ttl3[TTL_IDX(3U, va_start + offs)] = DESCR_PA(paddr + offs) | attrs;
+	for (offs = 0U; offs < span; offs += SIZE_PAGE) {
+		pmap_common.devices_ttl3[pmap_common.dev_i] = DESCR_PA(paddr + offs) | attrs;
 		pmap_common.dev_i++;
 	}
 
 	hal_cpuDataSyncBarrier();
-	return (void *)va_start + pageOffs;
+	return (void *)vaStart + pageOffs;
```
ziemleszcz: `void*` pointer arithmetic. Use `char*` instead.

hal/armv7a/imx6ull/console.c
```diff
@@ -15,34 +15,46 @@
...
-	volatile u32 *uart1;
-	volatile u32 *uart2;
+	volatile u32 *base;
 	u8 type;
 	u32 speed;
 } console_common;
 
 
+/* TODO: add more UARTs - they may require extra configuration, e.g. IOMUX or clock initialization */
+static const addr_t console_baseAddr[] = {
```
ziemleszcz, line 35: I'm not sure if compiler will optimize out this memory. Might be better to simply define `CONSOLE_BASE_ADDR` based on `UART_CONSOLE_KERNEL`.

hal/armv7a/imx6ull/imx6ull.c
```diff
@@ -19,6 +19,15 @@
 #include "hal/spinlock.h"
 #include "include/arch/armv7a/imx6ull/imx6ull.h"
 
+#define CCM_BASE        0x020c4000U
```
ziemleszcz, line 22: Maybe move these definitions to `config.h` (the same for `timer.c`)?

hal/armv7a/imx6ull/console.c
jmaksymowicz: I checked the disassembly and it does get optimized out. 
I want to keep the table because if we decide to add more than two UART options for debugging, the `#if` ladder would get annoyingly long - there are 8 UARTs in total on this platform.

hal/armv7a/imx6ull/imx6ull.c
jmaksymowicz: I decided to move address/IRQ definitions for the timer into `config.h` similar to IMXRT platforms. I left other addresses in the C files with the corresponding driver - this makes sense IMO because we assume that the addresses won't change or need to be configured (unlike something like timer where we could select which timer to use for the kernel).

hal/armv7a/imx6ull/interrupts.c
jmaksymowicz: Done

hal/armv7a/zynq7000/interrupts.c
jmaksymowicz: Done

hal/aarch64/pmap.c
jmaksymowicz: Avoided pointer arithmetic by replacing it with `return (void *)(vaStart + pageOffs)` (`vaStart` is integer type).

ziemleszcz (approved):

Darchiv (approved):

PR #788, opened 2026-06-08
posix/unix: new features for AF_UNIX sockets

julianuziemblo:
* handle MSG_TRUNC and MSG_CTRUNC
* keep track of the address of a UNIX socket
* **TODO**: report the source address in recv{from|msg}
* **TODO**: report the address of the connecting socket in accept

YT: RTOS-1343

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/493

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/libphoenix/pull/509


posix/fdpass.c
```diff
@@ -133,21 +133,25 @@ int fdpass_pack(fdpack_t **packs, const void *control, socklen_t controllen)
...
 		return 0;
 	}
 
+	if (*controllen < CMSG_LEN(sizeof(int))) {
+		return MSG_CTRUNC;
+	}
+
 	p = pinfo_find(process_getPid(proc_current()->process));
 	if (p == NULL) {
 		return -1;
```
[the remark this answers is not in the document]
julianuziemblo: they are delivered, which is confirmed by the linked tests that run on host also

posix/fdpass.c
```diff
@@ -133,21 +133,25 @@ int fdpass_pack(fdpack_t **packs, const void *control, socklen_t controllen)
...
 {
 	process_info_t *p;
-	fdpack_t *pack;
+	fdpack_t *pack, *removed = NULL;
 	struct cmsghdr *cmsg;
 	unsigned char *cmsg_data;
 	open_file_t *file;
 	unsigned int cnt, flags;
-	int fd;
+	int fd, ret;
```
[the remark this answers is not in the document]
julianuziemblo: these 2 are swapped in the loop, effectivelly restoring their values (as they are the "inverse" of each other).

lib/strutil.c
```diff
@@ -65,3 +65,13 @@
 		*slash = '\0';
 	}
 }
+
+
+void lib_unsplitname(char *path, const char *dir)
+{
+	size_t dirlen = hal_strlen(dir);
+
+	if (dirlen != 1) {
```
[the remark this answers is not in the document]
julianuziemblo: ?

posix/unix.c
```diff
@@ -862,29 +952,58 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
+					}
+
+					if (r == NULL) {
+						err = -ECONNRESET;
+					}
+					else {
+						unix_copyName(r, src_addr, src_len);
+					}
+
+					unixsock_put(r);
```
ziemleszcz: NULL pointer dereference (`r == NULL` check above).

posix/unix.c
```diff
@@ -773,14 +801,76 @@ int unix_connect(unsigned int socket, const struct sockaddr *address, socklen_t
...
+	if (s->name == NULL) {
+		return;
+	}
+
+	const socklen_t namelen = (socklen_t)hal_strlen(s->name) + 1U;
+	const socklen_t addrlen = *address_len;
+	const socklen_t familylen = (socklen_t)sizeof(address->sa_family);
+	const socklen_t copylen = min(namelen, addrlen - familylen);
+
+	*address_len = namelen + familylen;
```
ziemleszcz: Overwriting user provided `*address_len`.

posix/unix.c
```diff
@@ -862,29 +952,58 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
+					}
+					else {
+						r = unixsock_get(rid);
+					}
+
+					if (r == NULL) {
+						err = -ECONNRESET;
+					}
+					else {
+						unix_copyName(r, src_addr, src_len);
```
ziemleszcz: For DGRAM, consider storing the address alongside each packet instead of storing only RID.
For connection-oriented sockets, store the address once in the socket structure during connection setup.

posix/unix.c
```diff
@@ -862,29 +952,58 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
+				if (src_addr != NULL && src_len != NULL && *src_len != 0U) {
+					if (s->type != SOCK_DGRAM) {
+						r = unixsock_get_remote(s);
+					}
+					else {
+						r = unixsock_get(rid);
+					}
+
+					if (r == NULL) {
+						err = -ECONNRESET;
```
ziemleszcz: This discards correctly received data. It should return the data first and report the error on the next call.

posix/unix.c
```diff
@@ -862,29 +952,58 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
-			else {
-				/* No action required */
-			}
-			/* TODO: peek control data */
-			if (peek == 0U) {
-				if (err > 0 && control != NULL && controllen != NULL && *controllen > 0U) {
-					(void)fdpass_unpack(&s->fdpacks, control, controllen);
+
+			if (err > 0 && control != NULL && controllen != NULL && *controllen > 0U) {
+				if (fdpass_unpack(&s->fdpacks, control, controllen, peek) == (int)MSG_CTRUNC) {
```
ziemleszcz: I don't think that peek should touch fdpacks.

posix/fdpass.c
```diff
@@ -168,7 +174,9 @@ int fdpass_unpack(fdpack_t **packs, void *control, socklen_t *controllen)
 
 		fd = _posix_addOpenFile(p, file, flags);
 		if (fd < 0) {
-			(void)posix_fileDeref(file);
+			if (peek == 0U) {
```
ziemleszcz: ?

posix/fdpass.c
```diff
@@ -178,16 +186,39 @@ int fdpass_unpack(fdpack_t **packs, void *control, socklen_t *controllen)
 
 		if (pack->cnt == 0U) {
 			LIST_REMOVE(packs, pack);
-			vm_kfree(pack);
+
+			if (peek != 0U) {
+				/* increment the file's reference count */
+				(void)proc_lockSet(&file->lock);
+				++file->refs;
```
ziemleszcz: This is incorrect, but as mentioned in my other comment, I don't think MSG_PEEK should touch file descriptors.

posix/unix.c
```diff
@@ -550,22 +572,28 @@ int unix_bind(unsigned int socket, const struct sockaddr *address, socklen_t add
...
+				}
 				if (s->type == SOCK_DGRAM) {
 					_cbuffer_init(&s->buffer, NULL, 0);
 					vm_kfree(v);
 				}
 				break;
 			}
 
-			s->state |= US_BOUND;
+			lib_unsplitname(s->name, dir);
```
ziemleszcz: `lib_unsplitname` is no-op if dir is one char (for example `./mysocket`).

posix/fdpass.c
julianuziemblo: Wwe only want to deref the file if we actually want it removed from the list on error - while peeking, we want to leave everything as-is, to be "destroyed" by reading.

posix/unix.c
julianuziemblo: It does in Linux, but this behavior is implementation-defined. Do you think it's not worth for us to implement it?

posix/unix.c
julianuziemblo: Yes, do you suggest doing something with it here?

posix/fdpass.c
julianuziemblo: Could you elaborate? If we wanted to implement MSG_PEEK for ancillary data, why would we not want to increment the refcount here?

posix/unix.c
julianuziemblo: Why is it better in your opinion?

posix/unix.c
julianuziemblo: Yes, I thought those calls should always give the full length here and only truncate the address written to the buffer if necessary?

posix/unix.c
ziemleszcz: 1. It also fixes the potential data loss at L997 (`ECONNRESET`).
2. You probably won't need to acquire `unix_common.lock`.

posix/unix.c
ziemleszcz: Maybe I wasn't clear enough. What I meant is that `MSG_PEEK` should not install file descriptors received via `SCM_RIGHTS`.

It may copy ancillary data other than `SCM_RIGHTS`, but it should not remove it from the socket. Is that how it works in Linux?

`MSG_PEEK` shouldn't have any side effects. Repeated `MSG_PEEK` calls (assuming no new data arrives) should return identical results.

Since we currently support only `SCM_RIGHTS` ancillary data, it should effectively be a no-op.

posix/unix.c
ziemleszcz: True. Sorry, I reviewed it quickly and didn't express myself clearly. What I meant was that you overwrite `*address_len` without validating it first. I didn't notice that you copy it to `addrlen`, so that part is OK.

However, the value after validation (`copylen`) won't be correct if `*address_len` < `familylen` (socklen_t is unsigned).

I also don't like the name `addrlen`, as it's too easy to confuse with `address_len`.

posix/fdpass.c
ziemleszcz: OK.

posix/fdpass.c
ziemleszcz: At L192-L194, you implicitly assume there is one file descriptor per pack.

posix/unix.c
ziemleszcz: No idea. It's your function. It clearly doesn't work for `./mysocket`. After calling `lib_unsplitname()`, you will end up with `\0` at `s->name[1]`.

posix/unix.c
julianuziemblo: In linux from what I've tested, `MSG_PEEK` does not remove ancillary data from the socket, however it does install the file descriptors (up to how many fit within the provided `msg_controllen`).

This also now works on Phoenix. 

I agree this way we have a side effect of leaking file descriptors if the caller does not handle the ancillary data properly - which seems to be both POSIX compliant and what Linux does.

Maybe the rest of the team can provide their feedback on if this behavior is something we want to support: @phoenix-rtos/phoenix-core 

<details>
<summary>Vibed code to test this</summary>

```c

#pragma GCC diagnostic ignored "-Wattribute-warning"

#include <sys/socket.h>
#include <sys/un.h>
#include <fcntl.h>
#include <unistd.h>

#include <errno.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

static int recv_fd(int sock, int flags)
{
    char data;
    struct iovec iov = {
        .iov_base = &data,
        .iov_len = 1,
    };

    char control[CMSG_SPACE(sizeof(int))];
    memset(control, 0, sizeof(control));

    struct msghdr msg;
    memset(&msg, 0, sizeof(msg));
    msg.msg_iov = &iov;
    msg.msg_iovlen = 1;
    msg.msg_control = control;
    msg.msg_controllen = sizeof(control);

    ssize_t n = recvmsg(sock, &msg, flags);
    if (n == -1) {
        perror("recvmsg");
        return -1;
    }

    struct cmsghdr *cmsg = CMSG_FIRSTHDR(&msg);
    if (!cmsg) {
        fprintf(stderr, "No control message received\n");
        return -1;
    }

    if (cmsg->cmsg_level != SOL_SOCKET ||
        cmsg->cmsg_type != SCM_RIGHTS) {
        fprintf(stderr, "Unexpected control message\n");
        return -1;
    }

    int fd;
    memcpy(&fd, CMSG_DATA(cmsg), sizeof(fd));

    return fd;
}

static void write_string(int fd, const char *s)
{
    ssize_t len = (ssize_t)strlen(s);

    if (write(fd, s, len) != len) {
        perror("write");
        exit(EXIT_FAILURE);
    }
}

int main(void)
{
    int sv[2];

    if (socketpair(AF_UNIX, SOCK_DGRAM, 0, sv) == -1) {
        perror("socketpair");
        return 1;
    }

    char tmpl[] = "/tmp/peekfdXXXXXX";
    int filefd = mkstemp(tmpl);
    if (filefd == -1) {
        perror("mkstemp");
        return 1;
    }

    printf("Temporary file: %s\n", tmpl);

    char data = 'X';
    struct iovec iov = {
        .iov_base = &data,
        .iov_len = 1,
    };

    char control[CMSG_SPACE(sizeof(int))];
    memset(control, 0, sizeof(control));

    struct msghdr msg;
    memset(&msg, 0, sizeof(msg));
    msg.msg_iov = &iov;
    msg.msg_iovlen = 1;
    msg.msg_control = control;
    msg.msg_controllen = sizeof(control);

    struct cmsghdr *cmsg = CMSG_FIRSTHDR(&msg);
    cmsg->cmsg_level = SOL_SOCKET;
    cmsg->cmsg_type = SCM_RIGHTS;
    cmsg->cmsg_len = CMSG_LEN(sizeof(int));

    memcpy(CMSG_DATA(cmsg), &filefd, sizeof(filefd));

    if (sendmsg(sv[0], &msg, 0) == -1) {
        perror("sendmsg");
        return 1;
    }

    printf("Sent FD %d\n", filefd);

    int peekfd = recv_fd(sv[1], MSG_PEEK);
    if (peekfd == -1)
        return 1;

    printf("Peek recv returned FD %d\n", peekfd);

    int recvfd = recv_fd(sv[1], 0);
    if (recvfd == -1)
        return 1;

    printf("Normal recv returned FD %d\n", recvfd);

    printf("Writing through peeked FD...\n");
    write_string(peekfd, "written via peeked FD\n");

    printf("Writing through normally received FD...\n");
    write_string(recvfd, "written via normal FD\n");

    if (fsync(peekfd) == -1)
        perror("fsync(peekfd)");

    if (lseek(filefd, 0, SEEK_SET) == -1) {
        perror("lseek");
        return 1;
    }

    printf("\nContents of file:\n");
    printf("-----------------\n");

    char buf[4096];
    ssize_t n;
    while ((n = read(filefd, buf, sizeof(buf))) > 0)
        fwrite(buf, 1, (size_t)n, stdout);

    printf("-----------------\n");

    close(peekfd);
    close(recvfd);
    close(filefd);
    close(sv[0]);
    close(sv[1]);

    unlink(tmpl);

    return 0;
}
```

</details>

ziemleszcz: Currently, this feature is used only by openiked on DATARO. You can check the sources in phoenix-rtos-ports (there is also a 33-fdpass.patch). This PR must not break this tool.

posix/unix.c
julianuziemblo: I think this is a bigger problem of that we don't have link resolution in the kernel. What do you think about adding link resolution to this piece of code in libphoenix somehow, and assuming here it is an absolute path as is done in the rest of the kernel (e.g. in `posix_link`)?

posix/unix.c
```diff
@@ -861,24 +985,47 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
+				hdrsz = (s->type == SOCK_DGRAM) ? sizeof(hdr.dgram) : sizeof(hdr.seqpkt);
+
+				if (_cbuffer_avail(&s->buffer) >= hdrsz) {
+					if (s->type == SOCK_DGRAM) {
+						(void)_cbuffer_peek(&s->buffer, &hdr.dgram, sizeof(hdr.dgram));
+						rlen = hdr.dgram.len;
+						if (hdr.dgram.namelen > 0U && src_addr != NULL && src_len != NULL && *src_len != 0U) {
+							(void)_cbuffer_peekOffs(&s->buffer, src_addr, min(hdr.dgram.namelen, *src_len), hdrsz);
+							*src_len = hdr.dgram.namelen;
+							hdrsz += hdr.dgram.namelen;
```
ziemleszcz: Advancing past `namelen`  only if `recv()` asks for source address. `recv(src_addr=NULL)` gets mangled packet.

posix/unix.c
```diff
@@ -861,24 +985,47 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
-				/* No action required */
+			else { /* SOCK_DGRAM or SOCK_SEQPACKET */
+				hdrsz = (s->type == SOCK_DGRAM) ? sizeof(hdr.dgram) : sizeof(hdr.seqpkt);
+
+				if (_cbuffer_avail(&s->buffer) >= hdrsz) {
+					if (s->type == SOCK_DGRAM) {
+						(void)_cbuffer_peek(&s->buffer, &hdr.dgram, sizeof(hdr.dgram));
+						rlen = hdr.dgram.len;
+						if (hdr.dgram.namelen > 0U && src_addr != NULL && src_len != NULL && *src_len != 0U) {
+							(void)_cbuffer_peekOffs(&s->buffer, src_addr, min(hdr.dgram.namelen, *src_len), hdrsz);
```
ziemleszcz: `src_addr` points to a `struct sockaddr`. This data should be copied into `sa_data`, and `sa_family` should be set to AF_UNIX.

posix/unix.c
```diff
@@ -471,18 +498,42 @@ int unix_accept4(unsigned int socket, struct sockaddr *address, socklen_t *addre
...
 
 		s->state &= ~US_ACCEPTING;
 		hal_spinlockClear(&s->spinlock, &sc);
 
 		/* FIXME: handle connecting socket removal */
 
+		if (s->name != NULL) {
+			r->conn.remoteaddr = lib_strdup(s->name);
+			if (r->conn.remoteaddr == NULL) {
+				vm_kfree(v);
```
ziemleszcz: Double free on `v`. Note that `v` is already set in `new` socket so unixsock_put will free it.

posix/unix.c
```diff
@@ -471,18 +498,42 @@ int unix_accept4(unsigned int socket, struct sockaddr *address, socklen_t *addre
...
+				break;
+			}
+		}
+
+		if (r->name != NULL) {
+			new->conn.remoteaddr = lib_strdup(r->name);
+			if (new->conn.remoteaddr == NULL) {
+				vm_kfree(r->conn.remoteaddr);
+				r->conn.remoteaddr = NULL;
+				vm_kfree(v);
```
ziemleszcz: Double free on `v`. Note that `v `is already set in `new` socket so `unixsock_put` will free it.

posix/unix.c
```diff
@@ -471,18 +498,42 @@ int unix_accept4(unsigned int socket, struct sockaddr *address, socklen_t *addre
...
 		/* FIXME: handle connecting socket removal */
 
+		if (s->name != NULL) {
+			r->conn.remoteaddr = lib_strdup(s->name);
+			if (r->conn.remoteaddr == NULL) {
+				vm_kfree(v);
+				unixsock_put(new);
+				unixsock_put(new);
+				err = -ENOMEM;
+				break;
```
ziemleszcz: Peer stays forever in `connect()`.

posix/unix.c
```diff
@@ -471,18 +498,42 @@ int unix_accept4(unsigned int socket, struct sockaddr *address, socklen_t *addre
...
+		if (r->name != NULL) {
+			new->conn.remoteaddr = lib_strdup(r->name);
+			if (new->conn.remoteaddr == NULL) {
+				vm_kfree(r->conn.remoteaddr);
+				r->conn.remoteaddr = NULL;
+				vm_kfree(v);
+				unixsock_put(new);
+				unixsock_put(new);
+				err = -ENOMEM;
+				break;
```
ziemleszcz: Peer stays forever in `connect()`.

posix/fdpass.c
```diff
@@ -168,26 +174,53 @@ int fdpass_unpack(fdpack_t **packs, void *control, socklen_t *controllen)
...
 
+		if (peek != 0U) {
+			/* increment the file's reference count */
+			(void)proc_lockSet(&file->lock);
+			++file->refs;
+			(void)proc_lockClear(&file->lock);
+		}
+
 		if (pack->cnt == 0U) {
 			LIST_REMOVE(packs, pack);
```
ziemleszcz, line 187: Pack restoration works only if we process all entries and reach `pack->cnt == 0U`. This is not guaranteed because of the `controllen` check in the `while` loop.

lib/cbuffer.c
```diff
@@ -75,24 +76,29 @@ size_t _cbuffer_read(cbuffer_t *buf, void *data, size_t sz)
...
 {
 	size_t bytes = 0;
+	const size_t read = (buf->r + offs) & (buf->sz - 1U);
+	const size_t avail = _cbuffer_avail(buf);
 
 	/* FIXME: see note in _cbuffer_discard */
 	LIB_ASSERT(buf->sz > 0U, "attempted to peek at zero-sized buffer");
+	if (offs != 0U && avail != 0U) {
+		LIB_ASSERT(offs < avail, "attempted to peek at offset %zu when there's only %zu bytes in the buffer", offs, avail);
+	}
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist the assertion is inside an if statement that prevents your example from being true while disallowing invalid state which is peeking at offc >= avail

lib/strutil.c
```diff
@@ -63,3 +63,14 @@ void lib_splitname(char *path, char **base, const char **dir)
...
+
+
+void lib_unsplitname(char *path, const char *dir)
+{
+	size_t dirlen = hal_strlen(dir);
+
+	/* parasoft-suppress-next-line MISRAC2012-RULE_14_3-ac-2 "False-positive: dirlen can be == 1 and != 1 as it depends on user-provided string." */
+	if (dirlen != 1U) {
+		path[dirlen] = '/';
+	}
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist for paths like `/socket`, `dir` is `/`, so its length is 1. You got path (`base`) and `dir` mixed up.

posix/unix.c
```diff
@@ -550,22 +588,28 @@ int unix_bind(unsigned int socket, const struct sockaddr *address, socklen_t add
...
+				}
 				if (s->type == SOCK_DGRAM) {
 					_cbuffer_init(&s->buffer, NULL, 0);
 					vm_kfree(v);
 				}
 				break;
 			}
 
-			s->state |= US_BOUND;
+			lib_unsplitname(s->name, dir);
```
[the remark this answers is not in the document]
julianuziemblo: yeah, but it requires another allocation, which is not free, and we have to worry about dealocating it on every early return path.

posix/unix.c
Darchiv: BSDs, on the other hand, ignore `MSG_PEEK` on `SCM_RIGHTS`. I don't think we should mimic Linux in this case, especially since the behavior was exploited a few times (CVE-2026-53361, CVE-2021-0920).

As said above, we should not break existing software (openiked). Maybe just ignore `MSG_PEEK` then (return 0)?

julianuziemblo: It should not break openiked from what I've checked. But the BSD compatibility/the CVEs are good arguments, will remove it and not handle MSG_PEEK for ancillary data. Thanks for the input.

posix/unix.c
```diff
@@ -864,24 +978,55 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
+						*flags |= MSG_TRUNC;
+					}
+				}
 			}
-			/* TODO: peek control data */
-			if (peek == 0U) {
-				if (err > 0 && control != NULL && controllen != NULL && *controllen > 0U) {
-					(void)fdpass_unpack(&s->fdpacks, control, controllen);
+
+			if (peek == 0U && err > 0 && control != NULL && controllen != NULL && *controllen > 0U) {
```
ziemleszcz: Add `else` branch (fix for pre-exisiting issue):
```
} else if (controllen != NULL) {
    *controllen = 0;
}
```

ziemleszcz: Maybe add a comment that Linux extracts file descriptors on MSG_PEEK as well.

posix/unix.c
```diff
@@ -468,21 +495,33 @@ int unix_accept4(unsigned int socket, struct sockaddr *address, socklen_t *addre
...
 
-		LIST_REMOVE(&s->connecting, r);
+		LIST_REMOVE(&s->conn.connecting, r);
 
 		s->state &= ~US_ACCEPTING;
 		hal_spinlockClear(&s->spinlock, &sc);
 
 		/* FIXME: handle connecting socket removal */
 
+		swap(new->conn.remoteaddr, r->conn.remoteaddr);
```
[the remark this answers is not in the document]
ziemleszcz: Correct. `connect()` doesn't set `r->conn.remoteaddr`. Anyway, this code is hard to understand. `conn` is the server socket structure (a list of waiting client sockets), so using it for a client socket is unintuitive. This code is already complex, so please try not to make it even harder to understand.

posix/unix.c
```diff
@@ -776,14 +821,77 @@ int unix_connect(unsigned int socket, const struct sockaddr *address, socklen_t
...
+		return -ENOTSOCK;
+	}
+
+	r = unixsock_get_remote(s);
+	if (r == NULL) {
+		unixsock_put(s);
+		return -ENOTCONN;
+	}
+
+	unix_copyName(r->name, address, address_len);
```
ziemleszcz: `r` is the accepted socket (the new socket created by `accept4()`), and its `name` is not set.

posix/unix.c
```diff
@@ -776,14 +821,77 @@ int unix_connect(unsigned int socket, const struct sockaddr *address, socklen_t
...
+	if (address == NULL || address_len == NULL) {
+		return -EINVAL;
+	}
+
+	s = unixsock_get(socket);
+	if (s == NULL) {
+		return -ENOTSOCK;
+	}
+
+	unix_copyName(s->name, address, address_len);
```
ziemleszcz: `name` is not set for an accepted socket (returned by `accept4()`).

posix/unix.c
```diff
@@ -864,24 +978,55 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
+									src_addr->sa_data[copylen] = '\0';
+								}
+							}
+							*src_len = hdr.dgram.namelen + sizeof(src_addr->sa_family);
+						}
+						hdrsz += hdr.dgram.namelen;
+					}
+					else { /* SOCK_SEQPACKET */
+						(void)_cbuffer_peek(&s->buffer, &hdr.seqpkt, sizeof(hdr.seqpkt));
+						rlen = hdr.seqpkt.len;
```
ziemleszcz, line 1029: `src_addr` is not set for `SOCK_SEQPACKET`.

posix/unix.c
```diff
@@ -776,14 +821,77 @@ int unix_connect(unsigned int socket, const struct sockaddr *address, socklen_t
 }
 
 
+static void unix_copyName(const char *name, struct sockaddr *address, socklen_t *address_len)
+{
+	if (name == NULL) {
```
ziemleszcz: Set a zero-len `AF_UNIX` address if `name` is NULL. You need to return a proper address for unnamed socket as well.

posix/unix.c
```diff
@@ -776,14 +821,77 @@ int unix_connect(unsigned int socket, const struct sockaddr *address, socklen_t
...
+	if (buflen < familylen) {
+		return;
+	}
+
+	const socklen_t copylen = min(namelen, buflen - familylen);
+
+	address->sa_family = AF_UNIX;
+	hal_memcpy(address->sa_data, name, copylen);
+	if (buflen - familylen > copylen) {
+		address->sa_data[copylen] = '\0';
```
ziemleszcz: `copylen - 1`?

posix/unix.c
```diff
@@ -776,14 +821,77 @@ int unix_connect(unsigned int socket, const struct sockaddr *address, socklen_t
...
+{
+	if (name == NULL) {
+		return;
+	}
+
+	const socklen_t namelen = (socklen_t)hal_strlen(name) + 1U;
+	const socklen_t familylen = (socklen_t)sizeof(address->sa_family);
+	const socklen_t buflen = *address_len;
+
+	*address_len = namelen + familylen;
```
ziemleszcz: This is not truncated len.

julianuziemblo: yes, and I think it should work like this - check how this works on e.g. Linux

ziemleszcz: OK, you're right. POSIX is not clear on this.

posix/unix.c
julianuziemblo: `copylen - 1` is the last copied character - we terminate the buffer with a NULL only if we can

posix/unix.c
julianuziemblo: I chose to explicitly disallow binding unix sockets to relative addresses. POSIX does not specify that every path type should be supported so this seems reasonable.

posix/unix.c
```diff
@@ -865,23 +1004,50 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
+
+					if (peek == 0U) {
+						/* discard the data */
+						(void)_cbuffer_discard(&s->buffer, hdrsz + rlen);
+					}
+
+					if (len < rlen) {
+						*flags |= MSG_TRUNC;
+					}
 				}
```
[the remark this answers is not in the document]
julianuziemblo: The corruption of a buffer is an internal unwanted event, so I'll add an assertion

posix/unix.c
```diff
@@ -492,6 +555,10 @@ int unix_accept4(unsigned int socket, struct sockaddr *address, socklen_t *addre
 		(void)proc_threadWakeup(&r->queue);
 		hal_spinlockClear(&r->spinlock, &sc);
 
+		if (address != NULL && address_len != NULL) {
+			unix_copyName(r->name, r->namelen, address, address_len);
```
ziemleszcz, line 556: Potential use-after-free. Peer can close this socket after wake up and release it.

posix/unix.c
```diff
@@ -526,16 +593,34 @@ int unix_bind(unsigned int socket, const struct sockaddr *address, socklen_t add
...
+		 * Disallow binding sockets to relative address in
+		 * the kernel as we don't track the CWD of a process.
+		 */
+		if (address->sa_data[0] != '/') {
+			err = -EINVAL;
+			break;
+		}
+
+		pathlen = address_len - (socklen_t)sizeof(address->sa_family);
+		s->name = vm_kmalloc((size_t)pathlen + 1U);
```
ziemleszcz: Publishing `s->name` is not protected in any way. Another thread can observe an intermediate state, for example when calling `getsockname()`. The same applies to freeing `s->name` in the error path.

posix/unix.c
```diff
@@ -865,23 +1007,52 @@ static ssize_t recv(unsigned int socket, void *buf, size_t len, unsigned int fla
...
+					unixsock_put(r);
+				}
 			}
-			/* TODO: peek control data */
-			if (peek == 0U) {
-				if (err > 0 && control != NULL && controllen != NULL && *controllen > 0U) {
-					(void)fdpass_unpack(&s->fdpacks, control, controllen);
+
+			/* NOTE: Linux extracts file descriptors on MSG_PEEK, but we don't */
+			if (peek == 0U && err > 0 && control != NULL && controllen != NULL) {
```
ziemleszcz: Set `*controllen` to 0 when peek is on.

posix/unix.c
julianuziemblo: could this be done under `r->spinlock`?

posix/uchannel.c
```diff
@@ -221,8 +221,11 @@ ssize_t uchannel_read(uchannel_t *ch, void *buf, size_t len, unsigned int flags,
 			(void)_cbuffer_peek(&ch->buffer, &rlen, sizeof(rlen));
 			ret = (ssize_t)min(len, rlen);
 
-			if ((flags & UCHANNEL_OP_PEEK) != 0U) {
+			if ((op & UCHANNEL_OP_PEEK) != 0U) {
 				(void)_cbuffer_peekAt(&ch->buffer, sizeof(rlen), buf, (size_t)ret);
+				if (rlen > (size_t)ret) {
```
ziemleszcz: rlen > len

posix/uchannel.c
```diff
@@ -231,6 +234,7 @@ ssize_t uchannel_read(uchannel_t *ch, void *buf, size_t len, unsigned int flags,
 				if (rlen > (size_t)ret) {
```
ziemleszcz: rlen > len

posix/uchannel.c
ziemleszcz: It doesn't make sense to set `MSG_TRUNC` for peek since no data is lost. Please verify this behavior on Linux.

PR #789, opened 2026-06-15
hal/armv7r/zynqmp: watchdog handler (ZynqMP RPU)

kber-ps:
## Description
This PR introduces Watchdog handlers for ZynqMP RPU

- LPD (low power domain) SWDT (Software watchdog) 
- LPD bus clock with 512 scaler used for watchdog counter
- `hal_wdgReload` is now defined
- `platformctl`: `pctl_wdg_reload` and `pctl_wdg_enable` (set new reload value, 
enable/disable respectively) are now defined

## Motivation and Context
The ZynqMP RPU targets critical application in UAV domain. By default it is configured 
to  run in lock-step. Watchdog support is required for this kind of application.s

## Types of changes

## How Has This Been Tested?
- [X] Tested by hand on: (list targets here).

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7r/zynqmp/zynqmp.c
+++ b/hal/armv7r/zynqmp/zynqmp.c
@@ -5,8 +5,8 @@
  *
  * ZynqMP internal peripheral control functions
  *
- * Copyright 2025 Phoenix Systems
- * Author: Jacek Maksymowicz
+ * Copyright 2025, 2026 Phoenix Systems
+ * Author: Jacek Maksymowicz, Kamil Ber
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -17,18 +17,29 @@
 #include "hal/hal.h"
 #include "hal/armv7r/armv7r.h"
 #include "hal/spinlock.h"
+#include "include/errno.h"
 #include "include/arch/armv7r/zynqmp/zynqmp.h"
 #include "zynqmp.h"
 
 #include "hal/armv7r/halsyspage.h"
 #include "zynqmp_regs.h"
 
+#include <board_config.h>
+
 #define IOU_SLCR_BASE_ADDRESS 0xff180000U
 #define APU_BASE_ADDRESS      0xfd5c0000U
 #define CRF_APB_BASE_ADDRESS  0xfd1a0000U
 #define CRL_APB_BASE_ADDRESS  0xff5e0000U
 #define PMU_GLOBAL_MODULE     0xffd80000U
+#define SWDT_LPD_MODULE       0xff150000U
 
+#if defined(WATCHDOG)
+#if !defined(WATCHDOG_TIMEOUT)
+#error "WATCHDOG_TIMEOUT not defined!"
+#elif (WATCHDOG_TIMEOUT > 0xFFFU)
+#error "Watchdog timeout out of bounds!"
+#endif
+#endif
 
 /* PLO entrypoint */
 /* parasoft-suppress-next-line MISRAC2012-RULE_8_6 "Definition in assembly code" */
@@ -39,6 +50,7 @@ static struct {
 	volatile u32 *crf_apb;
 	volatile u32 *crl_apb;
 	volatile u32 *pmu_glb;
+	volatile u32 *swdt_lpd;
 	spinlock_t pltctlSp;
 } zynq_common;
 
@@ -423,9 +435,9 @@ __attribute__((noreturn)) void hal_cpuReboot(void)
 }
 
 
-/* TODO */
 void hal_wdgReload(void)
 {
+	*(zynq_common.swdt_lpd + swdt_lpd_restart) = SWDT_RESTART_RKEY;
 }
 
 
@@ -489,6 +501,40 @@ int hal_platformctl(void *ptr)
 			}
 			break;
 
+		case pctl_wdg_config:
+			/* If watchdog support is not enabled (WATCHDOG not defined) this will take no effects */
+			ret = 0;
+			if ((data->action == pctl_set) && (data->wdg_config.reload >= 0x0U) && (data->wdg_config.reload <= 0xFFFU)) {
+				*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | (data->wdg_config.reload << SWDT_CONTROL_CRV_BIT) | SWDT_CONTROL_CLKSEL);
+				*(zynq_common.swdt_lpd + swdt_lpd_restart) = SWDT_RESTART_RKEY;
+			}
+			else if (data->action == pctl_get) {
+				data->wdg_config.reload = (unsigned int)((*(zynq_common.swdt_lpd + swdt_lpd_control) >> SWDT_CONTROL_CRV_BIT) & 0xFFFU);
+			}
+			else {
+				ret = -EINVAL;
+			}
+			break;
+
+		case pctl_wdg_enable:
+			/* If watchdog support is not enabled (WATCHDOG not defined) this will take no effects */
+			ret = 0;
+			if ((data->action == pctl_set) && (data->wdg_enable.enable == 1U)) {
+				*(zynq_common.swdt_lpd + swdt_lpd_restart) = SWDT_RESTART_RKEY;
+				*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY | SWDT_MODE_RSTLN | SWDT_MODE_RSTEN | SWDT_MODE_WDEN);
+			}
+			else if ((data->action == pctl_set) && (data->wdg_enable.enable == 0U)) {
+				*(zynq_common.swdt_lpd + swdt_lpd_restart) = SWDT_RESTART_RKEY;
+				*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY);
+			}
+			else if (data->action == pctl_get) {
+				data->wdg_enable.enable = (*(zynq_common.swdt_lpd + swdt_lpd_mode) & SWDT_MODE_WDEN) != 0U ? 1 : 0;
+			}
+			else {
+				ret = -EINVAL;
+			}
+			break;
+
 		default:
 			/* Error by default */
 			break;
@@ -507,7 +553,48 @@ void _hal_platformInit(void)
 	zynq_common.crf_apb = (void *)CRF_APB_BASE_ADDRESS;
 	zynq_common.crl_apb = (void *)CRL_APB_BASE_ADDRESS;
 	zynq_common.pmu_glb = (void *)PMU_GLOBAL_MODULE;
+	zynq_common.swdt_lpd = (void *)SWDT_LPD_MODULE;
+
+	/* On ZynqMP platform, the hardware error handler is triggered on PMU_ERR_STATUS bits which
+	 * are only clear on external POR or by writing to this register (WTC). It means that after
+	 * software reset, status bits are preserved and hence a new reset would be triggered again
+	 * effectively leading to boot-loop. To avoid that, the error status bits should be cleared.
+	 * Here, we assume that this clear operation is performed in plo during the reboot reason
+	 * investigation.
+	 */
+
+	/* Enable reset on RPU lockstep errors */
 	_zynqmp_setSysErrHandler(PMU_ERR_RPU_LS, system_reset, 1);
+
+	/* Disable Watchdog */
+	/* Xilinx UG1085: unlocking/key values for registers:
+	 * mode: 0xabc
+	 * control: 0x248
+	 * restart: 0x1999
+	 *
+	 * Here both mode and control registers are reset after power on
+	 * or after watchdog timeout triggered reset. the control reset is specified
+	 * in the UG: "Note: If a restart signal is received the prescaler should be reset.
+	 */
+	(void)_zynq_setDevRst(pctl_devreset_lpd_swdt, 0);
+	*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY);
+	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | (0xFFFU << SWDT_CONTROL_CRV_BIT) | SWDT_CONTROL_CLKSEL);
+
+#if defined(WATCHDOG)
+	/* set cotnrol: LPD_LSBUS_CLK scaler = 512 (gives reload value from ~20ms to 85s for 100MHz LPD_LSBUS_CLK)*/
+	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | ((WATCHDOG_TIMEOUT & 0xFFFU) << SWDT_CONTROL_CRV_BIT) | SWDT_CONTROL_CLKSEL);
+
+	/* reload */
+	*(zynq_common.swdt_lpd + swdt_lpd_restart) = SWDT_RESTART_RKEY;
+
+	/* enable watchdog: no irq, reset enabled (32 clock cycles pulse) */
+	*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY | SWDT_MODE_RSTLN | SWDT_MODE_RSTEN | SWDT_MODE_WDEN);
+
+	/* By default the PMU error handler is disabled and resets generated by SWDT are ignored */
+
+	/* Enable reset on LPD SWDT  */
+	_zynqmp_setSysErrHandler(PMU_ERR_LPD_SWDT, system_reset, 1);
+#endif
 }
 
 
--- a/hal/armv7r/zynqmp/zynqmp_regs.h
+++ b/hal/armv7r/zynqmp/zynqmp_regs.h
@@ -245,5 +245,12 @@ enum {
 	pmu_glb_error_en_2
 };
 
+enum {
+	swdt_lpd_mode = 0x00,
+	swdt_lpd_control,
+	swdt_lpd_restart,
+	swdt_lpd_status
+};
+
 
 #endif /* _PH_ZYNQMP_REGS_H_ */
--- a/include/arch/armv7r/zynqmp/zynqmp.h
+++ b/include/arch/armv7r/zynqmp/zynqmp.h
@@ -5,8 +5,8 @@
  *
  * ZynqMP basic peripherals control functions
  *
- * Copyright 2021, 2024 Phoenix Systems
- * Author: Hubert Buczynski, Jacek Maksymowicz
+ * Copyright 2021, 2024, 2026 Phoenix Systems
+ * Author: Hubert Buczynski, Jacek Maksymowicz, Kamil Ber
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -44,6 +44,18 @@
 #define PMU_ERR_PL       ((0xFUL << 2) + PMU_ERR_SECOND_SET)
 #define PMU_ERR_PLL_LOCK ((0x1FUL << 8) + PMU_ERR_SECOND_SET)
 #define PMU_ERR_CSU      ((0x1UL << 16) + PMU_ERR_SECOND_SET)
+
+#define SWDT_ZKEY            (0xABCU)
+#define SWDT_CKEY            (0x248U)
+#define SWDT_RKEY            (0x1999U)
+#define SWDT_MODE_ZKEY       (SWDT_ZKEY << 12)
+#define SWDT_MODE_WDEN       (0x1U)
+#define SWDT_MODE_RSTEN      (0x1U << 1)
+#define SWDT_MODE_RSTLN      (0x4U << 4) /* reset len = 4 -> 32 clock cycles*/
+#define SWDT_CONTROL_CKEY    (SWDT_CKEY << 14)
+#define SWDT_CONTROL_CRV_BIT (2)
+#define SWDT_CONTROL_CLKSEL  (0x2U) /* LPD_LSBUS_CLK/512 */
+#define SWDT_RESTART_RKEY    (SWDT_RKEY)
 
 
 /* clang-format off */
@@ -145,7 +157,7 @@ enum {
 typedef struct {
 	/* clang-format off */
 	enum { pctl_set = 0, pctl_get } action;
-	enum { pctl_devclock = 0, pctl_mioclock, pctl_mio, pctl_devreset, pctl_reboot } type;
+	enum { pctl_devclock = 0, pctl_mioclock, pctl_mio, pctl_devreset, pctl_reboot, pctl_wdg_config, pctl_wdg_enable } type;
 	/* clang-format on */
 	union {
 		struct {
@@ -174,6 +186,14 @@ typedef struct {
 			__u32 magic;
 			__u32 reason;
 		} reboot;
+
+		struct {
+			__u32 reload;
+		} wdg_config;
+
+		struct {
+			__u8 enable;
+		} wdg_enable;
 	};
 } __attribute__((packed)) platformctl_t;
 


hal/armv7r/zynqmp/zynqmp_regs.h
```diff
@@ -245,5 +245,12 @@ enum {
 	pmu_glb_error_en_2
 };
 
+enum {
+	swdt_lpd_mode = 0x00,
+	swdt_lpd_control,
+	swdt_lpd_restart,
+	swdt_lpd_status
```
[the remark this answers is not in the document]
kber-ps: image

image

include/arch/armv7r/zynqmp/zynqmp.h
```diff
@@ -45,6 +45,18 @@
 #define PMU_ERR_PLL_LOCK ((0x1FUL << 8) + PMU_ERR_SECOND_SET)
 #define PMU_ERR_CSU      ((0x1UL << 16) + PMU_ERR_SECOND_SET)
 
+#define SWDT_ZKEY            (0xABCU)
+#define SWDT_CKEY            (0x248U)
+#define SWDT_RKEY            (0x1999U)
+#define SWDT_MODE_ZKEY       (SWDT_ZKEY << 12)
+#define SWDT_MODE_WDEN       (0x1U)
+#define SWDT_MODE_RSTEN      (0x1U << 1)
+#define SWDT_MODE_RSTLN      (0x4U << 4) /* reset len = 4 -> 32 clock cycles*/
```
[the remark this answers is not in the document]
kber-ps: image

include/arch/armv7r/zynqmp/zynqmp.h
```diff
@@ -174,6 +186,14 @@ typedef struct {
...
 			__u32 reason;
 		} reboot;
+
+		struct {
+			__u32 reload;
+		} wdg_reload;
+
+		struct {
+			__u8 enable;
+		} wdg_enable;
```
julianuziemblo, line 196: nitpick: I could be wrong on this, but wouldn't it be clearer to do something like this:

```c
union {
    __u32 reload;
    __u8 enable;
} wdg;
```

Then again, all the other options have corresponding structures (except for `mioclock` which I guess belongs to `mio` structure?)...

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -489,6 +500,42 @@ int hal_platformctl(void *ptr)
 			}
 			break;
 
+		case pctl_wdg_reload:
+			/* If watchdog is not enabled this will take no effects*/
+			if ((data->action == pctl_set) && (data->wdg_reload.reload >= 0x000FFFU) && (data->wdg_reload.reload <= 0xFFFFFFU)) {
```
julianuziemblo: returning -EINVAL in case the latter 2 conditions are wrong could be more informative to the user

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -489,6 +500,42 @@ int hal_platformctl(void *ptr)
...
+				*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | ((data->wdg_reload.reload >> 12) & 0xFFFU) << SWDT_CONTROL_CRV_BIT | SWDT_CONTROL_CLKSEL);
```
julianuziemblo: nitpick: add parenthesis over `((data->wdg_reload.reload >> 12) & 0xFFFU) << SWDT_CONTROL_CRV_BIT`

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -489,6 +500,39 @@ int hal_platformctl(void *ptr)
...
+
+		case pctl_wdg_enable:
+			if ((data->action == pctl_set) && (data->wdg_enable.enable == 1)) {
+				*(zynq_common.swdt_lpd + swdt_lpd_restart) = SWDT_RESTART_RKEY;
+				*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY | SWDT_MODE_RSTLN | SWDT_MODE_RSTEN | SWDT_MODE_WDEN);
+			}
+			else if ((data->action == pctl_set) && (data->wdg_enable.enable == 0)) {
+				*(zynq_common.swdt_lpd + swdt_lpd_restart) = SWDT_RESTART_RKEY;
+				*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY);
+			}
```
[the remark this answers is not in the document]
kber-ps: Fixed in new version

kber-ps:
Short summary of manual test performed on the ZCU104 board:

In the default configuration, watchdog timeout is set to ~85 seconds. 

* Watchdog is not active in plo:
Screenshot from 2026-07-08 20-50-33

* Watchdog is active in phoenix-rtos, resets are properly triggered:
Screenshot from 2026-07-08 21-30-42

* `wdgreload` syscall properly refreshes watchdog counter
Screenshot from 2026-07-08 21-36-39

* reload/enable platformctl effects:
Screenshot from 2026-07-08 21-37-18

Here, watchdog has been re-configured for 10s timout, after reset the reset reason is also properly 
populated in plo and then passed on to kernel.

include/arch/armv7r/zynqmp/zynqmp.h
kber-ps: I wanted to keep the enable and reload action separated. Now I realized that this is not reload but rather configuration, for reload we have the dedicated syscall. 

Unions/structure I'm leaving as is, but I'm changing the name of the platformctl action.

hal/armv7r/zynqmp/zynqmp.c
kber-ps: Done in new version

hal/armv7r/zynqmp/zynqmp.c
kber-ps: Done, thanks

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -5,8 +5,8 @@
  *
  * ZynqMP internal peripheral control functions
  *
- * Copyright 2025 Phoenix Systems
- * Author: Jacek Maksymowicz
+ * Copyright 2026 Phoenix Systems
```
agkaminski:
```suggestion
 * Copyright 2025, 2026 Phoenix Systems
```

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -489,6 +501,43 @@ int hal_platformctl(void *ptr)
...
+				*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | (((data->wdg_reload.reload >> 12U) & 0xFFFU) << SWDT_CONTROL_CRV_BIT) | SWDT_CONTROL_CLKSEL);
+				*(zynq_common.swdt_lpd + swdt_lpd_restart) = SWDT_RESTART_RKEY;
+				ret = 0;
+			}
+			else if (data->action == pctl_get) {
+				data->wdg_reload.reload = (unsigned int)((((*(zynq_common.swdt_lpd + swdt_lpd_control) >> SWDT_CONTROL_CRV_BIT) & 0xFFFU) << 12U) | 0xFFFU);
+				ret = 0;
+			}
+			else {
+				ret = -EINVAL;
```
agkaminski, line 515: Consider replacing default `ret` value with `0` instead and putting `ret = 0` everywhere. Right now the new code is inconsistent with the old one

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -489,6 +501,43 @@ int hal_platformctl(void *ptr)
...
+			}
+			break;
+
+		case pctl_wdg_enable:
+			/* If watchdog is not enabled this will take no effects*/
```
agkaminski: Enable will take no effect if not already enabled?

include/arch/armv7r/zynqmp/zynqmp.h
agkaminski, line 196: I am not a fan of that - watchdog, to be effective should be enabled as early as possible (best if enabled by bootrom, even before user, i.e. our, code starts). IMHO if wachdog is enabled in the build config it should be enabled ASAP by plo.

include/arch/armv7r/zynqmp/zynqmp.h
```diff
@@ -174,6 +186,14 @@ typedef struct {
 			__u32 magic;
 			__u32 reason;
 		} reboot;
+
+		struct {
+			__u32 reload;
+		} wdg_reload;
```
agkaminski: Structure name is inconsistent with the action

include/arch/armv7r/zynqmp/zynqmp.h
kber-ps: I do not agree. Enabling watchdog in bootloader is not the best idea, especially if in fact it does not provide userspace as it was never designed for this. 

I looked into other implementation of watchdog in our kernel and it seems the the general consensus was to provide a user space action, i.e. syscall for refreshing (BTW it would be also better name that the established in our codebase reload)  - hence it's user space responsibility for managing it. 

By default WATCHDOG param is not defined and therefore watchdog support is not enabled - and by that I mean, the watchdog reset action at hardware level is never enabled. This action is only configured in the `_hal_platformInit()` - this line to be exact:
```
	_zynqmp_setSysErrHandler(PMU_ERR_LPD_SWDT, system_reset, 1);
``` 

To summarize, watchdog support is enabled only when WATCHDOG 
param is defined. Without it the reset action is not enabled. 

To limit the lines of so-called dead code (see DO-178) I decided to not add the `#ifdef` blocks to `platformCtl` handlers.

include/arch/armv7r/zynqmp/zynqmp.h
kber-ps: Fixed in new version

hal/armv7r/zynqmp/zynqmp.c
kber-ps: OK, i'm updating it, but please note I've only added the watchdog related cases so the structure has been defined in the `old` code co I will not agree entirely that the change is not consistent with the old code.

hal/armv7r/zynqmp/zynqmp.c
kber-ps: I updated the comment so it now better describes the situation - my intention was to refer to the watchdog support which is enabled with WATCHDOG define

include/arch/armv7r/zynqmp/zynqmp.h
```diff
@@ -45,6 +45,15 @@
 #define PMU_ERR_PLL_LOCK ((0x1FUL << 8) + PMU_ERR_SECOND_SET)
 #define PMU_ERR_CSU      ((0x1UL << 16) + PMU_ERR_SECOND_SET)
 
+#define SWDT_MODE_ZKEY       (0xABCU << 12)
```
rmikielis: acc. to our coding convention (MISRA recom.) - "When shifting unsigned constants by more than 7 bits the UL suffix should be used"

include/arch/armv7r/zynqmp/zynqmp.h
```diff
@@ -45,6 +45,15 @@
...
+#define SWDT_MODE_WDEN       (0x1U)
+#define SWDT_MODE_RSTEN      (0x1U << 1)
+#define SWDT_MODE_RSTLN      (0x4U << 4)    /* reset len = 4 -> 32 clock cycles*/
+#define SWDT_CONTROL_CKEY    (0x248U << 12)
```
rmikielis: Isn't shift for control key 14 bits?

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -507,7 +520,35 @@ void _hal_platformInit(void)
...
+	 * or after watchdog timeout triggered reset. the control reset is specified
+	 * in the UG: "Note: If a restart signal is received the prescaler should be reset."
+	 */
+	*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY);
+	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | 0xFFF << SWDT_CONTROL_CRV_BIT);
+
+	#if defined(WATCHDOG)
+
+	/* set cotnrol: LPD_LSBUS_CLK scaler = 512 (gives reload value from ~20ms to 85s fro 100MHz LPD_LSBUS_CLK)*/
+	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | ((WATCHDOG_TIMEOUT >> 12) & 0xFFF) << SWDT_CONTROL_CRV_BIT | SWDT_CONTROL_CLKSEL );
```
rmikielis: nit: I wonder if we can't make from `((WATCHDOG_TIMEOUT >> 12) & 0xFFF) << SWDT_CONTROL_CRV_BIT` another macro, pasting it in source code is hard to read

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -507,7 +520,35 @@ void _hal_platformInit(void)
 	zynq_common.crf_apb = (void *)CRF_APB_BASE_ADDRESS;
 	zynq_common.crl_apb = (void *)CRL_APB_BASE_ADDRESS;
 	zynq_common.pmu_glb = (void *)PMU_GLOBAL_MODULE;
+	zynq_common.swdt_lpd = (void *)SWDT_LPD_MODULE;
 	_zynqmp_setSysErrHandler(PMU_ERR_RPU_LS, system_reset, 1);
+
+
+	/* Disable Watchdog */
```
rmikielis, line 569: I wonder if we want to do this all in `_hal_platforminit()`. If we have `_zynqmp_setSysErrHandler`, maybe adding smth similar for watchdog handler would be more coherent. Just a suggestion.

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -532,23 +574,27 @@ void _hal_platformInit(void)
...
-	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | 0xFFF << SWDT_CONTROL_CRV_BIT);
-
-	#if defined(WATCHDOG)
+	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | (0xFFFU << SWDT_CONTROL_CRV_BIT) | SWDT_CONTROL_CLKSEL);
 
-	/* set cotnrol: LPD_LSBUS_CLK scaler = 512 (gives reload value from ~20ms to 85s fro 100MHz LPD_LSBUS_CLK)*/
-	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | ((WATCHDOG_TIMEOUT >> 12) & 0xFFF) << SWDT_CONTROL_CRV_BIT | SWDT_CONTROL_CLKSEL );
+#if defined(WATCHDOG)
+	/* set cotnrol: LPD_LSBUS_CLK scaler = 512 (gives reload value from ~20ms to 85s for 100MHz LPD_LSBUS_CLK)*/
+	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | ((WATCHDOG_TIMEOUT >> 12U) & 0xFFFU) << SWDT_CONTROL_CRV_BIT | SWDT_CONTROL_CLKSEL);
```
rmikielis: nit: I wonder if we can't make from `((WATCHDOG_TIMEOUT >> 12) & 0xFFF) << SWDT_CONTROL_CRV_BIT` another macro, pasting it in source code is hard to read

include/arch/armv7r/zynqmp/zynqmp.h
```diff
@@ -45,14 +45,17 @@
...
+#define SWDT_MODE_ZKEY       (SWDT_ZKEY << 12)
 #define SWDT_MODE_WDEN       (0x1U)
 #define SWDT_MODE_RSTEN      (0x1U << 1)
-#define SWDT_MODE_RSTLN      (0x4U << 4)    /* reset len = 4 -> 32 clock cycles*/
-#define SWDT_CONTROL_CKEY    (0x248U << 12)
-#define SWDT_CONTROL_CRV_BIT 2
-#define SWDT_CONTROL_CLKSEL  (0x2U)         /* LPD_LSBUS_CLK/512 */
-#define SWDT_RESTART_RKEY    (0x1999U)
+#define SWDT_MODE_RSTLN      (0x4U << 4) /* reset len = 4 -> 32 clock cycles*/
+#define SWDT_CONTROL_CKEY    (SWDT_CKEY << 14)
```
rmikielis, line 55: acc. to our coding convention (MISRA recom.) - "When shifting unsigned constants by more than 7 bits the UL suffix should be used". Applies to all.

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -502,6 +501,40 @@ int hal_platformctl(void *ptr)
 			}
 			break;
 
+		case pctl_wdg_config:
+			/* If watchdog support is not enabled (WATCHDOG not defined) this will take no effects */
+			ret = 0;
+			if ((data->action == pctl_set) && (data->wdg_config.reload >= 0x000FFFU) && (data->wdg_config.reload <= 0xFFFFFFU)) {
```
rmikielis: I wonder if reload value received from user should be 24 bit value. Acc. to TRM, lower 12 bits are always 0xFFF, user can only set upper 12 bits, but we still require from him values > than what can be set (acc. to TRM)

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -502,6 +501,40 @@ int hal_platformctl(void *ptr)
...
+			}
+			else {
+				ret = -EINVAL;
+			}
+			break;
+
+		case pctl_wdg_enable:
+			/* If watchdog support is not enabled (WATCHDOG not defined) this will take no effects */
+			ret = 0;
+			if ((data->action == pctl_set) && (data->wdg_enable.enable == 1U)) {
```
rmikielis, line 522: we hardcode setting reset when watchdog is started, but from what I see in TRM also IRQ can be set as a handler. I think user should be the one to determine action on SWDT expiry. Or return error if we don't plan to support IRQ as a handler, but user issues such request.

rmikielis:
Also can you please squash those 2 commits

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -507,7 +551,33 @@ void _hal_platformInit(void)
...
+	 *
+	 * Here both mode and control registers are reset after power on
+	 * or after watchdog timeout triggered reset. the control reset is specified
+	 * in the UG: "Note: If a restart signal is received the prescaler should be reset."
+	 */
+	*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY);
+	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | (0xFFF << SWDT_CONTROL_CRV_BIT)| SWDT_CONTROL_CLKSEL );
+
+#if defined(WATCHDOG)
+	/* set cotnrol: LPD_LSBUS_CLK scaler = 512 (gives reload value from ~20ms to 85s fro 100MHz LPD_LSBUS_CLK)*/
```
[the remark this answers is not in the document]
kber-ps: Fixed

hal/armv7r/zynqmp/zynqmp.c
```diff
@@ -507,7 +551,33 @@ void _hal_platformInit(void)
...
+	 * mode: 0xabc
+	 * control: 0x248
+	 * restart: 0x1999
+	 *
+	 * Here both mode and control registers are reset after power on
+	 * or after watchdog timeout triggered reset. the control reset is specified
+	 * in the UG: "Note: If a restart signal is received the prescaler should be reset."
+	 */
+	*(zynq_common.swdt_lpd + swdt_lpd_mode) = (SWDT_MODE_ZKEY);
+	*(zynq_common.swdt_lpd + swdt_lpd_control) = (SWDT_CONTROL_CKEY | (0xFFF << SWDT_CONTROL_CRV_BIT)| SWDT_CONTROL_CLKSEL );
```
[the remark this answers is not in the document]
kber-ps: Fixed

include/arch/armv7r/zynqmp/zynqmp.h
```diff
@@ -45,6 +45,18 @@
...
 #define PMU_ERR_CSU      ((0x1UL << 16) + PMU_ERR_SECOND_SET)
 
+#define SWDT_ZKEY            (0xABCU)
+#define SWDT_CKEY            (0x248U)
+#define SWDT_RKEY            (0x1999U)
+#define SWDT_MODE_ZKEY       (SWDT_ZKEY << 12)
+#define SWDT_MODE_WDEN       (0x1U)
+#define SWDT_MODE_RSTEN      (0x1U << 1)
+#define SWDT_MODE_RSTLN      (0x4U << 4) /* reset len = 4 -> 32 clock cycles*/
+#define SWDT_CONTROL_CKEY    (SWDT_CKEY << 12)
```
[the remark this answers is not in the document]
kber-ps: Fixed:

```c
#define SWDT_CONTROL_CKEY    (SWDT_CKEY << 14)
```

hal/armv7r/zynqmp/zynqmp.c
kber-ps: OK, I changed it and added comment to board_config.h

I could also define a macro to calculate the CNT value for time defined 
in ms (or s) but the clocks are configured outside of the kernel (in plo) and we would need a function to extract the watchdog clock frequency and configuration. This is not in the scope of this PR.

hal/armv7r/zynqmp/zynqmp.c
kber-ps: That's a valid idea but for this PR. I will add TODO so we can think to redefine this interface in the future. 

Watchdog is required for new DO-178 version. The hardware solution is the most robust one. With interrupt there's a a very small chance that program hands in the interrupt handler when interrupts, including the one from watchdog are masked out.

hal/armv7r/zynqmp/zynqmp.c
kber-ps: I considered Your second comment and changed the interface which simplified this code. For DO-178 I would like to limit number's of different macros as much as possible.

kber-ps:
> Also can you please squash those 2 commits

Done

hal/armv7r/zynqmp/zynqmp.c
agkaminski: Sorry, it should've been 2 comments really.

During `ret` declaration it would be better to init it with `0` instead - I mean the change to the old code.

Separate issue was/is that the new code use different pattern - preferably should be kept the same.

include/arch/armv7r/zynqmp/zynqmp.h
agkaminski: Yes, that is ok that sometimes we want watchdog to be disabled entirely. Side note - please make sure that the reload action do not enable it then.

What I meant is that if watchdog is enabled that it should be enabled ASAP, not by platformctl - due to the transient error we might never reach platformctl in the user code and thus never be reseted. So action of enabling the watchdog from the user space should not be needed nor available.

damianloew (commented):
Can we merge that? It's strongly needed for DO project

hal/armv7r/zynqmp/zynqmp.c
rmikielis: Right, we don't have to do IRQ watchdog right now. Just TODO comment, so someone calling platformctl is aware he is setting system reset w/o searching in kernel source code

Darchiv (approved):

PR #791, opened 2026-06-17
proc: add sched_info syscall

adamgreloch:
Pthreads need to validate sched_param->sched_priority values. It'd be best not to hardcode the bounds in libphoenix.

TASK: RTOS-1353

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/include/sched.h
+++ b/include/sched.h
@@ -1,0 +1,23 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Operating system kernel
+ *
+ * Scheduler definitions
+ *
+ * Copyright 2026 Phoenix Systems
+ * Authors: Adam Greloch
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _PH_SCHED_H_
+#define _PH_SCHED_H_
+
+
+#define SCHED_FIFO  0
+#define SCHED_RR    1
+#define SCHED_OTHER 2
+
+
+#endif
--- a/include/sched.h
+++ b/include/sched.h
@@ -14,10 +14,21 @@
 #ifndef _PH_SCHED_H_
 #define _PH_SCHED_H_
 
+#include "types.h"
 
 #define SCHED_FIFO  0
 #define SCHED_RR    1
 #define SCHED_OTHER 2
+
+
+typedef struct {
+	time_t interval;
+	int minPriority;
+	int maxPriority;
+	union {
+		char reserved[32]; /* reserved for policy-specific values */
+	} policy;
+} sched_info_t;
 
 
 #endif
--- a/syscalls.c
+++ b/syscalls.c
@@ -21,6 +21,7 @@
 #include "include/errno.h"
 #include "include/sysinfo.h"
 #include "include/mman.h"
+#include "include/sched.h"
 #include "include/syscalls.h"
 #include "include/threads.h"
 #include "include/utsname.h"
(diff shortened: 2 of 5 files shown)


include/threads.h
```diff
@@ -17,6 +17,14 @@
 #define _PH_THREADS_H_
 
 
+/* Scheduling constants */
+
+
+/* NOTE: lower numerical values represent more critical scheduling priorities */
+#define PH_MAX_PRIO 7 /* least-critical priority */
```
ziemleszcz: I would avoid exposing these constants in the public API and introducing a new `PH_*` naming scheme unless they are required by external applications. For now, they can be moved to the source file.

etiaro:
This API is problematic e.g. when considering changing priorities number, as it requires recompilation of both kernel and userspace programs with new headers to keep compatibility.
Another thing is future-proofing for multiple scheduling policies, etc.
If we really need to fetch this values from kernel now, I would opt for syscall.

adamgreloch:
Good suggestions. I have thus replaced the defines with `sched_info` syscall that will be nice to have anyways once we get into implementing more scheduling policies

syscalls.c
```diff
@@ -375,6 +376,30 @@ int syscalls_priority(u8 *ustack)
...
 
 
+int syscalls_sched_info(u8 *ustack)
+{
+	process_t *proc;
+	int policy;
+	pid_t pid;
+	sched_info_t *info;
+
+	GETFROMSTACK(ustack, pid_t, pid, 0U);
```
ziemleszcz, line 386: How is `pid` used/going to be used by this syscall?

adamgreloch: Parameters for some more advanced policies may potentially differ per process if we ever decide to implement PTHREAD_SCOPE_PROCESS. This is also the API that QNX uses: https://www.qnx.com/developers/docs/8.0/com.qnx.doc.neutrino.lib_ref/topic/s/schedinfo.html

include/sched.h
```diff
@@ -0,0 +1,40 @@
...
+#define _PH_SCHED_H_
+
+#include "types.h"
+
+#define SCHED_FIFO  0
+#define SCHED_RR    1
+#define SCHED_OTHER 2
+
+
+struct sched_param {
```
etiaro: Why is it moved here? I don't see any use inside kernel currently, and it's apparently not a stable API with a TODO inside.

syscalls.c
```diff
@@ -375,6 +376,30 @@ int syscalls_priority(u8 *ustack)
...
+	proc = proc_find(pid);
+	if (proc == NULL) {
+		return -EINVAL;
+	}
+
+	if (vm_mapBelongs(proc_current()->process, info, sizeof(*info)) < 0) {
+		return -EINVAL;
+	}
+
+	return proc_schedInfo(proc, policy, info);
```
etiaro: Missing proc_put

syscalls.c
```diff
@@ -375,6 +376,35 @@
...
+	GETFROMSTACK(ustack, int, policy, 1U);
+	GETFROMSTACK(ustack, sched_info_t *, info, 2U);
+
+	proc = proc_find(pid);
+	if (proc == NULL) {
+		return -EINVAL;
+	}
+
+	if (vm_mapBelongs(proc_current()->process, info, sizeof(*info)) < 0) {
+		proc_put(proc);
```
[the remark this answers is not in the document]
ziemleszcz: Use `(void)`.

syscalls.c
```diff
@@ -375,6 +376,35 @@
...
+		return -EINVAL;
+	}
+
+	err = proc_schedInfo(proc, policy, info);
+
+	proc_put(proc);
```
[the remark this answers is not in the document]
ziemleszcz: Use `(void)`.

syscalls.c
```diff
@@ -375,6 +376,35 @@ int syscalls_priority(u8 *ustack)
 }
 
 
+int syscalls_sched_info(u8 *ustack)
```
ziemleszcz: sched_info -> schedInfo (camelCase) or schedinfo?

ziemleszcz (approved):

etiaro (approved):

PR #793, opened 2026-06-19
posix: return ECHILD from multithreaded waitpid

etiaro:
According to POSIX, when there are no more remaining threads matching waitpid call, it should return with ECHILD error code.

TASK: RTOS-1357

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -2721,9 +2721,7 @@ int posix_waitpid(pid_t child, int *status, unsigned int options)
 			break;
 		}
 
-		do {
-			err = proc_lockWait(&pinfo->wait, &pinfo->lock, 0);
-		} while ((pinfo->zombies == NULL) && (err == EOK));
+		err = proc_lockWait(&pinfo->wait, &pinfo->lock, 0);
 
 		if (err == -EINTR) {
 			/* pinfo->lock is clear */


posix/posix.c
```diff
@@ -2723,7 +2723,7 @@ int posix_waitpid(pid_t child, int *status, unsigned int options)
 
 		do {
 			err = proc_lockWait(&pinfo->wait, &pinfo->lock, 0);
-		} while ((pinfo->zombies == NULL) && (err == EOK));
+		} while ((pinfo->zombies == NULL) && (err == EOK) && (pinfo->children != NULL));
```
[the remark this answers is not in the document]
etiaro: The example is flawed, but he actually has a good point.

julianuziemblo (approved):

posix/posix.c
ziemleszcz, line 2769: Unrelated directly to this commit but why is `SIGCHILD` sent only when no one was waiting in waitpid()?

ziemleszcz (approved):
LGTM, except for my question about waitpid(), which can be addressed in a separate PR.

posix/posix.c
etiaro: It is a workaround to not interrupt the thread waiting in waitpid. As signal handlers are defined currently inside libphoenix, kernel has no idea whether it should be ignored or handled, so it just interrupts and sends the signal to any thread that is available, and then libphoenix calls the handler or discards it.

This is a good point to address in signal rework PR, I am not sure what is expected behavior here. In case a handler is installed, it should "steal" the child from waitpid?

ziemleszcz: No. IMHO reaping the child with `waitpid()` should not suppress signal delivery. The signal should still be delivered if it is not ignored.

agkaminski: @ziemleszcz https://github.com/phoenix-rtos/phoenix-rtos-project/issues/188

etiaro: Even more details in phoenix-rtos/phoenix-rtos-project#1319

ziemleszcz: Worth adding a comment in the code.

etiaro: After some consideration, there is still an error here – in case there is an active waitpid that doesn't match this particular dying process the SIGCHLD should be sent, but it won't be sent.

PR #797, opened 2026-06-24
vm/map: implement mapBelongs for NOMMU

### LINKED ISSUE #1593: Unrechable ifdef for NOMMU in _vm_mapBelongs()
The ifdef for NOMMU in _vm_mapBelongs() is guarded by mutually exclusive ifndef NOMMU

```c
static int _vm_mapBelongs(const struct _process_t *proc, const void *ptr, size_t size)
{
/* Disabled for now on NOMMU, as we can now receive memory
 * from different maps and processes via msg */
#ifndef NOMMU
	map_entry_t e, *f;

	if (size == 0U) {
		return 0;
	}
	/* parasoft-suppress-next-line MISRAC2012-RULE_11_8 "Structure 'e' is used only for tree searching, pointer will not be modified. We cannot change vaddr member of the map_entry_t to const." */
	e.vaddr = (void *)ptr;
	e.size = size;

	f = lib_treeof(map_entry_t, linkage, lib_rbFind(&proc->mapp->tree, &e.linkage));
	if ((f == NULL) && (proc->imapp != NULL)) {
		f = lib_treeof(map_entry_t, linkage, lib_rbFind(&proc->imapp->tree, &e.linkage));
	}

	if (f == NULL) {
		return -1;
	}

#ifdef NOMMU
	if (f->process != proc) {
		return -1;
	}
#endif

#endif

	return 0;
}
```

etiaro:
## Description

Improve separation between processes with different physical maps.

## Motivation and Context

Improves future partitioning isolation.

Fixes phoenix-rtos/phoenix-rtos-project#1593

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/vm/map.c
+++ b/vm/map.c
@@ -1178,14 +1178,21 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
 
 static int _vm_mapBelongs(const struct _process_t *proc, const void *ptr, size_t size)
 {
-/* Disabled for now on NOMMU, as we can now receive memory
- * from different maps and processes via msg */
-#ifndef NOMMU
-	map_entry_t e, *f;
-
 	if (size == 0U) {
 		return 0;
 	}
+
+#if defined(NOMMU) && defined(NOMMU_MAPBELONGS_VALIDATE_MAP)
+	return (pmap_isAllowed(proc->pmapp, ptr, size) == 0) ? -1 : 0;
+	/*
+	 * Disabled checking entry owner for now on NOMMU, as we can
+	 * receive memory from different maps and processes via msg
+	 * and RO elf segments may not be assigned to any process.
+	 * TODO: create memory objects for syspage programs
+	 */
+#elif !defined(NOMMU)
+	map_entry_t e, *f;
+
 	/* parasoft-suppress-next-line MISRAC2012-RULE_11_8 "Structure 'e' is used only for tree searching, pointer will not be modified. We cannot change vaddr member of the map_entry_t to const." */
 	e.vaddr = (void *)ptr;
 	e.size = size;
@@ -1199,14 +1206,13 @@ static int _vm_mapBelongs(const struct _process_t *proc, const void *ptr, size_t
 		return -1;
 	}
 
-#ifdef NOMMU
+#if 0 /* NOMMU */
 	if (f->process != proc) {
 		return -1;
 	}
 #endif
 
 #endif
-
 	return 0;
 }
 
--- a/vm/map.c
+++ b/vm/map.c
@@ -1176,7 +1176,7 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
 }
 
 
-static int _vm_mapBelongs(const struct _process_t *proc, const void *ptr, size_t size)
+static int map_belongs(const struct _process_t *proc, const void *ptr, size_t size)
 {
 	if (size == 0U) {
 		return 0;
@@ -1188,7 +1188,6 @@ static int _vm_mapBelongs(const struct _process_t *proc, const void *ptr, size_t
 	 * Disabled checking entry owner for now on NOMMU, as we can
 	 * receive memory from different maps and processes via msg
 	 * and RO elf segments may not be assigned to any process.
-	 * TODO: create memory objects for syspage programs
 	 */
 #elif !defined(NOMMU)
 	map_entry_t e, *f;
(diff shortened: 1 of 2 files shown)


vm/map.c
```diff
@@ -1190,21 +1187,46 @@ static int _vm_mapBelongs(const struct _process_t *proc, const void *ptr, size_t
...
+	spMap = syspage_mapAddrResolve((addr_t)ptr);
+	if (spMap == NULL) {
+		return -1;
+	}
+	map = vm_getSharedMap(spMap->id);
+	if (map == NULL) {
 		return -1;
 	}
+	f = lib_treeof(map_entry_t, linkage, lib_rbFind(&map->tree, &e.linkage));
 #endif
```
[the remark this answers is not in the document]
etiaro: Boundary check is not needed as lib_rbFind will ensure that.

etiaro:
Containment and overflow checks suggested by gemini will be introduced in separate PR, as they break [some CI tests](https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/495).

etiaro:
It's worth to notice this check introduces a silent assumption that checked region is contained within a single map.
I guess this assumption is reasonable, but correct me if I'm missing something.

julianuziemblo (commented):
Could you also add simple test to test the case where `map_belongs` should fail and the syscall produce `-EFAULT` (on NOMMU and MMU)?

etiaro:
> Could you also add simple test to test the case where map_belongs should fail and the syscall produce -EFAULT (on NOMMU and MMU)?

This is problematic, as currently kernel hangs on assertion when mapBelongs check fails in debug mode.
Our test suite doesn't seem to support checking for a system hang as the expected test behavior.

vm/map.c
```diff
@@ -1176,35 +1176,56 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
...
 	/* parasoft-suppress-next-line MISRAC2012-RULE_11_8 "Structure 'e' is used only for tree searching, pointer will not be modified. We cannot change vaddr member of the map_entry_t to const." */
 	e.vaddr = (void *)ptr;
 	e.size = size;
 
-	f = lib_treeof(map_entry_t, linkage, lib_rbFind(&proc->mapp->tree, &e.linkage));
-	if ((f == NULL) && (proc->imapp != NULL)) {
-		f = lib_treeof(map_entry_t, linkage, lib_rbFind(&proc->imapp->tree, &e.linkage));
-	}
+	(void)proc_lockSet(&map->lock);
+	f = lib_treeof(map_entry_t, linkage, lib_rbFind(&map->tree, &e.linkage));
```
julianuziemblo: so here, `map` is `proc->mapp` on MMU and the shared map resolved from syspage on NOMMU, yes?

etiaro: Exactly

etiaro:
Performance tests on `armv7r5f-zynqmp-som` show similar overhead for `pmap_isAllowed` call as for removed lock operations (>600 CPU cycles, ~1000 on pessimistic maps configuration, 1% runtime impact on full test campaign).

I'm open to suggestions for better name for "enable NOMMU mapBelongs checks" compilation flag.

Darchiv:
> Performance tests on `armv7r5f-zynqmp-som` show similar overhead for `pmap_isAllowed` call as for removed lock operations (>600 CPU cycles, ~1000 on pessimistic maps configuration, 1% runtime impact on full test campaign).
> 
> I'm open to suggestions for better name for "enable NOMMU mapBelongs checks" compilation flag.

@agkaminski What do you think?

agkaminski:
Assumption that the whole range fits within one map_entry_t breaks if someone would map two adjacent regions with different flags - then we cannot merge the entries.

This is like a super nitpicky edge-case, but still - check should pass it such case, current code will determine that the range does not fit.

There's also other, more architectural problem that is kinda outside of the scope of this PR - we do not check if user have write access to the region (we have no information if it is needed though - that's why it is architectural problem).

Regarding performance - security has it's costs I guess. I would rather use HW mechanism whether possible - it's pretty much free.

Current checks are still vulnerable - I can unmap the region using other thread after checks have passed.

etiaro:
> Assumption that the whole range fits within one map_entry_t breaks if someone would map two adjacent regions with different flags - then we cannot merge the entries.
> 
> This is like a super nitpicky edge-case, but still - check should pass it such case, current code will determine that the range does not fit.
> 

@agkaminski 
This PR only introduces validation according to NOMMU maps configuration (instead of no checks at all), no map_entry_t is checked here. So the assumption made is that the whole range fits into a single shared physical map.

There is a separate idea to introduce map containment check for MMU targets (instead of current overlap check), which would need to iterate over all related entries (usually 1 which gives hopes for low additional overhead in typical setups).

> There's also other, more architectural problem that is kinda outside of the scope of this PR - we do not check if user have write access to the region (we have no information if it is needed though - that's why it is architectural problem).

Yes, we noticed that, for both MMU and NOMMU this should be taken into consideration - following Zephyr or freeRTOS MPU, RO/RW argument could be added to mapBelongs and pmap_isAllowed if we want mapBelongs to provide at least basic protection (esp. for NOMMU where TOCTOU is not a problem as explained below).

> Current checks are still vulnerable - I can unmap the region using other thread after checks have passed.

Actually, thanks to map configuration being static, the checks introduced in this particular PR are not subject to TOCTOU vulnerabilities.

> Regarding performance - security has it's costs I guess. I would rather use HW mechanism whether possible - it's pretty much free.

HW check is the target implementation, on hold due to high risk of breaking mistakes. Also, it's not clear for now how well this can be done on NOMMU targets, e.g. it would require restricting kernel memory access with MPU and creating temporary configurations or disabling that kernel MPU in order to make message passing between maps work.

EDIT: I now see that I made a mistake during refactor. The intention was that when `NOMMU_MAPBELONGS` is not set, `mapBelongs` should remain a no-op on NOMMU targets. I'll push a fix in a moment.

agkaminski:
Ok, check against maps is indeed safe and valid, sorry for that. Check against owner won't work though (ok for now, as it is disabled)

Darchiv (approved):

PR #806, opened 2026-07-13
!proc: add schedGet and schedSet syscalls and make schedInfo thread-centric

adamgreloch:
TASK: RTOS-1364

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [X] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
https://github.com/phoenix-rtos/libphoenix/pull/504

### DIFF
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -2101,10 +2101,8 @@ int proc_threadsList(int n, threadinfo_t *info)
 }
 
 
-int proc_schedInfo(process_t *proc, int policy, sched_info_t *info)
+int proc_schedInfo(int policy, sched_info_t *info)
 {
-	LIB_ASSERT(proc != NULL, "null proc");
-
 	if (policy != SCHED_FIFO && policy != SCHED_RR && policy != SCHED_OTHER) {
 		return -EINVAL;
 	}
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -2108,7 +2108,7 @@ int proc_schedInfo(int policy, sched_info_t *info)
 	}
 
 	if (policy != SCHED_RR) {
-		return -ENOSYS;
+		return -ENOTSUP;
 	}
 
 	info->interval = SYSTICK_INTERVAL;
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -2087,7 +2087,7 @@ int proc_schedInfo(process_t *proc, int policy, sched_info_t *info)
 {
 	LIB_ASSERT(proc != NULL, "null proc");
 
-	if (policy < SCHED_FIFO || SCHED_OTHER < policy) {
+	if (policy != SCHED_FIFO && policy != SCHED_RR && policy != SCHED_OTHER) {
 		return -EINVAL;
 	}
 
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -728,18 +728,13 @@ static void _proc_threadSetPriority(thread_t *thread, u8 priority)
 }
 
 
-int proc_threadPriority(int signedPriority)
+int proc_threadPriority(thread_t *t, int signedPriority)
 {
-	thread_t *current;
 	spinlock_ctx_t sc;
 	int ret, reschedule = 0;
 	u8 priority;
 
-	if (signedPriority < -1) {
-		return -EINVAL;
-	}
-
-	if ((signedPriority >= 0) && ((size_t)signedPriority >= sizeof(threads_common.ready) / sizeof(threads_common.ready[0]))) {
+	if ((signedPriority < -1) || (signedPriority > (int)MAX_PRIO)) {
 		return -EINVAL;
 	}
 
@@ -747,29 +742,34 @@ int proc_threadPriority(int signedPriority)
 
 	hal_spinlockSet(&threads_common.spinlock, &sc);
 
-	current = _proc_current();
-
-	/* NOTE: -1 is used to retrieve the current thread priority only */
+	/* NOTE: -1 is used to retrieve the thread priority only */
 	if (signedPriority >= 0) {
-		if (priority < current->priority) {
-			current->priority = priority;
+		/*
+		 * _proc_threadSetPriority will clamp the priority to priorityBase, so it
+		 * must be updated prior to the call
+		 */
+		t->priorityBase = priority;
+
+		if (priority < t->priority) {
+			_proc_threadSetPriority(t, priority);
 		}
-		else if (priority > current->priority) {
+		else if (priority > t->priority) {
 			/* Make sure that the inherited priority from the lock is not reduced */
-			if ((current->locks == NULL) || (priority <= _proc_threadGetLockPriority(current))) {
-				current->priority = priority;
-				/* Trigger immediate rescheduling if the task has lowered its priority */
-				reschedule = 1;
+			if ((t->locks == NULL) || (priority <= _proc_threadGetLockPriority(t))) {
+				_proc_threadSetPriority(t, priority);
+
+				if (t == _proc_current()) {
+					/* Trigger immediate rescheduling if the task has lowered its priority */
+					reschedule = 1;
+				}
 			}
 		}
 		else {
 			/* No action required */
 		}
-
-		current->priorityBase = priority;
 	}
 
-	ret = (int)current->priorityBase;
+	ret = (int)t->priorityBase;
 
 	if (reschedule != 0) {
 		(void)hal_cpuReschedule(&threads_common.spinlock, &sc);
@@ -778,7 +778,7 @@ int proc_threadPriority(int signedPriority)
 		(void)hal_spinlockClear(&threads_common.spinlock, &sc);
 	}
 
-	trace_eventThreadPriority(proc_getTid(current), current->priority);
+	trace_eventThreadPriority(proc_getTid(t), t->priority);
 
 	return ret;
 }
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -311,6 +311,24 @@ thread_t *threads_findThread(int tid)
 }
 
 
+int proc_firstThreadTid(process_t *proc)
+{
+	spinlock_ctx_t sc;
+	int tid;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	if (proc->threads == NULL) {
+		tid = -1;
+	}
+	else {
+		tid = proc_getTid(proc->threads);
+	}
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+
+	return tid;
+}
+
+
 void threads_put(thread_t *thread)
 {
 	int refs;
@@ -2100,6 +2118,45 @@ int proc_schedInfo(process_t *proc, int policy, sched_info_t *info)
 	info->maxPriority = (int)MAX_PRIO;
 
 	return EOK;
+}
+
+
+int proc_schedGet(thread_t *t, sched_params_t *params)
+{
+	spinlock_ctx_t sc;
+	int priority, priorityBase;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	priority = (int)t->priority;
+	priorityBase = (int)t->priorityBase;
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+
+	params->priority = priority;
+	params->priorityBase = priorityBase;
+	params->policy = SCHED_RR;
+
+	return EOK;
+}
+
+
+int proc_schedSet(thread_t *t, int policy, sched_params_t *params)
+{
+	int err;
+
+	if (policy != SCHED_FIFO && policy != SCHED_RR && policy != SCHED_OTHER) {
+		return -EINVAL;
+	}
+
+	if (policy != SCHED_RR) {
+		return -ENOTSUP;
+	}
+
+	if (params->priorityBase < 0) {
+		return -EINVAL;
+	}
+
+	err = proc_threadPriority(t, params->priorityBase);
+	return err < 0 ? err : EOK;
 }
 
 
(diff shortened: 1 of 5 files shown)


syscalls.c
```diff
@@ -372,34 +372,137 @@ int syscalls_priority(u8 *ustack)
...
+
+		if (t->process != proc) {
+			err = -EINVAL;
+			break;
+		}
+
+		err = proc_schedGet(t, params);
+	} while (0);
+
+	if (t != NULL && t != current) {
```
ziemleszcz: `threads_put()` is not called if tid is set to match `proc_current()`.

syscalls.c
```diff
@@ -372,34 +372,137 @@ int syscalls_priority(u8 *ustack)
...
+		if (t->process != proc) {
+			err = -EINVAL;
+			break;
+		}
+
+		err = proc_schedSet(t, policy, params);
+	} while (0);
+
+	if (t != NULL && t != current) {
+		(void)threads_put(t);
```
ziemleszcz: `threads_put()` is not called if tid is set to match `proc_current()`.

proc/threads.c
```diff
@@ -727,29 +726,34 @@ int proc_threadPriority(int signedPriority)
 
 	hal_spinlockSet(&threads_common.spinlock, &sc);
 
-	current = _proc_current();
-
 	/* NOTE: -1 is used to retrieve the current thread priority only */
```
ziemleszcz: This comment is no longer true (not the current thread but thread t).

proc/threads.c
```diff
@@ -727,29 +726,34 @@ int proc_threadPriority(int signedPriority)
...
 		}
 		else {
 			/* No action required */
 		}
 
-		current->priorityBase = priority;
+		t->priorityBase = priority;
+
+		if (t != _proc_current()) {
+			_proc_threadSetPriority(t, priority);
```
ziemleszcz: `t->priority` must not be modified before calling `_proc_threadSetPriority()`. This function relies on the original value to identify thread's current `ready` queue.

proc/threads.c
```diff
@@ -2053,6 +2057,41 @@ int proc_schedInfo(process_t *proc, int policy, sched_info_t *info)
...
+
+	if (policy != SCHED_FIFO && policy != SCHED_RR && policy != SCHED_OTHER) {
+		return -EINVAL;
+	}
+
+	if (policy != SCHED_RR) {
+		return -ENOSYS;
+	}
+
+	err = proc_threadPriority(t, params->priorityBase);
```
ziemleszcz, line 2156: Note that params->priorityBase==-1 has different meaning. Should we reject -1 (return EINVAL)?

proc/threads.c
adamgreloch: uh, obviously, thanks for catching that

adamgreloch:
reverting to draft for now - will need to rethink the pid/tid relationship in the kernel for the libphoenix changes first.

syscalls.c
```diff
@@ -370,34 +370,149 @@ int syscalls_priority(u8 *ustack)
...
+	if (err < 0) {
+		return err;
+	}
+
+	err = proc_schedSet(t, policy, params);
+
+	putPidTid(pid, tid, proc, t);
 
 	return err;
 }
```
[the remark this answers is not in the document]
adamgreloch: This is a common issue with our syscalls. Leaving as-is for now.

proc/threads.c
```diff
@@ -2083,6 +2083,45 @@ int proc_schedInfo(process_t *proc, int policy, sched_info_t *info)
...
+	priority = (int)t->priority;
+	priorityBase = (int)t->priorityBase;
+	hal_spinlockClear(&threads_common.spinlock, &sc);
+
+	params->priority = priority;
+	params->priorityBase = priorityBase;
+	params->policy = SCHED_RR;
+
+	return EOK;
+}
```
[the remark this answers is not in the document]
adamgreloch: @gemini-code-assist how is this an issue if the params is provided by the userspace? even if it was copied to intermediary kernel buffer, the buffer contents would mirror the userspace garbage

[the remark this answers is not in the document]
adamgreloch: @gemini-code-assist `params` comes from userspace, not the kernel stack.

proc/threads.c
ziemleszcz, line 1069: Suggestion:
```
LIB_ASSERT((spinlock == NULL) == (scp == NULL), "spinlock and scp must both be NULL or both be non-NULL");
```

adamgreloch: Seems out of scope of this PR

ziemleszcz: Sorry, wrong PR. Too many reviews at once.

adamgreloch: No worries, thanks for taking the time

syscalls.c
```diff
@@ -370,34 +370,138 @@ int syscalls_priority(u8 *ustack)
...
-		(void)proc_put(proc);
+		return -EFAULT;
+	}
+
+	err = targetGet(pid, tid, &proc, &t);
+	if (err < 0) {
+		return err;
+	}
+
+	err = proc_schedInfo(t, policy, info);
```
ziemleszcz: Why do we even need `t` here? It seems that these calls don't need it at all:

sched_get_priority_max()
sched_get_priority_min()
pthread_attr_setschedparam()

These calls require only `pid`:

sched_rr_get_interval()
sched_getparam()
sched_setparam()
sched_getscheduler()

These require only `tid`:

pthread_setschedprio()
pthread_getschedparam()
pthread_setschedparam()

`threads_findThreadInProc()` has some issues, and `targetGet()/targetPut()` could be simplified if you handle only the actual cases: pid only or tid only.

adamgreloch: We need `t`, because the scheduling is thread-centric, so ultimately some thread must be picked (this was discussed in the RTOS-1397 issue). The complexity of targetGet()/targetPut() is exactly why they exist - they translate the pid/tid POSIX mess into kernel-native concept as soon as possible, in a common way for all the sched syscalls.

ziemleszcz: `proc_schedInfo()` is not using `t`.

adamgreloch: Currently it doesn't, because we don't support anything other than SCHED_RR. If we were to assume current state of the implementation, this whole syscall would be pointless, as we could just define the policies statically.

`proc_schedInfo()` may use `t` if:
- we implement per-partition policy ranges
- per-thread time quantums get implemented (sched_info_t:interval is then per-thread)

adamgreloch: After reconsideration, schedInfo is unlikely to need `t` ever. In per-partition case, we actually can depend on `proc_current()`

proc/threads.c
```diff
@@ -311,6 +311,24 @@ thread_t *threads_findThread(int tid)
...
 
 
+thread_t *proc_firstThread(process_t *proc)
+{
+	thread_t *t;
+	spinlock_ctx_t sc;
+
+	(void)proc_lockSet(&threads_common.lock);
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	t = proc->threads;
```
ziemleszcz: `t` may be already thread-to-be-destroyed (with `refs==0`) - see L343 in `threads_put()`. Incrementing `refs` will save it from being destroyed.

syscalls.c
```diff
@@ -375,34 +375,139 @@ int syscalls_priority(u8 *ustack)
...
+static int targetGet(int pid, int tid, process_t **outProcess, thread_t **outThread)
+{
+	process_t *proc = NULL;
+	thread_t *t;
+
+	if (pid == 0 && tid == 0) {
+		t = proc_current();
+	}
+	else if (tid != 0) {
+		t = threads_findThread(tid);
```
ziemleszcz: `threads_findThread()` may return NULL.

syscalls.c
```diff
@@ -375,34 +375,139 @@
...
+	}
+	else if (pid != 0) {
```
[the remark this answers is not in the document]
ziemleszcz: Correct.

syscalls.c
```diff
@@ -375,34 +375,139 @@ int syscalls_priority(u8 *ustack)
...
+		if (proc == NULL) {
+			return -ESRCH;
+		}
+		t = proc_firstThread(proc);
+		if (t == NULL) {
+			(void)proc_put(proc);
+			return -ESRCH;
+		}
+	}
+	else {
```
ziemleszcz: Dead code.

syscalls.c
```diff
@@ -375,34 +375,139 @@ int syscalls_priority(u8 *ustack)
...
+	*outThread = t;
+
+	return EOK;
+}
+
+
+static void targetPut(int tid, process_t *proc, thread_t *t)
+{
+	if (tid != 0) {
+		threads_put(t);
```
ziemleszcz: Leaks thread taken by `proc_firstThread()`.

proc/threads.c
adamgreloch: It will save it from being destroyed for the duration of the syscall, which seems pretty normal? Every usage of threads_findThread will cause the same

syscalls.c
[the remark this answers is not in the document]
adamgreloch: wow, epic brain lag

proc/threads.c
ziemleszcz: Look closely at L343. The problem is that the thread is already removed from the tree (before you increment `refs`).

ziemleszcz: `threads_findThread()` get thread from the tree so it doesn't have this problem.

syscalls.c
```diff
@@ -375,34 +375,129 @@ int syscalls_priority(u8 *ustack)
...
+			}
+			tid = proc_firstThreadTid(proc);
+			(void)proc_put(proc);
+			if (tid < 0) {
+				return NULL;
+			}
+		}
+
+		t = threads_findThread(tid);
+	}
```
ziemleszcz, line 436: If `pid != 0` we should check that `t` belongs to `pid`.

adamgreloch: pid != 0 && tid !=0 is redundant, added a return with EINVAL in this case.

syscalls.c
```diff
@@ -375,34 +375,143 @@ int syscalls_priority(u8 *ustack)
...
+		 */
+		return -EINVAL;
+	}
+
+	if (pid == 0 && tid == 0) {
+		*resThread = proc_current();
+		return EOK;
+	}
+
+	if (tid == 0) {
```
ziemleszcz, line 415: Consider adding this so that `sched_setparam(0, …)` and `sched_setparam(getpid(), …)` target the same thread:
```
if (tid == 0 && pid == process_getPid(proc_current()->process)) {
    tid = proc_getTid(proc_current());
}

```

syscalls.c
```diff
@@ -375,34 +375,143 @@ int syscalls_priority(u8 *ustack)
...
+	if (pid != 0 && tid != 0) {
+		/*
+		 * tid uniquely identifies a thread. Passing both pid and tid nonempty is
+		 * redundant and likely an error.
+		 */
+		return -EINVAL;
+	}
+
+	if (pid == 0 && tid == 0) {
+		*resThread = proc_current();
```
ziemleszcz: Consider just setting `tid= threads_findThread(proc_getTid(proc_current()))` here. This way we always take the thread ref and `targetPut()` is reduced to `threads_put()`.

syscalls.c
```diff
@@ -375,34 +375,135 @@ int syscalls_priority(u8 *ustack)
...
+				return -ESRCH;
+			}
+		}
+	}
+
+	t = threads_findThread(tid);
+	if (t == NULL) {
+		return -ESRCH;
+	}
+
```
ziemleszcz, line 442: Please add check `t->process == NULL` to reject threads created by the kernel (idle, usrv). I think is currently the only way to distinguish kernel threads from user threads. Make sure to add some comment.

proc/threads.c
```diff
@@ -2103,6 +2119,45 @@ int proc_schedInfo(process_t *proc, int policy, sched_info_t *info)
...
+int proc_schedSet(thread_t *t, int policy, sched_params_t *params)
+{
+	int err;
+
+	if (policy != SCHED_FIFO && policy != SCHED_RR && policy != SCHED_OTHER) {
+		return -EINVAL;
+	}
+
+	if (policy != SCHED_RR) {
+		return -ENOSYS;
```
ziemleszcz: ENOTSUP?

proc/threads.c
```diff
@@ -739,37 +756,38 @@ int proc_threadPriority(int signedPriority)
 		return -EINVAL;
 	}
 
-	if ((signedPriority >= 0) && ((size_t)signedPriority >= sizeof(threads_common.ready) / sizeof(threads_common.ready[0]))) {
+	if ((signedPriority >= 0) && ((size_t)signedPriority > MAX_PRIO)) {
```
ziemleszcz: Could be combined with previous check and simplified to:
```
if ((signedPriority < -1) || (signedPriority > (int)MAX_PRIO)) {
    return -EINVAL;
}
```

proc/threads.c
```diff
@@ -739,37 +756,38 @@ int proc_threadPriority(int signedPriority)
...
 	hal_spinlockSet(&threads_common.spinlock, &sc);
 
-	current = _proc_current();
-
-	/* NOTE: -1 is used to retrieve the current thread priority only */
+	/* NOTE: -1 is used to retrieve the thread priority only */
 	if (signedPriority >= 0) {
-		if (priority < current->priority) {
-			current->priority = priority;
+		t->priorityBase = priority;
```
ziemleszcz, line 769: That's tricky code. Please add a comment explaining why `priorityBase` must be set before the `_proc_threadSetPriority()` call.

ziemleszcz (approved):

PR #808, opened 2026-07-14
posix: implement pid == -1 and pid == 0 cases for kill

dwhitting:
## Description
Implements the `pid == -1` and `pid == 0` cases for `kill`/`tkill` per POSIX.1-2017:

- `pid == 0`: sends the signal to all processes sharing the caller's process group (reuses existing `posix_killGroup`).
- `pid == -1`: sends the signal to all processes (new `posix_killAll` function, following the same loop pattern as `posix_killGroup`).

Note: neither case performs additional permission checks beyond what `posix_killGroup` already does, since the kernel doesn't currently track per-process signal permissions. Happy to adjust if you'd like something different here.

## Motivation and Context
Previously, `kill(0, sig)` returned `-ENOSYS` and `kill(-1, sig)` returned `-ESRCH`, both were unimplemented (marked with a TODO in the code) rather than following POSIX-defined behavior.

Fixes #1692

## Types of changes

## How Has This Been Tested?
- [ ] Tested by hand on: ia32-generic-qemu (compiles cleanly; behavioral testing pending — planning to add a test case per phoenix-rtos-tests, tracked separately)

## Checklist:

## Special treatment

### DIFF
--- a/posix/posix.c
+++ b/posix/posix.c
@@ -2515,17 +2515,28 @@ static int posix_killOne(pid_t pid, int tid, int sig)
 }
 
 
-static int posix_killGroup(pid_t pgid, int sig)
+enum killMode { tkill_group, tkill_all };
+
+
+static int posix_killMulti(enum killMode mode, pid_t pgid, int sig)
 {
 	process_info_t *pinfo;
 	rbnode_t *node;
+	pid_t selfPid = process_getPid(proc_current()->process);
 
 	(void)proc_lockSet(&posix_common.lock);
 	for (node = lib_rbMinimum(posix_common.pid.root); node != NULL; node = lib_rbNext(node)) {
 		pinfo = lib_treeof(process_info_t, linkage, node);
 
-		if (pinfo->pgid == pgid) {
-			(void)proc_sigpost(pinfo->process, sig);
+		if (mode == tkill_group) {
+			if (pinfo->pgid == pgid) {
+				(void)proc_sigpost(pinfo->process, sig);
+			}
+		}
+		else {
+			if ((pinfo->process != 1) && (pinfo->process != selfPid)) {
+				(void)proc_sigpost(pinfo->process, sig);
+			}
 		}
 	}
 	(void)proc_lockClear(&posix_common.lock);
@@ -2536,22 +2547,32 @@ static int posix_killGroup(pid_t pgid, int sig)
 
 int posix_tkill(pid_t pid, int tid, int sig)
 {
+	process_info_t *pinfo;
+	pid_t myPgid;
+
 	TRACE("tkill(%p, %d, %d)", pid, tid, sig);
 
 	if ((sig < 0) || (sig > NSIG)) {
 		return -EINVAL;
 	}
 
-	/* TODO: handle pid = 0 */
 	if (pid == 0) {
-		return -ENOSYS;
+		pinfo = pinfo_find(process_getPid(proc_current()->process));
+		if (pinfo == NULL) {
+			return -ESRCH;
+		}
+
+		myPgid = pinfo->pgid;
+		pinfo_put(pinfo);
+
+		return posix_killMulti(tkill_group, myPgid, sig);
 	}
 
 	if (pid == -1) {
-		return -ESRCH;
+		return posix_killMulti(tkill_all, 0, sig);
 	}
 
-	return (pid > 0) ? posix_killOne(pid, tid, sig) : posix_killGroup(-pid, sig);
+	return (pid > 0) ? posix_killOne(pid, tid, sig) : posix_killMulti(tkill_group, -pid, sig);
 }
 
 


dwhitting:
@julianuziemblo This is ready for review. Since I don't have permission to assign, tagging you here per your earlier note on #1692.

posix/posix.c
```diff
@@ -2533,6 +2533,20 @@ static int posix_killGroup(pid_t pgid, int sig)
...
+
+	(void)proc_lockSet(&posix_common.lock);
+	for (node = lib_rbMinimum(posix_common.pid.root); node != NULL; node = lib_rbNext(node)) {
+		pinfo = lib_treeof(process_info_t, linkage, node);
+		(void)proc_sigpost(pinfo->process, sig);
+	}
+	(void)proc_lockClear(&posix_common.lock);
+
+	return EOK;
+}
```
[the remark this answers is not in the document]
dwhitting: Good catch, thanks — fixed in the latest commit. Excluded both PID 1 and the caller's own pid from the broadcast.

posix/posix.c
```diff
@@ -2542,13 +2559,25 @@ int posix_tkill(pid_t pid, int tid, int sig)
 		return -EINVAL;
 	}
 
-	/* TODO: handle pid = 0 */
+	/* Handle pid == 0: send signal to all processes in caller's process group */
 	if (pid == 0) {
-		return -ENOSYS;
+		process_info_t *pinfo;
```
etiaro: As you can see in the rest of this submodule, [our guidelines](https://github.com/phoenix-rtos/phoenix-rtos-doc/blob/5083598b45ec21355a90467656c2c101ed217ea4/coding/index.md#:~:text=Local%20variables%20should%20be%20defined%20before%20the%20function%20code%20in%20the%20kernel.) recommend defining variables at the beginning of the functions.

posix/posix.c
```diff
@@ -2533,6 +2533,23 @@ static int posix_killGroup(pid_t pgid, int sig)
 	return EOK;
 }
 
+static int posix_killAll(int sig)
+{
+	process_info_t *pinfo;
+	rbnode_t *node;
+	pid_t self_pid = process_getPid(proc_current()->process);
```
etiaro: [We prefer using camelCase for variable names.](https://github.com/phoenix-rtos/phoenix-rtos-doc/blob/5083598b45ec21355a90467656c2c101ed217ea4/coding/index.md#:~:text=For%20variable%20names%20longer%20than%20one%20word%2C%20the%20camelCase%20format%20should%20be%20used.)

posix/posix.c
```diff
@@ -2533,6 +2533,23 @@ static int posix_killGroup(pid_t pgid, int sig)
 	return EOK;
 }
 
+static int posix_killAll(int sig)
```
etiaro: I would include the fact that it leaves current process alive in the name or at least in the comment. You could also specify that this is a posix requirement

posix/posix.c
```diff
@@ -2533,6 +2533,23 @@ static int posix_killGroup(pid_t pgid, int sig)
 	return EOK;
 }
```
etiaro, line 2552: We do leave 2 empty lines between functions, please keep that style.

etiaro (commented):
The logic looks fine, I have a few minor comments to the style.
It would also be nice if you squashed your commits into one (we do not squash your commits on merge).
Also, we have a rule about comments saying "why" not "what", in this case you can change them to say that this is how POSIX specifies kill functions, or simply remove them.

posix/posix.c
```diff
@@ -2533,6 +2533,23 @@ static int posix_killGroup(pid_t pgid, int sig)
...
+	for (node = lib_rbMinimum(posix_common.pid.root); node != NULL; node = lib_rbNext(node)) {
+		pinfo = lib_treeof(process_info_t, linkage, node);
+		if ((pinfo->process != 1) && (pinfo->process != self_pid)) {
+			(void)proc_sigpost(pinfo->process, sig);
+		}
+	}
+	(void)proc_lockClear(&posix_common.lock);
+
+	return EOK;
+}
```
julianuziemblo, line 2584: Those 2 functions (`posix_killAll` and `posix_killGroup`) are essentially the same, the only difference seems to be the condition on which we sigpost. Consider merging them into one with e.g. an enum value (`tkill_group`, `tkill_all`) or other similar solution.

posix/posix.c
dwhitting: Fixed. Thanks for flagging.

posix/posix.c
dwhitting: @julianuziemblo @etiaro Addressed all feedback so far: merged posix_killGroup/posix_killAll into posix_killMulti with a killMode enum, removed the "what" comments, fixed blank-line spacing, renamed variables to camelCase, and moved variable declarations to the top of posix_tkill. Also squashed commits into one. Thank you!

ziemleszcz:
Please resolve the conflicts.

posix/posix.c
```diff
@@ -2549,45 +2515,64 @@ static int posix_killOne(pid_t pid, int tid, int sig)
...
+			if (pinfo->pgid == pgid) {
+				(void)proc_sigpost(pinfo->process, sig);
+			}
+		}
+		else {
+			if ((pinfo->process != 1) && (pinfo->process != selfPid)) {
+				(void)proc_sigpost(pinfo->process, sig);
+			}
 		}
 	}
```
julianuziemblo: this could be a `switch`, so you don't have to do this awkward `else { if (...) }` to avoid MISRA conflicts (because we'd not have to make more changes if a new case appears here somehow, and we'd get a warning if we added another value to the enum that it's not handled when using a switch). The "default" branch should probably be left empty.

posix/posix.c
```diff
@@ -2549,45 +2515,64 @@ static int posix_killOne(pid_t pid, int tid, int sig)
...
 
-	return err;
+	return EOK;
 }
 
 
 int posix_tkill(pid_t pid, int tid, int sig)
 {
+	process_info_t *pinfo;
+	pid_t myPgid;
```
julianuziemblo: nitpick: keep the variable names similar: you had `selfPid`, so here the more fitting name would IMO be `selfPgid` (or `pgid` as it's used immediately after assignment)

dwhitting: @julianuziemblo Resolved the merge conflicts, converted the mode dispatch to a switch statement (also merged in upstream's err-tracking improvement to posix_killGroup), and renamed myPgid to selfPgid for consistency. Thanks!

adamgreloch:
Hi, please rebase your changes onto the master instead of merging

PR #809, opened 2026-07-15
hal/{aarch64,arm}: fix possible race condition in hal_cpuReschedule

### LINKED ISSUE #1386: aarch64a53-zynqmp-qemu: some psh testcases freeze during test campaign
Commit 666681e1f175eec5da9d1ec04be1a684183959c1
Happens both locally and on CI only during full test campaign once every 5-10 runs, only on this specific target.
Upon running only psh tests didn't happen a single time during 50+ runs, same after increasing duration of that part of testcase 10fold if running only psh tests.
Increasing amount of random character commands 10fold didn't affect probability of freeze during full campaign.
Freezes in random places of same moment in each of those testcases (always after sending some random characters).

Image

Image

jmaksymowicz:
Fix a possible race condition between `hal_cpuReschedule` and `proc_reap`.

## Description
A race condition has been identified in `hal_cpuReschedule` implementation on aarch64 and armv7a that can manifest itself at the end of the `exit` syscall.

The unwanted interleave starts when `hal_cpuReschedule`, is called at the end of handling the `exit` syscall on core A. Within the function the spinlock is cleared, then on core B the thread running the `proc_reap` function is woken up. `proc_reap` then proceeds to free the resources belonging to the exited thread, including unmapping its kernel stack. However, the kernel stack is still in use by core A. This results in either:

1. An infinite loop of exception handling: the first step of exception handling involves pushing registers onto the stack - the stack is inaccessible due to being unmapped - this causes another exception in an infinite loop.
2. If another thread's kernel stack is mapped into the same memory various other errors may occur - e.g. core A handling the exception from step 1 may corrupt the stack contents of another kernel thread or core A may try to restore from an invalid stored state.

The issue is most pronounced on `aarch64a53-zynqmp-qemu` due to the fact that it uses spinlocks with `wfe`/ `sev` and emulation can execute large chunks of code from one CPU before switching to another.

On `armv7a-zynq7000-qemu` this may not occur due to the fact that emulation is done on many cores in parallel and spinlocks don't use `wfe` / `sev`.

On real hardware this may not occur very frequently due to code being executed in parallel, so the first CPU always "wins the race".

The fact that `exit` is called frequently within the test suite increases the likelihood of hitting this unwanted interleave - which is why the issue occurs frequently in testing, but seemingly not in practical applications on that platform.

The fix from `armv7a` was also applied to `armv7r` and `armv8r` for consistency - even though these platforms don't work in SMP so this race condition will never occur.

On `ia32` and `riscv64` platforms the implementation of `hal_cpuReschedule` is different and not susceptible to this error.

## Motivation and Context
May fix https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1386 and https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1387 - needs more testing

YT: RTOS-1392

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: aarch64a53-zynqmp-qemu, armv7a9-zynq7000-qemu

## Checklist:

## Special treatment

### DIFF
--- a/hal/aarch64/_exceptions.S
+++ b/hal/aarch64/_exceptions.S
@@ -220,22 +220,30 @@ _syscalls_dispatch:
 	bic X_SPSR_EL1, X_SPSR_EL1, #NO_INT /* Re-enable interrupts in the caller */
 	str X_SPSR_EL1, [sp, 0x220]
 
-	ldr x2, [sp, #0x230]  /* Load x0 -> pointer to spinlock */
-	cbz x2, 1f /* If pointer not NULL, clear the spinlock - but don't restore the DAIF */
-	mov w1, #1
-	stlrb w1, [x2] /* Global monitor clear generates an event, SEV not necessary */
-1:
+	ldr x19, [sp, #0x230]  /* Load x0 -> pointer to spinlock */
+	mov x1, sp            /* argument to function - cpu_context_t* */
 	str xzr, [sp, #0x230] /* Store 0 into x0 as return value (EOK) */
+	cbz x19, .Lno_spinlock
 
-	mov x1, sp /* argument to function - cpu_context_t* */
+	bl _threads_schedule
+	ldr x1, [sp] /* Load pointer to new context (must be done BEFORE clearing spinlock) */
 
+	/* Clear the given spinlock */
+	mov w0, #1
+	stlrb w0, [x19] /* Global monitor clear generates an event, SEV not necessary */
+	b .Lcontext_ready
+
+.Lno_spinlock:
 	/* Arguments to threads_schedule:
 	 * * x0 - n (unused)
 	 * * x1 - pointer to CPU context
 	 * * x2 - arg (unused)*/
 	bl threads_schedule
+	ldr x1, [sp] /* Load pointer to new context */
 
-	ldr x1, [sp]
+.Lcontext_ready:
+	/* NOTE: at this point our stack pointer may be invalid (e.g. if this is the ending of exit() syscall)
+	 * No code beyond this point until `eret` may use the stack! */
 	unlock_scheduler x0
 	b _hal_cpuRestoreCtx
 .size _exceptions_dispatch, .-_exceptions_dispatch
--- a/hal/armv7a/_interrupts.S
+++ b/hal/armv7a/_interrupts.S
@@ -159,14 +159,11 @@ hal_cpuReschedule:
 
 	mrs r4, cpsr
 
-	/* If spinlock not NULL, clear it */
+	/* If spinlock was given, restore spinlock context into SPSR */
 	cmp r0, #0
 	beq 1f
-	mov r3, #1
-	dmb
-	strb r3, [r0]
-	ldr r1, [r1]
 
+	ldr r1, [r1]
 	bic r4, #0xff
 	and r1, #0xff
 	orr r4, r4, r1
@@ -179,13 +176,28 @@ hal_cpuReschedule:
 
 	push_fpu_state r4
 
-	sub r1, sp, #8
+	sub r1, sp, #8 /* argument to function - cpu_context_t* */
 	push {r1}
 	push {r1}
 
+	movs r4, r0
+	beq .Lno_spinlock
+
+	/* Call the non-locked version of threads_schedule */
+	blx _threads_schedule
+	ldr sp, [sp] /* Load pointer to new context (must be done BEFORE clearing spinlock) */
+
+	/* Clear the given spinlock */
+	mov r0, #1
+	dmb
+	strb r0, [r4]
+	b .Lcontext_ready
+
+.Lno_spinlock:
 	blx threads_schedule
+	ldr sp, [sp] /* Load pointer to new context */
 
-	ldr sp, [sp]
+.Lcontext_ready:
 	add sp, sp, #8
 
 	unlock_scheduler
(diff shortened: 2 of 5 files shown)


lukileczo:
Are you sure this doesn't also affect other targets? From what I can see the implementations are pretty similar, e.g. on `riscv64` I also clear the spinlock before loading the stack pointer
https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/0d8d2835fc575c1a20e29b257f4a1241fa05b748/hal/riscv64/_interrupts.S#L471-L494
which you deemed the cause of the issue.

jmaksymowicz:
@lukileczo 

> Are you sure this doesn't also affect other targets? From what I can see the implementations are pretty similar, e.g. on `riscv64` I also clear the spinlock before loading the stack pointer
> ...
> which you deemed the cause of the issue.

Yes, it seems that `riscv` may also be affected, but to a much lesser extent. After clearing the spinlock there's only one jump instruction and then the new stack pointer is loaded from our current "doomed" stack, so there's much less room for the unwanted interleave described above. Considering that the scheduler is locked until after loading the new SP this would only happen in the rare case that the reaper thread is already running (so doesn't need to be scheduled in) and would need to work very fast to unmap the "doomed" stack before we manage to load from it.

On aarch64 and armv7a the spinlock was cleared a lot earlier, even before calling `threads_schedule` and taking the scheduler lock, leaving a lot of time and possibility for error.

hal/aarch64/_exceptions.S
```diff
@@ -220,22 +220,33 @@ _syscalls_dispatch:
...
-	mov w1, #1
-	stlrb w1, [x2] /* Global monitor clear generates an event, SEV not necessary */
-1:
+	ldr x19, [sp, #0x230]  /* Load x0 -> pointer to spinlock */
+	mov x1, sp            /* argument to function - cpu_context_t* */
 	str xzr, [sp, #0x230] /* Store 0 into x0 as return value (EOK) */
+	cbz x19, .Lno_spinlock
+
+	/* Call the non-locked version of threads_schedule.
+	 * TODO: In this and other implementations of hal_cpuReschedule() there's an implicit assumption
```
agkaminski: This assumption is well based - it is there by design. What should be made is to make it clear and explicit to avoid mistakes. Perhaps skip passing spinlock as an argument and pass flag clear/no clear instead?

agkaminski: For now I think the comment about what spinlock is passed to the `hal_cpuReschedule` may be duplicated in other archs to avoid confusion why we execute `_threads_schedule` with _some_ spinlock, when in fact we always do with `threads_common.spinlock`

agkaminski (commented):
That must've been quite a pain to find, good job. One nitpick about assumptions and comments.

hal/aarch64/_exceptions.S
```diff
@@ -220,22 +220,33 @@ _syscalls_dispatch:
...
+	mov w0, #1
+	stlrb w0, [x19] /* Global monitor clear generates an event, SEV not necessary */
+	b .Lcontext_ready
 
+.Lno_spinlock:
 	/* Arguments to threads_schedule:
 	 * * x0 - n (unused)
 	 * * x1 - pointer to CPU context
 	 * * x2 - arg (unused)*/
 	bl threads_schedule
```
etiaro, line 241: When we enter `hal_cpuReschedule` without `threads_common.spinlock`, after returning from `threads_schedule()` call the stack we use theoretically could be already freed, similarly to the described riscv64 situation.

This is probably really unlikely, especially since there is few instructions between spinlock clear here, but it could be possible when we consider that threads_schedule can add current thread to threads_common.ghosts (and it really does happen e.g. in `process_exception()` code path).

We probably could get an interrupt after clearing threads_common spinlock in threads_schedule(?), which would give enough time for the reaper to destroy our stack? (I guess in riscv64 case interrupts are disabled)

Correct me if i'm wrong, this is purely theoretical (meaning: I did not reproduce that)

ziemleszcz (dismissed):
LGTM. Consider updating the comment as @agkaminski  suggested.
Also update the PR description to mention that other architectures could theoretically be affected, although the race window so narrow that it's unlikely to occur in practice.
It might also be worth mentioning the similar narrow race windows in interrupt and exception handling that @etiaro  pointed out.

hal/aarch64/_exceptions.S
jmaksymowicz: I think you're right, in `process_exception` we need to call `hal_cpuReschedule` under spinlock - or just call `proc_threadEnd` instead, as it will handle things properly.

> We probably could get an interrupt after clearing threads_common spinlock in threads_schedule(?), which would give enough time for the reaper to destroy our stack? (I guess in riscv64 case interrupts are disabled)

In all implementations of `hal_cpuReschedule` `threads_schedule` is called with interrupts disabled and they aren't enabled again until context is restored.

I will open a separate issue for fixing this and the riscv64 potential race conditions.

hal/aarch64/_exceptions.S
jmaksymowicz: I decided to add a comment to `hal/cpu.h` explaining how the function should be used - I think that's a good solution for now. Ofc there's possibility of user error but at least the requirement is stated explicitly.

agkaminski (approved):

ziemleszcz (approved):

PR #813, opened 2026-07-20
!threads: split cpuTime into user and system time

oI0ck:
TASK: RTOS-1382

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/include/sysinfo.h
+++ b/include/sysinfo.h
@@ -30,7 +30,8 @@ typedef struct _threadinfo_t {
 	pid_t ppid;
 
 	int load;
-	time_t cpuTime;
+	time_t systemTime;
+	time_t userTime;
 	int priority;
 	int state;
 	int vmem;
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -311,12 +311,45 @@ void threads_put(thread_t *thread)
 }
 
 
+void threads_updateCpuTime(thread_t *current, time_kind_t kind)
+{
+	spinlock_ctx_t sc;
+	time_t now, timeSpent;
+
+	if (current != NULL) {
+		hal_spinlockSet(&threads_common.spinlock, &sc);
+		now = _proc_gettimeRaw();
+
+		timeSpent = now - current->lastTime;
+
+		if (kind == time_system) {
+			current->systemTime += timeSpent;
+		}
+		else if (kind == time_user) {
+			current->userTime += timeSpent;
+		}
+		else {
+			/* No action required */
+		}
+
+		current->lastTime = now;
+		hal_spinlockClear(&threads_common.spinlock, &sc);
+	}
+}
+
+
 static void _threads_cpuTimeCalc(thread_t *current, thread_t *selected)
 {
 	time_t now = _proc_gettimeRaw();
 
 	if (current != NULL) {
-		current->cpuTime += now - current->lastTime;
+		if (current->process == NULL || hal_cpuSupervisorMode(current->context) == 1) {
+			current->systemTime += now - current->lastTime;
+		}
+		else {
+			current->userTime += now - current->lastTime;
+		}
+
 		current->lastTime = now;
 	}
 
@@ -574,7 +607,8 @@ int proc_threadCreate(process_t *process, startFn_t start, int *id, u8 priority,
 	t->utick = 0;
 	t->priorityBase = priority;
 	t->priority = priority;
-	t->cpuTime = 0;
+	t->systemTime = 0;
+	t->userTime = 0;
 	t->maxWait = 0;
 	proc_gettime(&t->startTime, NULL);
 	t->lastTime = t->startTime;
@@ -1925,7 +1959,7 @@ int proc_threadsIter(int n, proc_threadsListCb_t cb, void *arg)
 	thread_t *t;
 	map_entry_t *entry;
 	vm_map_t *map;
-	time_t now;
+	time_t now, cpuTime;
 	spinlock_ctx_t sc;
 	threadinfo_t tinfo;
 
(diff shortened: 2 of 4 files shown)


proc/threads.c
```diff
@@ -311,12 +311,43 @@
...
+		hal_spinlockClear(&threads_common.spinlock, &sc);
+		timeSpent = now - current->lastTime;
+
+		if (kind == kSystem) {
+			current->systemTime += timeSpent;
+		}
+		else if (kind == kUser) {
+			current->userTime += timeSpent;
+		}
+		else {
```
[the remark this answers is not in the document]
oI0ck: fixed

proc/threads.c
```diff
@@ -311,12 +311,43 @@
...
+}
+
+
 static void _threads_cpuTimeCalc(thread_t *current, thread_t *selected)
 {
 	time_t now = _proc_gettimeRaw();
 
 	if (current != NULL) {
-		current->cpuTime += now - current->lastTime;
+		if (current->process == NULL || hal_cpuSupervisorMode(current->context) == 1U) {
```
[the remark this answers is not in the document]
oI0ck: fixed

proc/threads.c
```diff
@@ -311,12 +311,44 @@ void threads_put(thread_t *thread)
...
 
+void threads_updateCpuTime(thread_t *current, threadTimeKind_t kind)
+{
+	spinlock_ctx_t sc;
+	time_t now, timeSpent;
+
+	if (current != NULL) {
+		hal_spinlockSet(&threads_common.spinlock, &sc);
+		now = _proc_gettimeRaw();
+		hal_spinlockClear(&threads_common.spinlock, &sc);
```
adamgreloch: Since `thread_t` already tracks `lastTime`, you may take `now` directly from `hal_timerGetUs()` and detect non-monotonicity in DEBUG by comparing `lastTime` with `now`, instead of calling `_proc_gettimeRaw()`. That way you wouldn't need a spinlock.

include/sysinfo.h
```diff
@@ -30,7 +30,8 @@ typedef struct _threadinfo_t {
 	pid_t ppid;
 
 	int load;
-	time_t cpuTime;
+	time_t systemTime;
+	time_t userTime;
```
adamgreloch, line 34: This is breaking, please mark the commit/PR as such

proc/threads.h
```diff
@@ -90,6 +91,11 @@ static inline int proc_getTid(const thread_t *t)
 	return t->idlinkage.id;
 }
 
+typedef enum threadTimeKind {
+	kSystem,
+	kUser,
+} threadTimeKind_t;
```
adamgreloch: Not sure about the naming of this enum. Usually named kernel enums follow `xyz_t` -> values: `xyz_a`, `xyz_b`, `xyz_c` etc.

proc/threads.c
oI0ck: Then I'm essentially reimplementing what `_proc_gettimeRaw` does, without writing to `threads_common.prev`.
I don't really get the point of it; sure, it will save some time, but then, what is this function for?

proc/threads.h
oI0ck: There are two conventions, one is the one you mentioned, the other one is the one I used.
I don't think we have guidelines regarding enum definition styles.

adamgreloch: New enums follow the style I have described, look in the code. Your convention is not used anywhere except for message types, which are exported. Besides, `kSystem` is simply cryptic.

proc/threads.c
adamgreloch: Why take spinlock twice for *each* syscall when you can simply not?

`_proc_gettimeRaw()` checks non-monotonicity in threads.c from a common `prev`, since there is no other place to store this information. In your case however, you already store `lastTime` in the thread, so you can track it from there. You are concerned about reimplementing a two line function.

proc/threads.c
```diff
@@ -311,12 +311,43 @@ void threads_put(thread_t *thread)
...
+		}
+		else if (kind == kUser) {
+			current->userTime += timeSpent;
+		}
+		else {
+		}
+
+		current->lastTime = now;
+	}
+}
```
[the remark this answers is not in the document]
etiaro: He had a point, didn't he? why aren't these fields synchronized?

proc/threads.c
adamgreloch: ok, @etiaro (and the tin man) have a point that systemTime and userTime should  be synchronized due to use in threadsinfo, so this becomes irrelevant

proc/threads.c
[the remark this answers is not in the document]
oI0ck: Right, I dismissed that as processes are not directly protected with the spinlock, but without it, they are not protected at all.

Alternatively, the spinlock could be held only for getting the time and the mutex for update of `current`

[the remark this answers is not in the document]
etiaro: This is a thread not a process (proc_current() is really confusing btw). I would avoid mutexes for short, deterministic constant-time operations, as these are much more expensive than spinlocks.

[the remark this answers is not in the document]
oI0ck: Right, `s/processes/threads`, the lock I meant is `threads_common.lock`. Sure, I'll extend the spinlock holding duration.

proc/threads.h
oI0ck: Corrected

proc/threads.c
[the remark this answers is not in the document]
oI0ck: Corrected

etiaro (changes requested):
Please ensure all kernel/userspace boundaries are included, as we don't always pass through a syscall return, e.g. through a few calls of `hal_jmp`.

Consider what should happen on execve syscalls (currently the time will go over to the exec'd process, but maybe that's fine? Is it standardized in any way?).

Another boundaries are user interrupts and exceptions (especially pageFault, as we pretend to support lazy-allocations which would make it happen in regular execution).
Not sure what to do about these, maybe nothing, maybe leave a memo in a comment at least.

oI0ck:
@etiaro Linux differentiates between 'user time' vs 'system time' and 'irq time', we can do that too I think.

etiaro:
Sound good.
Keep in mind that we need to keep irq response fast (to the point we're executing them on kernel stack as of now).
I guess it would be nice to have "time you spent in irq's" and/or "time that was taken away from your time slices for irq's sake". The latter one could help with jitter/WCET estimations.

PR #814, opened 2026-07-21
threads: introduce threads_halt function

oI0ck:
hal_cpuHalt() is used for two purposes, for putting the processor into low-power mode and waiting for an interrupt, and for 'halt and catch fire' use case. Even moreso, it is still done incorrectly on SMP systems, as that would only halt the core the fault happened on.

This begs for differentiation.
threads_halt() is a replacement for the second use case.

TASK: RTOS-1401

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/hal/armv7a/imx6ull/memtest.c
+++ b/hal/armv7a/imx6ull/memtest.c
@@ -300,5 +300,7 @@ void test_ddr(void)
 	test_ddrAll();
 #endif
 
-	hal_cpuHalt();
+	for (;;) {
+		hal_cpuHalt();
+	}
 }
--- a/lib/assert.c
+++ b/lib/assert.c
@@ -19,6 +19,7 @@
 #include "printf.h"
 #include "log/log.h"
 #include "hal/hal.h"
+#include "proc/threads.h"
 
 
 void lib_assertPanic(const char *func, int line, const char *fmt, ...)
@@ -38,8 +39,6 @@ void lib_assertPanic(const char *func, int line, const char *fmt, ...)
 #ifdef NDEBUG
 	hal_cpuReboot();
 #else
-	for (;;) {
-		hal_cpuHalt();
-	}
+	threads_halt();
 #endif
 }
--- a/lib/assert.c
+++ b/lib/assert.c
@@ -22,7 +22,7 @@
 #include "proc/threads.h"
 
 
-void lib_assertPanic(const char *func, int line, const char *fmt, ...)
+__attribute__((noreturn)) void lib_assertPanic(const char *func, int line, const char *fmt, ...)
 {
 	va_list ap;
 
--- a/proc/process.c
+++ b/proc/process.c
@@ -288,9 +288,7 @@ static void process_exception(unsigned int n, exc_context_t *ctx)
 
 	process_dumpException(n, ctx);
 
-	if (thread->process == NULL) {
-		hal_cpuHalt();
-	}
+	LIB_ASSERT_ALWAYS(thread->process != NULL, "exception in kernel")
 
 	(void)threads_sigpost(thread->process, thread, signal_kill);
 
@@ -307,9 +305,7 @@ static void process_illegal(unsigned int n, exc_context_t *ctx)
 	thread_t *thread = proc_current();
 	process_t *process = thread->process;
 
-	if (process == NULL) {
-		hal_cpuHalt();
-	}
+	LIB_ASSERT_ALWAYS(process != NULL, "exception in kernel")
 
 	(void)threads_sigpost(process, thread, signal_illegal);
 }
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -248,6 +248,20 @@ static int threads_timeintr(unsigned int n, cpu_context_t *context, void *arg)
 static void proc_lockForceUnlock(lock_t *lock, int doYield);
 
 
+__attribute__((noreturn)) void threads_halt(void)
+{
+	spinlock_ctx_t sc;
+
+	/* take the spinlock, and not release it; other threads will eventually hang trying to reschedule */
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	for (;;) {
+		hal_cpuHalt();
+	}
+
+	__builtin_unreachable();
+}
+
+
 static void thread_destroy(thread_t *thread)
 {
 	process_t *process;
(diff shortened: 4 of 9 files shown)


hal/armv7a/imx6ull/memtest.c
```diff
@@ -300,5 +300,7 @@ void test_ddr(void)
 	test_ddrAll();
 #endif
 
-	hal_cpuHalt();
+	for (;;) {
+		hal_cpuHalt();
+	}
```
julianuziemblo, line 305: this looks like threads_halt() for me (though it is unused code so it's kinda hard to tell)

proc/threads.c
```diff
@@ -248,6 +248,21 @@ static int threads_timeintr(unsigned int n, cpu_context_t *context, void *arg)
 static void proc_lockForceUnlock(lock_t *lock, int doYield);
 
 
+__attribute__((noreturn)) void threads_halt(void)
+{
+	spinlock_ctx_t sc;
+
+	/* steal the spinlock, so that remaining cores on SMP systems spin forever */
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+	hal_cpuDisableInterrupts(); /* if not already disabled */
```
julianuziemblo: It was my impression that hal_spinlockSet should disable interrupts on every architecture, so the second call is unnecessary. Isn't this the assumption we make when setting a spinlock on non-SMP targets anyway?

proc/threads.c
```diff
@@ -248,6 +248,21 @@ static int threads_timeintr(unsigned int n, cpu_context_t *context, void *arg)
...
+	for (;;) {
+		hal_cpuHalt();
+	}
+
+	__builtin_unreachable();
```
julianuziemblo, line 264: Probably should compile without this. On one hand I see the benefit of marking this point of the program as unreachable for further maintainers. On the other hand, we don't like __builtins very much (but we do use them anyway sometimes) - I'd remove it if not necessary, but you do you.

julianuziemblo (commented):
There are also some other places in which `hal_cpuHalt` seems like it was done to halt the processor: `lib/assert.c:41-43` (in debug mode), `test/proc.c:72-74` - let me know why these weren't done or do them if they were just overlooked.

hal/armv7a/imx6ull/memtest.c
adamgreloch: better not to call `threads_` from hal (if even possible)

proc/threads.c
adamgreloch: actually, this specific builtin is very much welcome and common in hal code

proc/threads.c
oI0ck: right, will change

proc/process.c
```diff
@@ -289,7 +289,7 @@ static void process_exception(unsigned int n, exc_context_t *ctx)
 	process_dumpException(n, ctx);
 
 	if (thread->process == NULL) {
-		hal_cpuHalt();
+		threads_halt();
```
jmaksymowicz: In cases like this in production it might be better to reboot than HCF - perhaps replace this check with `LIB_ASSERT_ALWAYS`? The latter already chooses whether to spin forever or reboot based on `NDEBUG` flag.

proc/threads.c
```diff
@@ -248,6 +248,21 @@ static int threads_timeintr(unsigned int n, cpu_context_t *context, void *arg)
 static void proc_lockForceUnlock(lock_t *lock, int doYield);
 
 
+__attribute__((noreturn)) void threads_halt(void)
+{
+	spinlock_ctx_t sc;
+
+	/* steal the spinlock, so that remaining cores on SMP systems spin forever */
```
jmaksymowicz: I would change this comment to indicate this is an _attempt_ to stop other cores, but is not 100% effective:
```suggestion
	/* 
	 * Take the threads spinlock and not release it - this is an attempt to stop other cores in SMP from
	 * continuing to execute code (they should eventually hang trying to run the scheduler).
	 */
```

jmaksymowicz (commented):
In `lib_assertPanic` the loop with `hal_cpuHalt` should probably be replaced with `threads_halt` - otherwise it will not stop other cores in SMP as intended.

proc/process.c
```diff
@@ -288,8 +288,9 @@ static void process_exception(unsigned int n, exc_context_t *ctx)
 
 	process_dumpException(n, ctx);
 
+	LIB_ASSERT_ALWAYS(thread->process != NULL, "exception in kernel")
```
julianuziemblo, line 291: As the name suggests, the latter check is now redundant as this asserts ALWAYS :smile:

test/proc.c
```diff
@@ -69,9 +70,7 @@ static void test_proc_indthr(void *arg)
 
 static void test_proc_busythr(void *arg)
 {
-	for (;;) {
-		hal_cpuHalt();
-	}
+	threads_halt();
```
jmaksymowicz: Don't replace this with `threads_halt` - this is not meant to freeze the whole CPU but rather to test if thread pre-emption is working correctly.

vm/map.c
```diff
@@ -828,8 +828,7 @@ static void map_pageFault(unsigned int n, exc_context_t *ctx)
 		process_dumpException(n, ctx);
 
 		if (thread->process == NULL) {
-			hal_cpuDisableInterrupts();
-			hal_cpuHalt();
+			threads_halt();
```
jmaksymowicz: This also looks like a candidate for `LIB_ASSERT_ALWAYS`, see my previous comment.

lib/assert.c
```diff
@@ -19,6 +19,7 @@
 #include "printf.h"
 #include "log/log.h"
 #include "hal/hal.h"
+#include "proc/threads.h"
 
 
 void lib_assertPanic(const char *func, int line, const char *fmt, ...)
```
ziemleszcz: Add `noreturn` attribute.

test/proc.c
```diff
@@ -18,6 +18,7 @@
 
 #include "hal/hal.h"
 #include "proc/proc.h"
+#include "proc/threads.h"
```
ziemleszcz: Not required?

proc/threads.c
ziemleszcz: Please use @jmaksymowicz's suggested comment. It explains the situation better.

jmaksymowicz (approved):

PR #815, opened 2026-07-22
!threads: implement robust and priority ceiling locks

adamgreloch:
TASK: RTOS-1399

## Types of changes

## How Has This Been Tested?

- [x] New test added: lukkrusz/rebased_posix
- [X] Tested by hand on: ia32-generic-qemu

## Checklist:

- [X] My change requires a change to the documentation.

## Special treatment


Makefile
```diff
@@ -31,6 +31,11 @@ CPPFLAGS += -DVERSION=\"$(VERSION)\" -DRELEASE=\"$(RELEASE)\" -DTARGET_FAMILY=\"
 # uncomment to enable stack canary checking
 # CPPFLAGS += -DSTACK_CANARY
 
+ifeq ($(DEBUG_HARDER),1)
```
ziemleszcz: Consider changing `DEBUG` env variable semantics from a boolean to a debug level (0 = off, 1 = default, 2 = verbose). Add this to `phoenix-rtos-build/Makefile.common`:

CPPFLAGS += -DDEBUG_LEVEL=$(DEBUG)

and this to `assert.h`:

#ifndef DEBUG_LEVEL
#define DEBUG_LEVEL 0
#endif

#define DEBUG_DEFAULT 1
#define DEBUG_VERBOSE  2

Use `DEBUG_LEVEL` in to turn off/on specific asserts.

proc/threads.c
ziemleszcz: Check for `EOWNERDEAD` (the same as L1503)?

include/errno.h
```diff
@@ -72,5 +72,6 @@
 #define ECONNREFUSED    111
 #define EALREADY        114
 #define EINPROGRESS     115
+#define EOWNERDEAD      130
```
ziemleszcz: ENOTRECOVERABLE is not implemented.

proc/threads.c
```diff
@@ -1498,6 +1500,12 @@ static int _proc_lockTry(thread_t *current, lock_t *lock)
 
 	lock->owner = current;
 	lock->depth = 1;
+	if (lock->inconsistent != 0 && lock->attr.robust != 0) {
+		return -EOWNERDEAD;
+	}
+	if (lock->attr.protocol == PH_LOCK_PROTO_PRIOCEILING) {
+		_proc_threadSetPriority(current, _proc_threadGetPriority(current));
```
ziemleszcz: According to POSIX it should return EINVAL if  `The mutex was created with the protocol attribute having the value PTHREAD_PRIO_PROTECT and the calling thread's priority is higher than the mutex's current priority ceiling.`

ziemleszcz: The priority should be set in `EOWNERDEAD` case above as well.

Makefile
adamgreloch: This implies that we will want to order assertion levels monotonically by just one parameter (w.r.t. assertion inclusion). Not sure we want that. When discussing this with @Darchiv some time ago, it seemed better to have potentially disjoint sets of inclusions (e.g. per subsystem) that are independently toggleable, hence the new variable for "harder" assertions

ziemleszcz: Understood. Maybe the variable name isn't the best choice. `DEBUG_HARDER` suggests a higher debugging level, while it actually enables a specific class of assertions. Consider renaming it to `DEBUG_LOCKS` or `DEBUG_THREADS`.

proc/threads.c
```diff
@@ -1495,6 +1497,14 @@ int threads_sigsuspend(unsigned int mask)
 /* Assumes `lock->spinlock` and `threads_common.spinlock` are set. */
 static int _proc_lockTry(thread_t *current, lock_t *lock)
 {
+	if (lock->consistency == LOCK_NOT_RECOVERABLE) {
+		return -ENOTRECOVERABLE;
+	}
+
+	if (lock->attr.protocol == PH_LOCK_PROTO_PRIOCEILING && current->priorityBase > lock->attr.prioceiling) {
```
ziemleszcz: `current->priorityBase < lock->attr.prioceiling`

proc/threads.c
ziemleszcz: What about `lock->consistency == LOCK_NOT_RECOVERABLE`?

Makefile
adamgreloch: Maybe `DEBUG_PARANOID`? Currently, there is 1 vm assertion on this level and 2 on threads level, so I'm wary of adding new types in the current state.

proc/threads.c
```diff
@@ -1495,6 +1497,14 @@ int threads_sigsuspend(unsigned int mask)
 /* Assumes `lock->spinlock` and `threads_common.spinlock` are set. */
 static int _proc_lockTry(thread_t *current, lock_t *lock)
 {
+	if (lock->attr.robust != 0U && lock->consistency == LOCK_NOT_RECOVERABLE) {
+		return -ENOTRECOVERABLE;
+	}
+
+	if (lock->attr.protocol == PH_LOCK_PROTO_PRIOCEILING && current->priorityBase < lock->attr.prioceiling) {
+		return -EINVAL;
+	}
```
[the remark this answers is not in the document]
adamgreloch: Wouldn't make sense from the PCP standpoint. POSIX doesn't specify that this must be an active priority

Makefile
julianuziemblo: I suggest `DEBUG_PEDANTIC` or `ASSERT_PEDANTIC`

proc/threads.h
```diff
@@ -22,6 +22,8 @@
 #include "process.h"
 #include "lock.h"
 
+#define MAX_PRIO 7U
```
julianuziemblo: for us it's trivial, but maybe this is a good moment to add a comment here with a brief explanation that `MAX_PRIO` is the LEAST priority, and 0 is the MOST priority

proc/threads.c
```diff
@@ -657,7 +655,11 @@ static u8 _proc_lockGetPriority(lock_t *lock)
 	u8 priority = MAX_PRIO;
 	thread_t *thread = lock->queue;
 
-	if (thread != NULL) {
+	if (lock->attr.protocol == PH_LOCK_PROTO_PRIOCEILING) {
+		return lock->attr.prioceiling;
+	}
+
+	if (lock->attr.protocol == PH_LOCK_PROTO_INHERIT && thread != NULL) {
```
julianuziemblo, line 682: do if we're NO_INHERIT, we always have a priority of MAX_PRIO on a lock?

proc/threads.c
```diff
@@ -1843,6 +1885,73 @@ int proc_lockWait(thread_t **queue, lock_t *lock, time_t timeout)
...
+	if (err < 0) {
+		return err;
+	}
+
+	err = (int)lock->attr.prioceiling;
+	if (prioceiling >= 0) {
+		lock->attr.prioceiling = (unsigned char)prioceiling;
+	}
+	else {
+		err = -EPERM;
```
ziemleszcz: On -1 we are supposed to return `lock->attr.prioceiling`.

proc/threads.c
```diff
@@ -1843,6 +1885,73 @@ int proc_lockWait(thread_t **queue, lock_t *lock, time_t timeout)
...
+
+	if (prioceiling < -1 || prioceiling > (int)MAX_PRIO) {
+		return -EINVAL;
+	}
+
+	if (lock->attr.protocol != PH_LOCK_PROTO_PRIOCEILING) {
+		return -EINVAL;
+	}
+
+	err = proc_lockSet(lock);
```
ziemleszcz: According to POSIX this function:
> shall attempt to lock the mutex as if by a call to pthread_mutex_lock(),
> except that the process of locking the mutex need not adhere to the priority protect protocol

However proc_lockSet() is doing exactly that.

proc/threads.c
ziemleszcz: Calling `_proc_threadSetPriority()` at L1641 without `threads_common.spinlock`?

proc/threads.c
adamgreloch: `need not` so it is allowed, but not required

proc/threads.c
adamgreloch: lol, right, apologies

Makefile
adamgreloch: After reconsideration, I've changed this to `DEBUG_THREADS` and `DEBUG_VM`. I was wary of adding more specific assertion types prematurely, but the per-subsystem assertion type seems reasonable and more likely to be stable than `_PARANOID/_PEDANTIC`

proc/threads.h
```diff
@@ -22,6 +22,14 @@
 #include "process.h"
 #include "lock.h"
```
julianuziemblo, line 24: nitpick: import `lib/assert.h` for `LIB_ASSERT_ALWAYS`

Makefile
```diff
@@ -31,6 +31,7 @@ CPPFLAGS += -DVERSION=\"$(VERSION)\" -DRELEASE=\"$(RELEASE)\" -DTARGET_FAMILY=\"
 # uncomment to enable stack canary checking
 # CPPFLAGS += -DSTACK_CANARY
 
+
```
julianuziemblo: ?

vm/vm.h
```diff
@@ -24,6 +24,13 @@
 #include "amap.h"
```
julianuziemblo, line 27: nitpick: import `lib/assert.h` for `LIB_ASSERT_ALWAYS`

proc/threads.c
ziemleszcz: Ack. It might still be a good idea to add an explicit comment about this.

proc/threads.c
adamgreloch: Yes, in a sense that it will never influence the priority of a thread in L1755

proc/threads.h
adamgreloch: there's already lib/lib.h

julianuziemblo: huh yeah, it "exports" all of the lib headers, mb

proc/threads.c
```diff
@@ -1843,6 +1887,70 @@ int proc_lockWait(thread_t **queue, lock_t *lock, time_t timeout)
...
```
ziemleszcz: Can we get the current `prioceiling` without taking the lock (for example using atomic load)? POSIX doesn't say that `pthread_mutex_getprioceiling()` has to lock the mutex.

proc/threads.c
ziemleszcz: Not adhering to the lock protocol.

proc/threads.c
```diff
@@ -1843,6 +1887,70 @@ int proc_lockWait(thread_t **queue, lock_t *lock, time_t timeout)
...
+	if (err < 0) {
+		return err;
```
ziemleszcz: It could be a robust mutex which return `-EOWNERDEAD` (mutex taken).

proc/threads.c
ziemleszcz: `lock` could be a robust mutex:

pthread_cond_wait->condWait->phCondWait->proc_condWait->proc_lockWait

so it can return `-EOWNERDEAD` or `-ENOTRECOVERABLE`.

proc/threads.c
adamgreloch: Yes, this is exactly what happens here and it was intended -  pthread_mutex_setprioceiling may fail with -EOWNERDEAD and then shall keep the mutex acquired.

(Assuming that `prioceiling == -1`/pthread_mutex_getprioceiling won't lock it, as you've pointed out in another comment)

proc/threads.c
adamgreloch: You're right. POSIX even implies it must not lock it, based on possible errnos for getprioceiling. Used a spinlock, as atomic would be problematic - I would need to declare `prioceiling` as atomic in exported `struct lockAttr`

proc/threads.c
```diff
@@ -1863,6 +1948,79 @@ int proc_lockWait(thread_t **queue, lock_t *lock, time_t timeout)
...
+			hal_spinlockClear(&lock->spinlock, &sc);
+			return err;
+		}
+
+		err = (int)lock->attr.prioceiling;
+		lock->attr.prioceiling = (unsigned char)prioceiling;
+		hal_spinlockClear(&lock->spinlock, &sc);
+
+		(void)proc_lockClear(lock);
+	}
```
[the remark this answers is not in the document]
adamgreloch: @gemini-code-assist unlocking the lock on -EOWNERDEAD would result in making the lock unrecoverable without giving caller the chance to make it consistent (unlocking an inconsistent lock makes it permanently unrecoverable)

proc/threads.c
```diff
@@ -1863,6 +1950,79 @@ int proc_lockWait(thread_t **queue, lock_t *lock, time_t timeout)
...
+
+	if (lock->attr.protocol != PH_LOCK_PROTO_PRIOCEILING) {
+		return -EINVAL;
+	}
+
+	if (prioceiling >= 0) {
+		hal_spinlockSet(&lock->spinlock, &sc);
+		err = _proc_lockSetRaw(lock, 0, &sc);
+		if (err < 0) {
```
ziemleszcz, line 2037: `EOWNERDEAD` is not handled (leaks mutex).

proc/threads.c
```diff
@@ -1566,11 +1611,72 @@ int proc_lockTry(lock_t *lock)
...
+	spinlock_ctx_t sc;
+	int ret = EOK, tid;
+
+	hal_spinlockSet(&threads_common.spinlock, &sc);
+
+	current = _proc_current();
+	tid = proc_getTid(current);
+
+	_trace_eventLockSetEnter(lock, tid);
+
```
ziemleszcz, line 1678: Deadlock check is missing so `proc_lockPrioCeiling()` can deadlock if it already own the mutex.

proc/threads.c
```diff
@@ -1685,26 +1759,30 @@ int proc_lockSetInterruptible(lock_t *lock)
...
 
+	if (lock->consistency == LOCK_INCONSISTENT) {
+		lock->consistency = LOCK_NOT_RECOVERABLE;
+	}
+	if (doForceUnlock == UNLOCK_FORCE && lock->consistency == LOCK_CONSISTENT) {
+		lock->consistency = LOCK_INCONSISTENT;
+	}
+
 	if (doForceUnlock == UNLOCK_TRY) {
 		if ((lock->attr.type == PH_LOCK_ERRORCHECK) || (lock->attr.type == PH_LOCK_RECURSIVE)) {
```
ziemleszcz: Also for robust mutex according to POSIX.

proc/threads.c
ziemleszcz, line 1713: POSIX says that `pthread_mutex_lock/trylock` should return `EPERM` when the mutex type is PTHREAD_MUTEX_ERRORCHECK or PTHREAD_MUTEX_RECURSIVE, or the mutex is a robust mutex, and the current thread does not own the mutex.
The is already added in `_proc_lockUnlock()` (except got robust mutex).

proc/threads.c
adamgreloch: https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/815#discussion_r3802394754

As per my response to the tin-man:
> @gemini-code-assist unlocking the lock on -EOWNERDEAD would result in making the lock unrecoverable without giving caller the chance to make it consistent (unlocking an inconsistent lock makes it permanently unrecoverable)

proc/threads.c
adamgreloch: I don't follow nor see anything about EPERM for lock/trylock in POSIX. Could you elaborate?

ziemleszcz: https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_mutex_lock.html

adamgreloch: > The pthread_mutex_unlock() function shall fail if: [...]

ziemleszcz: ?

proc/threads.c
```diff
@@ -1512,20 +1514,51 @@ int threads_sigsuspend(unsigned int mask)
...
+
+	return EOK;
+}
+
+
 /* Assumes `lock->spinlock` and `threads_common.spinlock` are set. */
-static int _proc_lockTry(thread_t *current, lock_t *lock)
+static int _proc_lockTryRaw(thread_t *current, lock_t *lock)
 {
+	if ((lock->attr.type == PH_LOCK_ERRORCHECK) && (lock->owner == current)) {
```
ziemleszcz: What about robust mutex and `pthread_mutex_setprioceiling()`? Should we have here:
```
if (((lock->attr.type == PH_LOCK_ERRORCHECK) || (lock->attr.robust != 0U)) && (lock->attr.type != PH_LOCK_RECURSIVE) && (lock->owner == current)) {
    return -EDEADLK;
}
```
Also looking at POSIX it seems that `pthread_mutex_trylock()` should return `EBUSY` instead of `EDEADLK` though I'm not sure about it. Could you check it on Linux?

proc/threads.c
```diff
@@ -1685,26 +1757,30 @@ int proc_lockSetInterruptible(lock_t *lock)
...
-	hal_spinlockSet(&threads_common.spinlock, &sc);
-
 	current = _proc_current();
 
 	_trace_eventLockClear(lock, proc_getTid(current));
 
 	LIB_ASSERT(LIST_BELONGS(&owner->locks, lock) != 0, "lock: %s, owner pid: %d, owner tid: %d, lock is not on the list",
 			lock->name, (owner->process != NULL) ? process_getPid(owner->process) : 0, proc_getTid(owner));
 
+	if (lock->consistency == LOCK_INCONSISTENT) {
```
ziemleszcz: If thread which gets `EOWNERDEAD` mutex dies before calling `pthread_mutex_consistent()` the mutex is marked as unrecoverable. According to POSIX:
`If the thread which acquired the mutex lock with the return value [EOWNERDEAD] terminates before calling either pthread_mutex_consistent() or [pthread_mutex_unlock()], the next thread that acquires the mutex lock shall be notified about the state of the mutex by the return value [EOWNERDEAD].`

proc/threads.c
adamgreloch: Ad 1: `EDEADLK` shall be returned only for ERRORCHECK. NORMAL locks (robust or not) shall deadlock on relock (see table in: https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_mutex_lock.html)

Ad 2: you're right, it returns `EBUSY` (checked on Linux).

proc/threads.c
adamgreloch: Done.

proc/threads.c
```diff
@@ -1685,26 +1765,36 @@ int proc_lockSetInterruptible(lock_t *lock)
...
 
 	LIB_ASSERT(LIST_BELONGS(&owner->locks, lock) != 0, "lock: %s, owner pid: %d, owner tid: %d, lock is not on the list",
 			lock->name, (owner->process != NULL) ? process_getPid(owner->process) : 0, proc_getTid(owner));
 
+	/*
+	 * Don't make the lock irrecoverable on force unlock (thread death).
+	 * POSIX says that if the owner of inconsistent lock dies, the
+	 * inconsistency should be passed to the next lock owner.
+	 */
+	if ((doForceUnlock != UNLOCK_FORCE) && (lock->consistency == LOCK_INCONSISTENT)) {
```
ziemleszcz: Thread not owning the inconsistent robust mutex can call unlock and set mutex unrecoverable (L1788). Maybe move `EPERM` block higher and set `consistency` only for robust mutex:
```
if (doForceUnlock == UNLOCK_TRY) {
    if ((lock->attr.type == PH_LOCK_ERRORCHECK) || (lock->attr.type == PH_LOCK_RECURSIVE) || (lock->attr.robust != 0U)) {
        if (lock->owner != current) {
            return -EPERM;
        }
    }
}
    
if (lock->attr.robust != 0U) {
        /*
         * Don't make the lock irrecoverable on force unlock (thread death).
         * POSIX says that if the owner of inconsistent lock dies, the
         * inconsistency should be passed to the next lock owner.
         */
        if ((doForceUnlock != UNLOCK_FORCE) && (lock->consistency == LOCK_INCONSISTENT)) {
                lock->consistency = LOCK_NOT_RECOVERABLE;
        }
        
        if ((doForceUnlock == UNLOCK_FORCE) && (lock->consistency == LOCK_CONSISTENT)) {
                lock->consistency = LOCK_INCONSISTENT;
        }
}

```

proc/mutex.c
```diff
@@ -56,6 +56,14 @@ int proc_mutexCreate(const struct lockAttr *attr)
...
 	}
 
+	if ((attr->protocol != PH_LOCK_PROTO_INHERIT) && (attr->protocol != PH_LOCK_PROTO_NOINHERIT) && (attr->protocol != PH_LOCK_PROTO_PRIOCEILING)) {
+		return -EINVAL;
+	}
+
+	if ((attr->protocol == PH_LOCK_PROTO_PRIOCEILING) && (attr->prioceiling > MAX_PRIO)) {
+		return -EINVAL;
+	}
+
```
ziemleszcz, line 70: Validate `attr->robust`?

include/threads.h
```diff
@@ -18,13 +18,20 @@
...
-#define PH_LOCK_NORMAL     0
-#define PH_LOCK_RECURSIVE  1
-#define PH_LOCK_ERRORCHECK 2
+#define PH_LOCK_NORMAL     0U
+#define PH_LOCK_RECURSIVE  1U
+#define PH_LOCK_ERRORCHECK 2U
+
+#define PH_LOCK_PROTO_INHERIT     0U
+#define PH_LOCK_PROTO_NOINHERIT   1U
+#define PH_LOCK_PROTO_PRIOCEILING 2U
```
ziemleszcz, line 27: Add `PH_LOCK_STALLED` and `PH_LOCK_ROBUST` and use them in the kernel and `pthread.h`?

proc/threads.c
```diff
@@ -1863,6 +1964,79 @@ int proc_lockWait(thread_t **queue, lock_t *lock, time_t timeout)
...
+		lock->consistency = LOCK_CONSISTENT;
+	}
+
+	hal_spinlockClear(&lock->spinlock, &sc);
+
+	return err;
+}
+
+
+/* prioceiling == -1 obtains the lock and retrieves current priority ceiling for it */
```
ziemleszcz: Only second part is true.

adamgreloch: Ah, a left-over. Sorry.

proc/mutex.c
```diff
@@ -56,6 +56,18 @@ int proc_mutexCreate(const struct lockAttr *attr)
...
 
+	if ((attr->robust != PH_LOCK_STALLED) && (attr->robust != PH_LOCK_ROBUST)) {
+		return -EINVAL;
+	}
+
+	if ((attr->protocol != PH_LOCK_PROTO_INHERIT) && (attr->protocol != PH_LOCK_PROTO_NOINHERIT) && (attr->protocol != PH_LOCK_PROTO_PRIOCEILING)) {
+		return -EINVAL;
+	}
+
+	if ((attr->protocol == PH_LOCK_PROTO_PRIOCEILING) && (attr->prioceiling > MAX_PRIO)) {
```
ziemleszcz: Check `prioceiling` unconditionally?

proc/threads.c
```diff
@@ -677,7 +675,11 @@ static u8 _proc_lockGetPriority(lock_t *lock)
 	u8 priority = MAX_PRIO;
 	thread_t *thread = lock->queue;
 
-	if (thread != NULL) {
+	if (lock->attr.protocol == PH_LOCK_PROTO_PRIOCEILING) {
+		return lock->attr.prioceiling;
```
ziemleszcz: This read is not protected the same way as write to `prioceiling`. It's correct only because `prioceiling` is one byte.

ziemleszcz (approved):

PR #817, opened 2026-07-27
vm/map: Entries merging bugfixes

### LINKED ISSUE #1714: kernel/vm: anon use-after-free - `_map_map()` reuses/extends amaps shared with fork COW twins (SMP)
## Description

On SMP MMU targets there is a race between a process that unmaps and re-maps anonymous
memory and the (asynchronous) teardown of its earlier-forked child. It leads to a live
anon (and its physical page) being freed while still mapped and referenced by the parent,
which manifests as user-visible memory corruption, kernel Data Aborts and hangs.

The root cause is in `vm/map.c:_map_map()`: when mapping anonymous memory it merges the
new region into an adjacent entry (`lmerge`/`rmerge`) or reuses the adjacent entry's amap
(`e->amap = amap_ref(next/prev->amap)`), and then clears the anon slots for the new region
(`amap_clear()`). None of these paths check whether the amap is **shared** (`refs > 1`).

After `fork()`, the child's map entries are copies of the parent's entries pointing at the
**same amap with the same `aoffs` ranges** (the COW twin design - each slot's anon holds one
reference per twin). If the parent then:

1. `munmap()`s such an entry — its anon references are dropped; the slots now belong
   exclusively to the child's twin entry (which may still exist: the child can be a zombie
   whose `vm_mapDestroy()` runs later on the kernel reaper thread, on another core),
2. `mmap()`s the same address range again — `_map_map()` happily extends the neighbouring
   entry / reuses its amap and computes the same slot range (amap offsets are linear in
   vaddr), then `amap_clear()` wipes the child-owned slots,
3. touches the new memory — a fresh anon (refs=1, owned by the parent) is installed in a
   slot that the child's twin entry still considers part of its own range,

then the child's delayed teardown (`vm_mapDestroy()` → `amap_putanons()` over the twin's
`aoffs` range) decrements the **parent's** brand-new anon to zero and frees it together
with its page, while the parent still has it mapped. The freed page gets reallocated
(e.g. as another process's COW copy or a zeroed anon page) and the parent reads garbage;
the dangling anon pointer left in the slot is later dereferenced (e.g. by `amap_getanons()`
during the parent's next `fork()`), corrupting kernel state.

The same class of bug exists in the entry-merge path of `vm_mprotect()`, which merges
adjacent entries checking only `protOrig`/`object`/`flags` - neither amap identity nor
`aoffs` contiguity - so a merged entry can claim amap slots it does not own.

## Symptoms

Fail example on CI:
(b507d2382e06a84b87d76e84a7f448d85c0d5055 in p-r-project and 507e3632e2783fd7b2ff8e07eede3d2ea6d3f11c in plo)
https://github.com/phoenix-rtos/plo/actions/runs/29244676420/job/86799263016

`test-mprotect` on `aarch64a53-zynqmp-qemu` (DEBUG=1) fails non-deterministically,
typically within tens of runs; three variants are observed:

- `pages_in_parent_copied` fails with the parent reading the child's COW value or zero
  (the parent's page was freed and reallocated under it):

```
TEST(test_mprotect, pages_in_child_copied) PASS
ASSERTION phoenix-rtos-tests/mem/test_mprotect.c:124:FAIL: Expected 66 Was 65   (also "Was 0")
TEST(test_mprotect, pages_in_parent_copied) FAIL at phoenix-rtos-tests/mem/test_mprotect.c:124
```

- kernel `Exception #37: Data Abort (EL1)` right after the test run (rb-tree / lock
  operations on corrupted VM structures, e.g. `pc` in `lib_listAdd` called from
  `_proc_threadEnqueue` with a lock queue containing garbage, or `x16` holding the
  `map_augment` callback), often in an endless dump loop,
- a hang (a thread blocks forever on a lock whose owner structure was freed).

The test sequence matters: `pages_in_child_copied` creates the fork twin sharing the amap;
its `munmap(area)` + the next test's `mmap(area)` recreate the aliased range while the
previous child is still being reaped.

## Reproduction
(b507d2382e06a84b87d76e84a7f448d85c0d5055 in p-r-project and 507e3632e2783fd7b2ff8e07eede3d2ea6d3f11c in plo)
```
DEBUG=1 TARGET=aarch64a53-zynqmp-qemu ./docker-build.sh clean core fs image project test
for i in $(seq 1 100); do echo "=== iteration $i ==="; ./docker-devel.sh ./phoenix-rtos-tests/runner.py -T aarch64a53-zynqmp-qemu -t phoenix-rtos-tests/mem/ -S -r mprotect || break; done
```

Faster reproduction (no reboot between runs — failures typically within 3–200 iterations):
boot the target once in QEMU and run `/bin/test-mprotect` in a loop on the psh prompt.

## Debugging evidence

Diagnostic instrumentation (lock magic canary + anon "black box" recording who created
and who freed each anon, with freed anons quarantined instead of `vm_kfree`d) pinpoints
the exact scenario - during the parent's `fork()` (`vm_mapCopy` → `amap_getanons`):

```
kernel (amap_getanon:85): getanon: dead anon ffffffffc23eb500 diePath=2 dieIdx=4
    dieAmap=ffffffffc22f7800 bornIdx=4 bornAmap=ffffffffc22f7800
    nowAmap=ffffffffc22f7800 nowIdx=4
```

i.e. the anon was created by the parent (`bornAmap`/`bornIdx`), freed by
`vm_mapDestroy()` of the zombie child (`diePath=2`) via **the same amap and slot**, and
the parent's live entry still referenced that slot at the next fork. Reproduced twice
with identical signatures.

## Initial fix

Attached patch (`vm-amap-alias-fix.patch`, `vm/amap.c`, `vm/amap.h`, `vm/map.c`):

[vm-amap-alias-fix.patch](https://github.com/user-attachments/files/30129842/vm-amap-alias-fix.patch)

- `amap_isShared()` helper (checks `refs > 1` under the amap lock),
- `_map_map()`: don't `lmerge`/`rmerge` into an entry whose amap is shared, and don't
  reuse a neighbour's shared amap for a new entry - a new exclusive amap is created
  lazily on first fault instead (purely an optimization loss, no functional change),
- `vm_mprotect()`: merge entries only when they are on the same amap with adjacent
  `aoffs` ranges.

Note: the `_map_map()` guard is conservative - it also skips merges when the amap is
shared only by entries of the same map (e.g. `prev`/`next` sharing one amap). This only
costs a merge optimization, never correctness.

etiaro:
## Description

There is an memory management bug that can cause Use after free of anons or leave mapped pages that system considers free (and hence e.g. corrupt arbitrary memory page in the future). These errors are rare, as require very specific fork and remap combination with both RO and RW memory. 
Seems to also be a source of sporadic mprotect CI fails.

Adding NEEDSCOPY flag to entries with unwriteable protOrig should not impact performance - the amap copy is done only if entry is writeable anyway. This way NEEDSCOPY flag protects us from reusing amaps that can be also held by forked processes.

## Motivation and Context

Resolves phoenix-rtos/phoenix-rtos-project#1714 - this PR uses different solution than suggested in the issue, as this way amap merge'ing policy is less strict while covering detected problems.

AI-assisted test cases created for triggering all fixed errors: [test_anon_race.c](https://github.com/user-attachments/files/30410944/test_anon_race.c)

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/vm/map.c
+++ b/vm/map.c
@@ -367,11 +367,11 @@ static void *_map_map(vm_map_t *map, void *vaddr, process_t *proc, size_t size, 
 
 		if (o == NULL) {
 			/* Try to use existing amap */
-			if (next != NULL && next->amap != NULL && e->vaddr >= (next->vaddr - next->aoffs)) {
+			if (next != NULL && next->amap != NULL && e->vaddr >= (next->vaddr - next->aoffs) && (next->flags & MAP_NEEDSCOPY) == 0U) {
 				e->amap = amap_ref(next->amap);
 				e->aoffs = next->aoffs - ((ptr_t)next->vaddr - (ptr_t)e->vaddr);
 			}
-			else if (prev != NULL && prev->amap != NULL && (SIZE_PAGE * prev->amap->size - prev->aoffs + (size_t)prev->vaddr) >= ((size_t)e->vaddr + size)) {
+			else if (prev != NULL && prev->amap != NULL && (SIZE_PAGE * prev->amap->size - prev->aoffs + (size_t)prev->vaddr) >= ((size_t)e->vaddr + size) && (prev->flags & MAP_NEEDSCOPY) == 0U) {
 				e->amap = amap_ref(prev->amap);
 				e->aoffs = prev->aoffs + ((ptr_t)e->vaddr - (ptr_t)prev->vaddr);
 			}
@@ -947,7 +947,7 @@ int vm_mprotect(vm_map_t *map, void *vaddr, size_t len, vm_prot_t prot)
 			attr = (vm_protToAttr(e->prot) | vm_flagsToAttr(e->flags));
 			needscopyNonLazy = 0;
 			/* If an entry needs copy, enter it as a readonly to copy it on first access. */
-			if ((e->flags & MAP_NEEDSCOPY) != 0U) {
+			if (((e->flags & MAP_NEEDSCOPY) != 0U) && ((prot & PROT_WRITE) != 0U)) {
 				if ((p == NULL) || (p->lazy == 0U)) {
 					needscopyNonLazy = 1;
 				}
@@ -1146,7 +1146,7 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
 		_vm_mapEntryCopy(f, e, 1);
 		(void)_map_add(proc, dst, f);
 
-		if (((e->protOrig & PROT_WRITE) != 0U) && ((e->flags & MAP_DEVICE) == 0U)) {
+		if ((e->flags & MAP_DEVICE) == 0U) {
 			e->flags |= MAP_NEEDSCOPY;
 			f->flags |= MAP_NEEDSCOPY;
 
--- a/vm/map.c
+++ b/vm/map.c
@@ -854,7 +854,7 @@ int vm_mprotect(vm_map_t *map, void *vaddr, size_t len, vm_prot_t prot)
 	addr_t pa;
 	vm_attr_t attr;
 	int needscopyNonLazy;
-	map_entry_t *e, *buf = NULL, *prev;
+	map_entry_t *e, *buf = NULL, *prev, *head, *tail;
 	map_entry_t t;
 
 	if (((((ptr_t)vaddr) & (SIZE_PAGE - 1U)) != 0U) || (len == 0U) || ((len & (SIZE_PAGE - 1U)) != 0U)) {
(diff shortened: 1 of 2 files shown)


vm/map.c
```diff
@@ -932,6 +932,7 @@ int vm_mprotect(vm_map_t *map, void *vaddr, size_t len, vm_prot_t prot)
 			if (lenLeft < currSize) {
 				_vm_mapEntrySplit(p, map, e, buf, lenLeft + ((ptr_t)currVaddr - (ptr_t)e->vaddr));
 				currSize = lenLeft;
+				buf = buf->next; /* Avoid freeing the split entry in final cleanup */
```
ziemleszcz: `buf->next` is overwritten by  `_vm_mapEntrySplit()`.

vm/map.c
```diff
@@ -969,9 +973,16 @@ int vm_mprotect(vm_map_t *map, void *vaddr, size_t len, vm_prot_t prot)
 				}
 			}
```
ziemleszcz, line 969: Missing `result` check. Error could be overwritten by 0 in next iteration.

etiaro: Nice catch

jmaksymowicz (dismissed):

vm/map.c
```diff
@@ -859,7 +859,7 @@ int vm_mprotect(vm_map_t *map, void *vaddr, size_t len, vm_prot_t prot)
 	addr_t pa;
 	vm_attr_t attr;
 	int needscopyNonLazy;
-	map_entry_t *e, *buf = NULL, *prev;
+	map_entry_t *e, *buf = NULL, *prev, *next;
```
ziemleszcz: `next` is misleading. Maybe add `head` and `tail` variables instead and use them at L921 and L933 instead of `prev` and `next`. Use `prev` only for previous iteration entry. This function is already too complex.

vm/map.c
ziemleszcz: I recommend returning early on `result != EOK` (or using goto). This function has too many levels of indentation.

vm/map.c
```diff
@@ -1144,7 +1159,7 @@ int vm_mapCopy(process_t *proc, vm_map_t *dst, vm_map_t *src)
 		_vm_mapEntryCopy(f, e, 1);
 		(void)_map_add(proc, dst, f);
 
-		if (((e->protOrig & PROT_WRITE) != 0U) && ((e->flags & MAP_DEVICE) == 0U)) {
+		if ((e->flags & MAP_DEVICE) == 0U) {
```
ziemleszcz, line 1159: `e->object != VM_OBJ_PHYSMEM`?

Note that map_entry created by `mmap(NULL, ..., MAP_ANONYMOUS | MAP_DEVICE, ...)` calll inherits MAP_DEVICE flag.

vm/map.c
etiaro: Probably a good idea, in the beginning I did not expect this PR to be a full mprotect rewrite.

vm/map.c
etiaro: Are you suggesting adding the `e->object != VM_OBJ_PHYSMEM` part to the condition or replacing that condition?
If I understand correctly, either of this option would cause sharing some writeable memory between parent and child when forking from syspage programs, as they use VM_OBJ_PHYSMEM as of now. ([CI has plenty of examples](https://github.com/phoenix-rtos/phoenix-rtos-kernel/actions/runs/30535351121?pr=817)).

I'm not sure what to do about the note – my understanding of MAP_DEVICE flag is that it points to the kernel that this is a memory for device communication and hence should be shared between parent and child.

vm/map.c
```diff
@@ -900,76 +902,89 @@ int vm_mprotect(vm_map_t *map, void *vaddr, size_t len, vm_prot_t prot)
 	if ((result == EOK) && (needed != 0U)) {
 		buf = map_allocN(needed);
 		if (buf == NULL) {
-			result = -ENOMEM;
+			(void)proc_lockClear(&map->lock);
+			return result;
```
ziemleszcz: No. Leave the old code here and add second `if` below:

```
if (result != EOK) {
    (void)proc_lockClear(&map->lock);
	return result;
}
```

or change your `if` to these two `ifs`:

```
if (result != EOK) {
    (void)proc_lockClear(&map->lock);
	return result;
}

if (needed != 0U) {
    buf = map_allocN(needed);
    if (buf == NULL) {
        (void)proc_lockClear(&map->lock);
	    return -ENOMEM;
    }
}
```

vm/map.c
ziemleszcz: OK, forget `e->object != VM_OBJ_PHYSMEM` check. That's also incorrect.

There are two issues with the new check (`(e->flags & MAP_DEVICE) == 0U`):

1. mmap with `MAP_ANONYMOUS | MAP_DEVICE`

Consider:

```
mmap(NULL, 4096, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_DEVICE, -1, 0);
```

mmap() does not mask out MAP_DEVICE, so it ends up in the map entry flags. I don't see anything in the kernel preventing a user from doing this.

After fork(), the parent and child will share the same amap, but the entries will not be marked with `MAP_NEEDSCOPY` because of `(e->flags & MAP_DEVICE) == 0U` check. This breaks copy-on-write, allowing the parent to observe the child's modifications.

2. Physical RAM mappings without MAP_DEVICE

Some drivers map DMA buffers or other physical memory without MAP_DEVICE (I'm not sure whether this is intentional and correct), for example:

```
mmap(NULL, ..., PROT_READ | PROT_WRITE, MAP_PHYSMEM | MAP_ANONYMOUS, -1, paddr);
```

Examples include ad7779, ade9113 (OCRAM in `dma/imx6ull-sdma/libsdma`), ttypc_vga.c, ttypc_bioskbd.c, and dummyfs.c.

If such a process forks and then writes to this mapping, `_map_force()` will create an amap, even though this is a physical mapping and should not have one. Later, `amap_page()` will allocate a new page, so the child will end up accessing private memory instead of the original DMA buffer.

This is currently a theoretical scenario, but maybe the kernel should handle it better.

etiaro: I'm not sure if I'm following correctly, I'd like to notice that `(e->flags & MAP_DEVICE) == 0U` is not a new check. It was here before. I only removed the check that avoided seting up NEEDSCOPY for READONLY segments. MAP_DEVICE entries did not get MAP_NEEDSCOPY flag before these changes either.
Please correct me if I'm wrong.

I am trying to follow the already established conventions in the codebase.
Maybe we should redesign how mappings work, and we definitely should write a documentation about it. But I don't think this should be the part of this PR, which aims mainly at removing actual UAF bugs.

My understanding of the current vm/ codebase is 1) it was the intentional behavior of MAP_DEVICE mappings, so that users can choose when a mapping is a mapping e.g. to device registers and remains as such after fork, and 2) someone decided in the past that MAP_PHYSMEM is too little to make mapping remain the same after fork().

Additionally, I think there is much more chance of breaking compatibility with changing that MAP_DEVICE/MAP_PHYSMEM behavior than with keeping it as it was.

ziemleszcz: The two issues I described are pre-existing.

I agree that issue 2 can be addressed in a separate PR, as it's not critical. I'm also not suggesting changing the semantics of MAP_DEVICE. 

However, I consider issue 1 to be critical because it can also lead to UAF. The kernel should either reject
```
mmap(NULL, ..., MAP_ANONYMOUS | MAP_DEVICE, -1, 0)
```
or handle it correctly. As I mentioned, nothing in the kernel currently prevents a user app from making this call.  Honestly its semantics are unclear to me. This looks like a case the kernel was never designed to handle. I don't think MAP_DEVICE was intended to be used this way.

The easiest and safest fix would be reject this in `syscall_sys_mmap()`.  I couldn't find any uses of this kind of mmap() in our codebase.

etiaro: Introduced EINVAL from mmap syscall when using MAP_DEVICE without MAP_PHYSMEM, as indeed without that check errors from the example tests in description could still be triggered with MAP_DEVICE added to all mmap calls.

vm/map.c
```diff
@@ -901,72 +906,89 @@ int vm_mprotect(vm_map_t *map, void *vaddr, size_t len, vm_prot_t prot)
...
+			map_augment(&prev->linkage);
+			e = prev;
+		}
+
+		e->prot = prot;
+
+		attr = (vm_protToAttr(e->prot) | vm_flagsToAttr(e->flags));
+		needscopyNonLazy = 0;
+		/* If an entry needs copy, enter it as a readonly to copy it on first access. */
+		if ((e->flags & MAP_NEEDSCOPY) != 0U) {
```
ziemleszcz: Notice that your new check at L1164 removed `(e->protOrig & PROT_WRITE) != 0U` condition. As a result, read-only maps now also receive `MAP_NEEDSCOPY` flag.

The side effect is that maps whose `protOrig` does not include `PROT_WRITE` (for example read-only `mmap()` regions, ELF `.text` and `.rodata` sections) will permanently keep `MAP_NEEDSCOPY` flag. It is cleared only when the map has `PROT_WRITE`.

On MMU targets, this means that after `fork()` calling `mprotect()` on such maps will run `_map_force()` and execute page-fault path for every page. Since `MAP_NEEDSCOPY` is never cleared, this is done on every `mprotect()` call. This is unnecessary and degrades `mprotect()` performance for read-only maps. There will be a noticeable performance difference between `mprotect()` on read-write and read-only regions.

I confirmed this by running `mprotect()` on QEMU (`ia32` and `riscv64`). On `ia32`, `mprotect()` on read-only memory takes about 1.5 as long as it does on read-write memory.

One way to fix this is to change this check to:

```
`((e->flags & MAP_NEEDSCOPY) != 0U) && ((prot & PROT_WRITE) != 0U)
```

But please check that this doesn't break anything. There may also be a better way to solve this.

etiaro: Seems like a good idea, I did not find any breaking points. This has much more performance impact than I expected, but that makes sesne for large, uniform regions.

ziemleszcz (approved):

PR #820, opened 2026-07-29
!include/errno: transfer all errno defines from libphoenix

julianuziemblo:
It seems that we will need all of those definitions in the kernel at some point, so transfer all of them at once now so that they are available for use later and no more breaking commits than necessary need to be made.

YT: RTOS-1408

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

    - https://github.com/phoenix-rtos/libphoenix/pull/510

### DIFF
--- a/include/errno.h
+++ b/include/errno.h
@@ -58,19 +58,47 @@
 #define EDEADLK      38
 #define EWOULDBLOCK  EAGAIN /* Operation would block */
 
+#define ENOTEMPTY 39
+#define ELOOP     40
+#define ENOMSG    42
+#define EUNATCH   49
+#define ENODATA   61
+#define ENONET    64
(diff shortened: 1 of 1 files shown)


include/errno.h
```diff
@@ -56,19 +58,47 @@
...
+#define ENOMSG    42
+#define EUNATCH   49
+#define ENODATA   61
+#define ENONET    64
+#define EBADFD    66
+#define EPROTO    71
+#define EBADMSG   74
+#define EOVERFLOW 75
+#define EILSEQ    84
+
```
ziemleszcz: Why are these blank lines needed? Why not keep all the error codes in a single block?

adamgreloch (approved):

PR #825, opened 2026-08-04
cond: handle conditional without mutex

lukileczo:
## Description

This commit introduces `PH_COND_UNLOCKED`, attribute which allows to use conditional variable without mutex. This is useful for utilizing the interrupt-signaled condition variables, where the condition does not require a mutex to be safely checked.

## Motivation and Context

Get rid of redundant locks in device drivers, which surround conditionals signaled through interrupts.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/509 https://github.com/phoenix-rtos/phoenix-rtos-doc/pull/265 https://github.com/phoenix-rtos/libphoenix/pull/513 https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/825

### DIFF
--- a/proc/cond.c
+++ b/proc/cond.c
@@ -54,7 +54,8 @@ int proc_condCreate(const struct condAttr *attr)
 	cond_t *cond;
 	int id;
 
-	if ((attr->clock != PH_CLOCK_RELATIVE) && (attr->clock != PH_CLOCK_REALTIME) && (attr->clock != PH_CLOCK_MONOTONIC)) {
+	if (((attr->clock != PH_CLOCK_RELATIVE) && (attr->clock != PH_CLOCK_REALTIME) && (attr->clock != PH_CLOCK_MONOTONIC)) ||
+			((attr->type != PH_COND_NORMAL) && (attr->type != PH_COND_UNLOCKED))) {
 		return -EINVAL;
 	}
 
@@ -85,18 +86,33 @@ int proc_condCreate(const struct condAttr *attr)
 int proc_condWait(int c, int m, time_t timeout)
 {
 	cond_t *cond;
-	mutex_t *mutex;
+	mutex_t *mutex = NULL;
 	int err = 0;
 	time_t offs, abstime = 0;
+	int condType;
 
 	cond = cond_get(c);
 	if (cond == NULL) {
 		return -EINVAL;
 	}
 
-	mutex = mutex_get(m);
-	if (mutex == NULL) {
-		cond_put(cond);
+	condType = cond->attr.type;
+
+	if (condType == PH_COND_NORMAL) {
+		mutex = mutex_get(m);
+		if (mutex == NULL) {
+			cond_put(cond);
+			return -EINVAL;
+		}
+	}
+	else if (condType == PH_COND_UNLOCKED) {
+		if (m != -1) {
+			cond_put(cond);
+			return -EINVAL;
+		}
+	}
+	else {
+		/* wrong type, should never happen */
 		return -EINVAL;
 	}
 
@@ -132,13 +148,28 @@ int proc_condWait(int c, int m, time_t timeout)
 				err = -EINVAL;
 				break;
 		}
+
+		if (err != 0) {
+			if (mutex != NULL) {
+				mutex_put(mutex);
+			}
+			cond_put(cond);
+			return err;
+		}
 	}
 
-	if (err == 0) {
+	if (condType == PH_COND_NORMAL) {
 		err = proc_lockWait(&cond->queue, &mutex->lock, abstime);
+
+		mutex_put(mutex);
+	}
+	else if (condType == PH_COND_UNLOCKED) {
+		err = proc_threadWaitExclusive(&cond->queue, abstime);
+	}
+	else {
+		/* already checked, no action required */
 	}
 
-	mutex_put(mutex);
 	cond_put(cond);
 
 	return err;
--- a/proc/cond.c
+++ b/proc/cond.c
@@ -67,15 +67,15 @@ int proc_condCreate(const struct condAttr *attr)
 	cond->resource.payload.cond = cond;
 	cond->resource.type = rtCond;
 
+	cond->queue = NULL;
+
+	hal_memcpy(&cond->attr, attr, sizeof(cond->attr));
+
 	id = resource_alloc(p, &cond->resource);
 	if (id < 0) {
 		vm_kfree(cond);
 		return -ENOMEM;
 	}
-
-	cond->queue = NULL;
-
-	hal_memcpy(&cond->attr, attr, sizeof(cond->attr));
 
 	(void)resource_put(p, &cond->resource);
 
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -33,6 +33,8 @@ enum { event_scheduling, event_enqueued, event_waking, event_preempted };
 #define UNLOCK_TRY        0
 #define UNLOCK_FORCE      1
 
+#define THREAD_WAIT_INTERRUPTIBLE (1U << 0)
+#define THREAD_WAIT_EXCLUSIVE     (1U << 1) /* reject a second waiter (unlocked exclusive conditional) */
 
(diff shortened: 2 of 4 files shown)


proc/threads.c
```diff
@@ -1080,6 +1080,29 @@ int proc_threadWaitInterruptible(thread_t **queue, spinlock_t *spinlock, time_t
 }
 
 
+int proc_threadWaitIntNoLock(thread_t **queue, time_t timeout)
```
agkaminski: Pls try to implement that using `proc_threadWaitEx` - the purpose of this `Ex` variant is to handle special cases

agkaminski: Perhaps handle spinlock = NULL?

lukileczo: Done

proc/threads.c
```diff
@@ -1080,6 +1085,12 @@ int proc_threadWaitInterruptible(thread_t **queue, spinlock_t *spinlock, time_t
...
```
agkaminski: Perhaps `proc_wait`? Not sure, I've came up with that by removing `lock` from `proc_lockWait`

proc/cond.c
```diff
@@ -132,13 +127,32 @@ int proc_condWait(int c, int m, time_t timeout)
...
 		}
+
+		if (err != 0) {
+			cond_put(cond);
+			return err;
+		}
 	}
 
-	if (err == 0) {
+	if (cond->attr.type == PH_COND_UNLOCKED) {
```
xvuko: Consider requiring fixed invalid mutex handle value to be used with PH_COND_UNLOCKED cond:
```
if (cond->attr.type == PH_COND_UNLOCKED) {
  if (m != 0) {
    cond_put(cond);
    return -EINVAL;
  }
```

proc/cond.c
```diff
@@ -94,12 +95,6 @@ int proc_condWait(int c, int m, time_t timeout)
 		return -EINVAL;
 	}
 
-	mutex = mutex_get(m);
```
ziemleszcz, line 97: Now it can return `-ETIME` instead of `-EINVAL` for an invalid mutex. I suggest keeping the early mutex check here for `cond->attr.type == PH_COND_NORMAL`. The same applies to unsupported `cond->attr.type` -  check it early, before processing `abstime`.

ziemleszcz, line 97: Verify that locked/unlocked variant of this call matches the cond type, and return an error if it doesn't.

Currently, an error is returned only for the `m==-1` and locked cond combination because `mutex_get(m)` returns NULL. It's better to explicitly reject incorrect API usage.

Maybe add this check before L138 or as part of the early mutex check I suggested in the comment above.

````
if (m >= 0) {
    cond_put(cond);
    return -EINVAL
}
```

proc/threads.c
ziemleszcz: I would change the API of this function to:
```
define THREAD_WAIT_INTERRUPTIBLE (1U << 0)
#define THREAD_WAIT_SINGLE (1U << 1)  /* reject a second waiter (unlocked conditional) */

int proc_threadWaitEx(thread_t **queue, spinlock_t *spinlock, time_t timeout, unsigned int flags, spinlock_ctx_t *scp);
```
I also recommend enforcing the single-waiter restriction for `proc_threadWaitIntNoLock()` case here: 
```
if (((flags & THREAD_WAIT_SINGLE) != 0U) && (*queue != NULL) && (*queue != wakeupPending)) {
    hal_spinlockClear(&threads_common.spinlock, &tsc);
    return -EBUSY;
}
```

proc/cond.c
lukileczo: Done, used `-1` as I did so in libphoenix

proc/cond.c
lukileczo: Done

lukileczo: Done

proc/threads.c
lukileczo: Done

proc/threads.c
```diff
@@ -1039,44 +1041,60 @@ int proc_threadNanoSleep(time_t *sec, long int *nsec, int absolute)
...
+	}
 
 	return err;
 }
 
 
 int proc_threadWait(thread_t **queue, spinlock_t *spinlock, time_t timeout, spinlock_ctx_t *scp)
 {
-	return proc_threadWaitEx(queue, spinlock, timeout, 0U, scp);
+	return proc_threadWaitEx(queue, spinlock, timeout, THREAD_WAIT_INTERRUPTIBLE, scp);
```
ziemleszcz: THREAD_WAIT_INTERRUPTIBLE -> 0U

proc/threads.c
```diff
@@ -1039,44 +1041,60 @@ int proc_threadNanoSleep(time_t *sec, long int *nsec, int absolute)
...
+	}
+
+	if (((flags & THREAD_WAIT_INTERRUPTIBLE) != 0U) && (_proc_current()->exit != 0U)) {
 		/* Waiting in this state can lead to becoming a hanging zombie */
 		hal_spinlockClear(&threads_common.spinlock, &tsc);
 		return -EINTR;
 	}
 
-	_proc_threadEnqueue(queue, timeout, interruptible);
+	_proc_threadEnqueue(queue, timeout, flags & THREAD_WAIT_INTERRUPTIBLE);
```
ziemleszcz: Works only because `THREAD_WAIT_INTERRUPTIBLE` is 1.

proc/threads.c
lukileczo: After further discussion and analysis, I don't think the single waiter flag should be added in this place. It would probably be more correct for `_proc_threadEnqueue` to handle it. But this function doesn't have a return value currently; adding it would require mostly useless return value checks - as I don't think we have any more uses for a single waiter queue apart from the highly specific use case of interrupt handling.

ziemleszcz: What's the problem with this solution? I'd prefer to return an explicit error for incorrect API usage.

lukileczo: Ok, I couldn't really come up with anything better, just renamed it and the new proc_threadWait function to use `EXCLUSIVE` suffix.

proc/threads.c
lukileczo: Done

proc/threads.c
lukileczo: Renamed to `proc_threadWaitExclusive`

proc/cond.c
```diff
@@ -132,13 +148,25 @@ int proc_condWait(int c, int m, time_t timeout)
 				err = -EINVAL;
 				break;
 		}
+
+		if (err != 0) {
+			cond_put(cond);
+			return err;
```
ziemleszcz, line 157: Mutex leak (from L102).

proc/cond.c
ziemleszcz, line 74: Move it after L78 where cond is fully initialized (pre-existing issue)?

proc/cond.c
lukileczo: Done

proc/cond.c
lukileczo: Done

proc/threads.c
```diff
@@ -1039,30 +1041,40 @@ int proc_threadNanoSleep(time_t *sec, long int *nsec, int absolute)
 }
 
 
-static int proc_threadWaitEx(thread_t **queue, spinlock_t *spinlock, time_t timeout, u8 interruptible, spinlock_ctx_t *scp)
+static int proc_threadWaitEx(thread_t **queue, spinlock_t *spinlock, time_t timeout, u32 flags, spinlock_ctx_t *scp)
 {
 	int err;
 	spinlock_ctx_t tsc;
+	spinlock_ctx_t *rescheduleScp = (scp == NULL) ? &tsc : scp;
```
ziemleszcz, line 1049: Suggestion (last one for this PR):
```
LIB_ASSERT((spinlock == NULL) == (scp == NULL), "spinlock and scp must both be NULL or both be non-NULL");
```

lukileczo: Done

ziemleszcz (approved):

PR #827, opened 2026-08-05
(lib, posix): small fixes

julianuziemblo:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/lib/cbuffer.h
+++ b/lib/cbuffer.h
@@ -14,7 +14,12 @@
 #ifndef _PH_LIB_CBUFFER_H_
 #define _PH_LIB_CBUFFER_H_
 
+#include "hal/types.h"
+
 #include "attrs.h"
+#include "assert.h"
+#include "utils.h"
+
 
 typedef struct {
 	size_t sz, r, w;
--- a/lib/cbuffer.h
+++ b/lib/cbuffer.h
@@ -28,7 +28,7 @@ typedef struct {
 } cbuffer_t;
 
 
-static inline size_t _cbuffer_free(cbuffer_t *buf)
+static inline size_t _cbuffer_free(const cbuffer_t *buf)
 {
 	if (buf->w == buf->r) {
 		return (buf->full != 0U) ? 0U : buf->sz;
@@ -37,7 +37,7 @@ static inline size_t _cbuffer_free(cbuffer_t *buf)
 }
 
 
-static inline size_t _cbuffer_avail(cbuffer_t *buf)
+static inline size_t _cbuffer_avail(const cbuffer_t *buf)
 {
 	return buf->sz - _cbuffer_free(buf);
 }
--- a/lib/idtree.h
+++ b/lib/idtree.h
@@ -17,7 +17,6 @@
 #include "hal/types.h"
 #include "rb.h"
 
-
 #define MAX_ID ((1ULL << ((size_t)__CHAR_BIT__ * (sizeof(int)) - 1U)) - 1ULL)
 
 
@@ -29,14 +28,6 @@ typedef struct {
 	int lmaxgap, rmaxgap;
 	int id;
 } idnode_t;
-
-/* parasoft-begin-suppress MISRAC2012-RULE_20_7-a 'type' within (type *) can not be put in the parentheses due to compilation error */
-#define lib_idtreeof(type, node_field, node) ({ \
-	long _off = (long)&(((type *)0)->node_field); \
-	idnode_t *tmpnode = (node); \
-	(type *)((tmpnode == NULL) ? NULL : ((void *)tmpnode - _off)); \
-})
-/* parasoft-end-suppress MISRAC2012-RULE_20_7-a */
 
 
 idnode_t *lib_idtreeFind(idtree_t *tree, int id);
--- a/lib/idtree.h
+++ b/lib/idtree.h
@@ -14,6 +14,7 @@
 #ifndef _PH_IDTREE_H_
 #define _PH_IDTREE_H_
 
+#include "hal/types.h"
 #include "rb.h"
 
 
--- a/lib/lib.h
+++ b/lib/lib.h
@@ -15,53 +15,14 @@
 #ifndef _PH_LIB_LIB_H_
 #define _PH_LIB_LIB_H_
 
-#include "hal/hal.h"
-#include "printf.h"
+#include "assert.h"
 #include "bsearch.h"
+#include "cbuffer.h"
+#include "idtree.h"
+#include "list.h"
+#include "printf.h"
 #include "rand.h"
 #include "rb.h"
-#include "list.h"
-#include "assert.h"
 #include "strutil.h"
-#include "idtree.h"
-
-
-/* TODO: migrate to C11 stdatomic - check the performance impact of seq_cnt (mandatory by MISRA) */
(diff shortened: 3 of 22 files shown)


lib/lib.h
```diff
@@ -48,9 +48,6 @@
 })
 
 
-#include "cbuffer.h"
```
ziemleszcz, line 51: I think the intention was for this header to include all the other library headers, so including this single header outside lib is sufficient. Maybe just move it to the top.

lib/cbuffer.h
```diff
@@ -15,6 +15,7 @@
 #define _PH_LIB_CBUFFER_H_
 
 #include "attrs.h"
+#include "lib.h"
```
ziemleszcz: It might be better to include only the specific headers needed. I think only `assert.h` is required here (besides `attr.h`). See `idtree.h` for an example. Unfortunately, there isn't a consistent convention used across in this dir. My impression is that `lib.h` is intended for files outside lib directory and files within lib should include only the headers they actually need.

perf/buffer-mem.c
```diff
@@ -14,6 +14,7 @@
 
 #include "include/perf.h"
 #include "include/errno.h"
+#include "lib/cbuffer.h"
```
ziemleszcz: Or just `lib/lib.h` like in every other file outside lib.

lib/cbuffer.h
julianuziemblo: `lib.h` is required here for `min` macro

lib/rb.h
```diff
@@ -14,12 +14,10 @@
...
-/* parasoft-begin-suppress MISRAC2012-RULE_20_7-a 'type' within *(type *) can not be put in the parentheses due to compilation error */
-#define lib_treeof(type, node_field, node) ({ \
-	long _off = (long)&(((type *)0)->node_field); \
-	rbnode_t *tmpnode = (node); \
-	(type *)((tmpnode == NULL) ? NULL : ((void *)tmpnode - _off)); \
-})
+#include "utils.h"
+
+/* parasoft-suppress-next-line MISRAC2012-RULE_20_7-a 'type' within *(type *) can not be put in the parentheses due to compilation error */
+#define lib_treeof(type, node_field, node) (((node) == NULL) ? NULL : ((type *)((ptr_t)(node) - offsetof(type, node_field))))
```
ziemleszcz: Be careful with changing macros! `node` can be a function call, and now we have two calls per macro instead of just one. Revert to the previous version.

posix/posix.c
```diff
@@ -2269,6 +2269,10 @@ int posix_shutdown(int socket, int how)
 	open_file_t *f;
 	int err;
 
+	if (how != SHUT_RD && how != SHUT_WR && how != SHUT_RDWR) {
```
ziemleszcz: I would move this check into `unix_shutdown()`. Since `lwip_shutdown()` already returns EINVAL, there's no need to duplicate it here. Also the invalid file descriptor check takes precedence (please verify it on Linux).

lib/rb.h
```diff
@@ -14,12 +14,14 @@
 #ifndef _PH_LIB_RB_H_
 #define _PH_LIB_RB_H_
 
+#include "utils.h"
```
ziemleszcz, line 18: "hal/types.h" for ptr_t?

lib/idtree.h
```diff
@@ -16,7 +16,6 @@
 
 #include "rb.h"
```
ziemleszcz, line 18: "hal/types.h" for size_t?

lib/utils.h
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright 2012, 2014, 2016 Phoenix Systems
+ * Copyright 2001, 2006 Pawel Pisarczyk
+ * Author: Pawel Pisarczyk, Pawel Kolodziej
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#ifndef _PH_LIB_UTILS_H_
+#define _PH_LIB_UTILS_H_
+
```
ziemleszcz, line 17: Include for SIZE_PAGE (not sure which)?

lib/list.h
```diff
@@ -14,7 +14,7 @@
 #define _PH_LIB_LIST_H_
 
 #include "hal/hal.h"
-#include "lib/lib.h"
+#include "utils.h"
```
ziemleszcz, line 18: "assert.h" for LIB_STATIC_ASSERT_SAME_TYPE?

posix/posix.c
julianuziemblo: I can, but this way we don't have to send msg to lwip to check, because we know these are invalid. I will verify the prescedence and adjust accordingly though

posix/unix.c
```diff
@@ -1130,15 +1133,20 @@ ssize_t unix_sendmsg(unsigned int socket, const struct msghdr *msg, unsigned int
...
 	unixsock_t *s;
+	int err = EOK;
 
 	s = unixsock_get(socket);
 	if (s == NULL) {
 		return -ENOTSOCK;
 	}
 
+	if (how != SHUT_RD && how != SHUT_WR && how != SHUT_RDWR) {
+		err = -EINVAL;
```
ziemleszcz: We don't want to actually shutdown socket in this case - one put only comrade.

julianuziemblo: Huh yeah, sorry, thought that one was just a normal put because of the TODO comment... my bad

posix/unix.c
```diff
@@ -1130,15 +1133,22 @@ ssize_t unix_sendmsg(unsigned int socket, const struct msghdr *msg, unsigned int
...
 {
 	unixsock_t *s;
+	int err = EOK;
 
 	s = unixsock_get(socket);
 	if (s == NULL) {
 		return -ENOTSOCK;
 	}
 
+	if (how != SHUT_RD && how != SHUT_WR && how != SHUT_RDWR) {
```
ziemleszcz, line 1142: `err` is not needed:
```
unixsock_put(s);

if (how != SHUT_RD && how != SHUT_WR && how != SHUT_RDWR) {
  return -EINVAL;
}

unixsock_put(s);
return EOK;
```

lib/utils.h
```diff
@@ -0,0 +1,48 @@
...
+#include "include/page.h"
```
ziemleszcz, line 18: `page.h` defines `_PAGE_SIZE` not `SIZE_PAGE`. Maybe use this one in `round_page()` macro? Or change include `hal/cpu.h` (I think it has `SIZE_PAGE`)?

posix/unix.c
julianuziemblo: I'd argue the intent is not so clear here, but that would be marginal only, so I'll change it if You think this is better?

lib/utils.h
julianuziemblo: Will use _SIZE_PAGE here

posix/unix.c
ziemleszcz: The intent is not clear in both version. If you want to make it clear:
```
if (how != SHUT_RD && how != SHUT_WR && how != SHUT_RDWR) {
  unixsock_put(s);
  return -EINVAL;
}

unixsock_put(s);
unixsock_put(s);
return EOK;
```
I leave it up to you.

ziemleszcz (approved):

PR #829, opened 2026-08-06
hal/armv7a: fix kernel space pages sometimes being marked not global

jmaksymowicz:
Non-global entries in kernel space may fail to be evicted, because they are added to TLB with a different ASID than the one that's used to try to evict them. This fixes the problem by ensuring that entries in kernel space are always marked as global, while ones in user space always as non-global.

## Description
In our OS, during loading of programs sometimes pages are added into the kernel address space that have access privileges for PL0 (user programs) - see https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/52be634975d2d43409aba8335275d4bc839ad220/vm/object.c#L187 
Whether that's entirely intentional or not is another matter, but the big problem with that on armv7a was the fact that they were marked with "not global" attribute.

This attribute made it so these entries were added into the TLB tagged with the ASID of the currently running process. This creates a problem, because while removing kernel mappings we invalidate TLB entries with `hal_cpuInvalVAAll`. This function invokes the `TLBIMVA` / `TLBIMVAIS`[1] operation that clears entries matching a virtual address and ASID. For global entries the ASID field of this instruction is ignored and it will clear them on VA match. For non-global entries both VA and ASID must match. However, because entries in the TLB were marked with ASID of the running process, while the instruction clearing them had the ASID of the kernel, it may have failed to clear these entries correctly.

This manifested as frequent crashes on armv7a9-zynq7000-zedboard. A big breakthrough in debugging the problem came when @adamgreloch created a method for calculating CRC32 of the ELF file after it was loaded, which led us to the realization that the contents in memory were being corrupted. Subsequently, we determined that the memory corruption was consistent with the page that contained ELF file contents being overwritten by `vm_kmalloc` and then used for writing `msg_t` structures. Further debugging helped us determine that there is likely no bug in hardware-independent parts of the software, as this "address reuse" never happened with virtual addresses, only with physical page addresses when using hardware-based translation.

I'm still not entirely sure why the bug occurs only on SMP targets. I suspect that's because (as a quick-and-dirty workaround to ASID allocation) the entire TLB of a core is cleared on every address space switch - but only on the currently running core. Thus, the other core may still see the stale TLB entries.

This is likely the fix for the issues as it explains all of the symptoms observed:
1. It only occurs on armv7a - because on other platforms handling of the "global" attribute is different
2. It only occurs on targets with SMP - see above
3. It only occurs on real hardware, not QEMU - QEMU doesn't emulate the TLB accurately to achieve better speed
4. Memory corruption consistent with `vm_kmalloc` reusing pages - happened because of stale TLB entries pointing a virtual address that was allocated for a message into an old, already occupied page.

Also, it fixes a clear mistake in the code rather than hoping to fix some esoteric race condition.

[1] No, we cannot use "clear all ASID" variants `TLBIMVAA` / `TLBIMVAAIS` because they **both** are part of the Multiprocessing Extensions. While armv7a7-imx6ull happens to have them despite only running on a single core, we cannot, in general, count on this being the case.

## Motivation and Context
Fixes frequent system crashes while loading programs on armv7a9-zynq7000-zedboard.

YT: RTOS-1162

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a9-zynq7000-zedboard

## Checklist:

## Special treatment


adamgreloch:
This was crazy. Kudos for your discovery

hal/armv7a/pmap.c
```diff
@@ -268,11 +268,14 @@
...
 {
 	unsigned int pti = ID_PTABLE((ptr_t)va);
+	u32 hwAttr;
 
 	hal_cpuCleanDataCache((ptr_t)&ptable[pti], (ptr_t)&ptable[pti] + sizeof(ptr_t));
 	ptr_t oldEntry = ptable[pti];
 	if ((attr & PGHD_PRESENT) != 0U) {
-		ptable[pti] = (pa & ~0xfffU) | attrMap[attr & 0x1fU];
+		hwAttr = attrMap[attr & 0x1fU];
+		hwAttr |= ((ptr_t)va < VADDR_USR_MAX) ? TT2S_NOTGLOBAL : 0;
```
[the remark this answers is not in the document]
jmaksymowicz: Fixed

hal/armv7a/pmap.c
```diff
@@ -268,11 +268,14 @@ void pmap_switch(pmap_t *pmap)
...
```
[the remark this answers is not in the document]
jmaksymowicz: Fixed.

hal/armv7a/pmap.c
ziemleszcz: Unrelated but worth fixing here. It seems that PGHD_MASK (0x1f) is missing in `attrMap[SCRATCH_ATTRS]`. SCRATCH_ATTR = 36 so it's out if bound.

hal/armv7a/pmap.c
```diff
@@ -268,11 +268,14 @@ void pmap_switch(pmap_t *pmap)
 static void _pmap_writeEntry(ptr_t *ptable, void *va, addr_t pa, vm_attr_t attr, unsigned char asid)
 {
 	unsigned int pti = ID_PTABLE((ptr_t)va);
+	u32 hwAttr;
 
 	hal_cpuCleanDataCache((ptr_t)&ptable[pti], (ptr_t)&ptable[pti] + sizeof(ptr_t));
 	ptr_t oldEntry = ptable[pti];
 	if ((attr & PGHD_PRESENT) != 0U) {
-		ptable[pti] = (pa & ~0xfffU) | attrMap[attr & 0x1fU];
+		hwAttr = attrMap[attr & 0x1fU];
```
ziemleszcz: Nitpick: use PGHD_MASK?

hal/armv7a/pmap.c
jmaksymowicz: Good catch. Weird that it has gone unnoticed for so long with all the static analysis tools... 

In practice this resulted in scratch page being re-mapped even when not necessary, so it was not leading to incorrect execution, just lost performance.

hal/armv7a/pmap.c
jmaksymowicz: Done

ziemleszcz (approved):

damianloew:
It seems that this is a proper fix for https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1225 and https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1545. Tested on hardware, and typical test campaigns are passing consistently.

PR #830, opened 2026-08-12
syscalls: introduce rlimits

etiaro:
TASK: RTOS-1346

## Description

New syscall that implements POSIX getrlimit/setrlimit.
Starting with RLIMIT_NOFILE support.

POSIX does specify following limits:
- **RLIMIT_NOFILE** - Limit on number of open files. - **implemented in this PR** as max file descriptor number +1 (same as Linux does)
- RLIMIT_CPU - limit of CPU time per process - could be implemented with atomics in process_t
- RLIMIT_FSIZE - Limit on file size. - would have to be implemented inside fs servers?
- RLIMIT_AS - Limit on address space size. - could be implemented in the future, requires fast map accounting
- RLIMIT_DATA - Limit on data segment size. - as above
- RLIMIT_CORE - core file maximum size - coredumps currently unsupported in Phoenix
- RLIMIT_STACK - Limit on stack size. - probably not applicable to Phoenix

Systems tend to extend this list with their own specific limits.

I suggest `limit_target_t` syscall argument to allow for future extension of the abilities to change limiting scope to thread/partition/uid etc. 

I opt for tracking the limits inside the related subsystems, as in my opinion introducing some centralized mechanism for process limit tracking is not worth it due to the cost of additional centralized locks.

## Motivation and Context

POSIX API compliance, syscall interface for future partitioning limits.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/include/limits.h
+++ b/include/limits.h
@@ -17,4 +17,45 @@
 
 #define HOST_NAME_MAX 255U
 
+
+/* Limit on size of core image. */
+#define RLIMIT_CORE 0
+/* Limit on CPU time per process. */
+#define RLIMIT_CPU 1
+/* Limit on data segment size. */
+#define RLIMIT_DATA 2
+/* Limit on file size. */
+#define RLIMIT_FSIZE 3
+/* Limit on number of open files. */
+#define RLIMIT_NOFILE 4
+/* Limit on stack size. */
+#define RLIMIT_STACK 5
+/* Limit on address space size. */
+#define RLIMIT_AS 6
+
+
+#define RLIM_INFINITY ((rlim_t) - 1)
+
+/* We can represent all limits.  */
+#define RLIM_SAVED_MAX RLIM_INFINITY
+#define RLIM_SAVED_CUR RLIM_INFINITY
+
+
+typedef unsigned int rlim_t;
+
+
+struct rlimit {
+	rlim_t rlim_cur;
+	rlim_t rlim_max;
+};
+
+
+typedef struct {
+	/* To be extended with target kinds for partition, thread, uid etc. in the future */
+	enum { limit_target_process } kind;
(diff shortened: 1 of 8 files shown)


syscalls.c
```diff
@@ -1973,6 +1973,38 @@ int syscalls_sys_uname(u8 *ustack)
...
 
 
+int syscalls_rlimits(u8 *ustack)
+{
+	limit_target_t target;
+	int resource;
+	struct rlimit *oldLimit;
+	const struct rlimit *newLimit;
+
+	GETFROMSTACK(ustack, limit_target_t, target, 0U);
```
ziemleszcz, line 1983: I'm not sure passing a struct by value is supported. Pass a pointer to the struct instead, as with all other syscalls.

include/limits.h
```diff
@@ -17,4 +17,45 @@
...
+
+struct rlimit {
+	rlim_t rlim_cur;
+	rlim_t rlim_max;
+};
+
+
+typedef struct {
+	/* To be extended with target kinds for partition, thread, uid etc. in the future */
+	enum { limit_target_process } kind;
```
ziemleszcz, line 55: Add explicit type name:
```
typedef enum { limit_target_process = 0 } limit_target_kind_t;
```

include/limits.h
```diff
@@ -17,4 +17,45 @@
 
 #define HOST_NAME_MAX 255U
 
+
+/* Limit on size of core image. */
```
ziemleszcz, line 21: You should probably move these definitions to a separate `resources.h` header and include `phoenix/resources.h` only from `sys/resources.h` in libphoenix. Currently, everything ends up in `limits.h` in libphoenix.

PR #833, opened 2026-08-17
!proc: increase the number of priorities from 8 to 64

adamgreloch:
This change moves the priority range from [0,7] to [-32,31].

This change is breaking:
1. The priority had to be made signed.
2. The priority syscall has been renamed to sys_priority.
3. As a consequence of 1., the API of sys_priority and mutexPrioCeiling syscalls was modified to return current priority and previous priority ceiling, respectively, through an extra argument instead of the retval. Otherwise legal negative priority values would conflict with errors.

[-32,31] range instead of [0,63] was chosen, as the existing userspace code is concentrated in the high prio range (<4) with 4 usually being the default thread priority. Expanding the range to negative values is a way to relax that contention without the need to reconfigure/rescale the priorities in the userspace code.

If needed, the readyBitmask can be extended with an L2 bitmask to support more priorities.

If needed, NPRIOS can be set to lower even values if memory is a concern.

TASK: RTOS-1434

## Types of changes

## How Has This Been Tested?

- [X] New test added: kernel self-tests updated
- [x] Tested by hand on: ia32

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/815

### DIFF
--- a/include/threads.h
+++ b/include/threads.h
@@ -15,6 +15,10 @@
 #define _PH_THREADS_H_
 
 
+#define PH_GET_PRIO     0x7fffffff
+#define PH_PRIO_DEFAULT 4
+
+
 /* Mutex attributes */
 
 
@@ -34,7 +38,7 @@ struct lockAttr {
 	unsigned char type;
 	unsigned char protocol;
 	unsigned char robust;
-	unsigned char prioceiling;
+	signed char prioceiling;
 };
 
 
--- a/perf/trace-events.h
+++ b/perf/trace-events.h
@@ -205,13 +205,13 @@ static inline void trace_eventThreadMeta(u8 id, const thread_t *t)
 	struct {
 		u16 pid;
 		u16 tid;
-		u8 priority;
+		s8 priority;
 		char name[128];
 	} __attribute__((packed)) ev;
 
 	TRACE_META_BODY(id, ev, NULL, {
 		ev.tid = (u16)proc_getTid(t);
-		ev.priority = (u8)t->priority;
+		ev.priority = (s8)t->priority;
 
 		if (t->process != NULL) {
 			ev.pid = (u16)process_getPid(t->process);
@@ -288,16 +288,16 @@ static inline void trace_eventSchedExit(unsigned int cpuId)
 }
 
 
-static inline void trace_eventThreadPriority(int tid, u8 priority)
+static inline void trace_eventThreadPriority(int tid, priority_t priority)
 {
 	struct {
 		u16 tid;
-		u8 priority;
+		s8 priority;
 	} __attribute__((packed)) ev;
 
 	TRACE_EVENT_BODY(TRACE_EVENT_THREAD_PRIORITY, ev, NULL, {
 		ev.tid = (u16)tid;
-		ev.priority = priority;
+		ev.priority = (s8)priority;
 	});
 }
 
--- a/perf/trace.c
+++ b/perf/trace.c
@@ -170,12 +170,12 @@ static void _emitThreadsCb(void *arg, int i, threadinfo_t *tinfo)
 	struct {
 		u16 pid;
 		u16 tid;
-		u8 priority;
+		s8 priority;
 		char name[128];
 	} __attribute__((packed)) ev;
 
 	ev.tid = (u16)tinfo->tid;
-	ev.priority = (u8)tinfo->priority;
+	ev.priority = (s8)tinfo->priority;
 	ev.pid = (u16)tinfo->pid;
 
 	hal_memcpy(ev.name, tinfo->name, sizeof(tinfo->name));
--- a/proc/lock.h
+++ b/proc/lock.h
@@ -68,7 +68,7 @@ int proc_lockInit(lock_t *lock, const struct lockAttr *attr, const char *name);
 int proc_lockConsistent(lock_t *lock);
 
 
-int proc_lockPrioCeiling(lock_t *lock, int prioceiling);
+int proc_lockPrioCeiling(lock_t *lock, int prioceiling, int *res);
 
 
 int proc_lockDone(lock_t *lock);
--- a/proc/mutex.c
+++ b/proc/mutex.c
@@ -64,7 +64,7 @@ int proc_mutexCreate(const struct lockAttr *attr)
 		return -EINVAL;
 	}
 
-	if (attr->prioceiling > MAX_PRIO) {
+	if (((int)attr->prioceiling < MIN_PRIO) || ((int)attr->prioceiling > MAX_PRIO)) {
 		return -EINVAL;
 	}
 
@@ -144,7 +144,7 @@ int proc_mutexConsistent(int h)
 }
 
 
-int proc_mutexPrioCeiling(int h, int prioceiling)
+int proc_mutexPrioCeiling(int h, int prioceiling, int *res)
 {
 	mutex_t *mutex;
 	int err;
@@ -154,7 +154,7 @@ int proc_mutexPrioCeiling(int h, int prioceiling)
 		return -EINVAL;
 	}
 
-	err = proc_lockPrioCeiling(&mutex->lock, prioceiling);
+	err = proc_lockPrioCeiling(&mutex->lock, prioceiling, res);
 
 	mutex_put(mutex);
 
--- a/proc/mutex.h
+++ b/proc/mutex.h
@@ -39,7 +39,7 @@ int proc_mutexTry(int h);
 int proc_mutexConsistent(int h);
 
 
-int proc_mutexPrioCeiling(int h, int prioceiling);
+int proc_mutexPrioCeiling(int h, int prioceiling, int *res);
 
 
 int proc_mutexUnlock(int h);
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -603,8 +603,6 @@ int proc_threadCreate(process_t *process, startFn_t start, int *id, u8 priority,
 	t->execdata = NULL;
 	t->wait = NULL;
 	t->locks = NULL;
-	t->stick = 0;
-	t->utick = 0;
 	t->priorityBase = priority;
 	t->priority = priority;
 	t->cpuTime = 0;
--- a/proc/threads.c
+++ b/proc/threads.c
@@ -45,7 +45,14 @@ static struct {
 	vm_map_t *kmap;
 	spinlock_t spinlock;
 	lock_t lock;
-	thread_t *ready[MAX_PRIO + 1U];
+
+	thread_t *ready[NPRIOS];
+	/*
+	 * TODO: optimize and split readyBitmask into u32/u64 bitmask table depending
+	 * on the platform size. Currently, this is suboptimal for 32-bit platforms.
+	 */
+	u64 readyBitmask;
+
 	thread_t **current;
 	time_t utcoffs;
 
@@ -70,9 +77,7 @@ static struct {
 	time_t prev;
 } threads_common;
 
-
-_Static_assert(MAX_PRIO <= (u8)-1, "MAX_PRIO must fit into priority type");
-
+_Static_assert(NPRIOS <= sizeof(threads_common.readyBitmask) * 8U, "NPRIOS must fit into ready bitmask type");
 
 static thread_t *_proc_current(void);
 static void _proc_threadDequeue(thread_t *t);
@@ -198,6 +203,53 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
 	}
 
 	hal_timerSetWakeup((unsigned int)wakeup);
+}
+
+
+static void _readyAdd(thread_t *t)
+{
+	int sidx = (int)t->priority + (int)PRIO_OFFSET;
+	unsigned int idx = (unsigned int)sidx;
+
+	LIB_ASSERT_THREADS(sidx >= 0, "bad idx");
+	LIB_ASSERT_THREADS(LIST_BELONGS(&threads_common.ready[idx], t) == 0, "thread already on the ready list");
+
+	LIST_ADD(&threads_common.ready[idx], t);
+	threads_common.readyBitmask |= ((u64)1U << idx);
+}
+
+
+static void _readyRemove(thread_t *t)
+{
+	int sidx = (int)t->priority + (int)PRIO_OFFSET;
+	unsigned int idx = (unsigned int)sidx;
+
+	LIB_ASSERT_THREADS(sidx >= 0, "bad idx");
+	LIB_ASSERT_THREADS(LIST_BELONGS(&threads_common.ready[idx], t) != 0, "thread is not on the ready list");
+
+	LIST_REMOVE(&threads_common.ready[idx], t);
+	if (threads_common.ready[idx] == NULL) {
+		threads_common.readyBitmask &= ~((u64)1U << idx);
+	}
+}
+
+
+static unsigned int _readyMinPrioIdx(void)
+{
+	/* TODO: replace with __builtin_ctzll once GOT issues in sparc libgcc are resolved. */
+	u32 high, low;
+
+	if (threads_common.readyBitmask == 0ULL) {
+		return NPRIOS;
+	}
+
+	low = (u32)threads_common.readyBitmask;
+	if (low != 0U) {
+		return hal_cpuGetFirstBit(low);
+	}
+
+	high = (u32)(threads_common.readyBitmask >> 32);
+	return hal_cpuGetFirstBit(high) + 32U;
 }
 
 
@@ -379,8 +431,8 @@ static int _threads_checkSignal(thread_t *selected, process_t *proc, cpu_context
 /* parasoft-suppress-next-line MISRAC2012-RULE_8_4 "Function is used externally within assembler code" */
 int _threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
 {
-	thread_t *current, *selected;
-	unsigned int i;
+	thread_t *current, *selected = NULL;
+	unsigned int idx;
 	process_t *proc;
 	cpu_context_t *signalCtx, *selCtx;
 	unsigned int cpuId = hal_cpuGetID();
@@ -400,21 +452,17 @@ int _threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
 
 		/* Move thread to the end of queue */
 		if (current->state == READY) {
-			LIST_ADD(&threads_common.ready[current->priority], current);
+			_readyAdd(current);
 			_threads_preempted(current);
 		}
 	}
 
-	/* Get next thread */
-	i = 0;
-	while (i < sizeof(threads_common.ready) / sizeof(thread_t *)) {
-		selected = threads_common.ready[i];
-		if (selected == NULL) {
-			i++;
-			continue;
-		}
-
-		LIST_REMOVE(&threads_common.ready[i], selected);
+	/* Select next thread */
+	idx = _readyMinPrioIdx();
+	while (idx < NPRIOS) {
+		selected = threads_common.ready[idx];
+		LIB_ASSERT(selected != NULL, "empty queue marked as nonempty?");
+		_readyRemove(selected);
 
 		if (selected->exit == 0U) {
 			break;
@@ -427,6 +475,9 @@ int _threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
 		selected->state = GHOST;
 		LIST_ADD(&threads_common.ghosts, selected);
 		(void)_proc_threadWakeup(&threads_common.reaper);
+
+		idx = _readyMinPrioIdx();
+		selected = NULL;
 	}
 
 	LIB_ASSERT(selected != NULL, "no threads to schedule");
@@ -568,13 +619,13 @@ void threads_canaryInit(thread_t *t, void *ustack)
 }
 
 
-int proc_threadCreate(process_t *process, startFn_t start, int *id, u8 priority, size_t kstacksz, void *stack, size_t stacksz, unsigned int sigmask, void *arg)
+int proc_threadCreate(process_t *process, startFn_t start, int *id, priority_t priority, size_t kstacksz, void *stack, size_t stacksz, unsigned int sigmask, void *arg)
 {
 	thread_t *t;
 	spinlock_ctx_t sc;
 	int err;
 
-	if (priority >= sizeof(threads_common.ready) / sizeof(thread_t *)) {
+	if (priority < MIN_PRIO || priority > MAX_PRIO) {
 		return -EINVAL;
 	}
 
@@ -660,7 +711,7 @@ int proc_threadCreate(process_t *process, startFn_t start, int *id, u8 priority,
 	/* Insert thread to scheduler queue */
 
 	_threads_waking(t);
-	LIST_ADD(&threads_common.ready[priority], t);
+	_readyAdd(t);
 
 	hal_spinlockClear(&threads_common.spinlock, &sc);
 
@@ -668,9 +719,9 @@ int proc_threadCreate(process_t *process, startFn_t start, int *id, u8 priority,
 }
 
 
-static u8 _proc_lockGetPriority(lock_t *lock)
+static priority_t _proc_lockGetPriority(lock_t *lock)
 {
-	u8 priority = MAX_PRIO;
+	priority_t priority = MAX_PRIO;
 	thread_t *thread = lock->queue;
 
 	if (lock->attr.protocol == PH_LOCK_PROTO_PRIOCEILING) {
@@ -690,9 +741,9 @@ static u8 _proc_lockGetPriority(lock_t *lock)
 }
 
 
-static u8 _proc_threadGetLockPriority(thread_t *thread)
+static priority_t _proc_threadGetLockPriority(thread_t *thread)
 {
-	u8 ret, priority = MAX_PRIO;
+	priority_t ret, priority = MAX_PRIO;
 	lock_t *lock = thread->locks;
 
 	if (lock != NULL) {
@@ -709,16 +760,17 @@ static u8 _proc_threadGetLockPriority(thread_t *thread)
 }
 
 
-static u8 _proc_threadGetPriority(thread_t *thread)
+static priority_t _proc_threadGetPriority(thread_t *thread)
 {
-	u8 ret = _proc_threadGetLockPriority(thread);
+	priority_t ret = _proc_threadGetLockPriority(thread);
 	return (ret < thread->priorityBase) ? ret : thread->priorityBase;
 }
 
 
-static void _proc_threadSetPriority(thread_t *thread, u8 priority)
+static void _proc_threadSetPriority(thread_t *thread, priority_t priority)
 {
 	unsigned int i;
+	int onReadyList = 0;
 
 	/* Don't allow decreasing the priority below base level */
 	if (priority > thread->priorityBase) {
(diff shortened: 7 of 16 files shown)


proc/threads.c
```diff
@@ -45,7 +45,10 @@ static struct {
 	vm_map_t *kmap;
 	spinlock_t spinlock;
 	lock_t lock;
-	thread_t *ready[MAX_PRIO + 1U];
+
+	thread_t *ready[NPRIOS];
+	u64 readyNonempty;
```
etiaro: I would opt for name without negation, like readyBitmask/readyOccupied

proc/threads.c
```diff
@@ -201,6 +205,33 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
...
+	u8 prio = t->priority;
+	LIST_ADD(&threads_common.ready[prio], t);
+	threads_common.readyBitmask |= ((u64)1U << prio);
+}
+
+
+static void _readyRemove(thread_t *t)
+{
+	u8 prio = t->priority;
+	LIB_ASSERT(LIST_BELONGS(&threads_common.ready[prio], t) != 0,
```
julianuziemblo: maybe we can add similar centralized LIST_BELONGS to the `_readyAdd` function to check if the thread is NOT already in the ready queue somehow? or is this something not worth checking?

proc/threads.c
```diff
@@ -201,6 +205,33 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
...
+	LIST_REMOVE(&threads_common.ready[prio], t);
+	if (threads_common.ready[prio] == NULL) {
+		threads_common.readyBitmask &= ~((u64)1U << prio);
+	}
+}
+
+
+static u8 _readyMinPrio(void)
+{
+	return (threads_common.readyBitmask != 0ULL) ? (u8)__builtin_ctzll(threads_common.readyBitmask) : NPRIOS;
```
julianuziemblo: wouldn't we want to use `hal_cpuGetLastBit` here instead of the builtin? That's how it's done in the rest of the code, so I suppose it would be better even if we'd like to replace it in the future with the builtin.

proc/threads.c
```diff
@@ -201,6 +205,33 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
...
+			"thread: 0x%p, tid: %d, priority: %d, is not on the ready list",
+			t, proc_getTid(t), prio);
+	LIST_REMOVE(&threads_common.ready[prio], t);
+	if (threads_common.ready[prio] == NULL) {
+		threads_common.readyBitmask &= ~((u64)1U << prio);
+	}
+}
+
+
+static u8 _readyMinPrio(void)
```
julianuziemblo: I suppose the compiler will probably do it on its own, but this could be marked as `inline`

proc/threads.c
```diff
@@ -2274,15 +2305,16 @@ int _threads_init(vm_map_t *kmap, vm_object_t *kernel)
 		threads_common.stackCanary[i] = ((i & 1U) != 0U) ? 0xaaU : 0x55U;
 	}
 
-	/* Initiaizlie scheduler queue */
-	for (i = 0; i < sizeof(threads_common.ready) / sizeof(thread_t *); i++) {
+	/* Initialize scheduler queue */
+	threads_common.readyBitmask = 0;
+	for (i = 0; i < NPRIOS; i++) {
```
julianuziemblo: not better to `hal_memset` to `0`?

proc/threads.c
jmaksymowicz: `hal_cpuGetLastBit` takes `unsigned long` type and the size is platform-dependent, while `threads_common.readyBitmask` is always 64-bit

IMO we should get rid of those get first bit/get last bit functions entirely and just use builtins (esp. because ASM implementations prevent compiler optimizations for compile-time constants) but there's probably some big legacy reason why we shouldn't...

proc/threads.c
adamgreloch: `inline` in *.c files is essentially a no-op in modern compilers. I wouldn't bother about this

proc/threads.c
adamgreloch: yes, thanks

proc/threads.c
adamgreloch: Done

proc/threads.c
adamgreloch: Yeah, I'm also not a fan of these. Aside from degraded optimization opportunities, everyone knows how `__builtin_ctzll` and alike work, while hal_cpuGetLastBit et al. require time to figure out

ziemleszcz:
This PR depends on https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/815.

adamgreloch:
> This PR depends on #815.

Yes, I'm periodically rebasing it on top of `adamgreloch/RTOS-1399` branch (see the PR destination branch). Made this explicit in the description.

proc/threads.c
```diff
@@ -382,21 +416,18 @@ int _threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
...
-		if (selected == NULL) {
-			i++;
-			continue;
-		}
-
-		LIST_REMOVE(&threads_common.ready[i], selected);
+	/* Select next thread */
+	for (;;) {
+		prio = _readyMinPrio();
+		LIB_ASSERT(prio < NPRIOS, "no threads to schedule");
```
ziemleszcz: `_readyMinPrio()` literally returns NPRIOS in one branch. Old code was handling it gracefully. This versions takes out of bound element from `ready` queue.

proc/threads.c
```diff
@@ -382,21 +416,18 @@ int _threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
...
-	while (i < sizeof(threads_common.ready) / sizeof(thread_t *)) {
-		selected = threads_common.ready[i];
-		if (selected == NULL) {
-			i++;
-			continue;
-		}
-
-		LIST_REMOVE(&threads_common.ready[i], selected);
+	/* Select next thread */
+	for (;;) {
```
ziemleszcz: Unbounded loop. Old loop could do one pass only.

adamgreloch: Uh, yes, replaced with `while (prio < NPRIOS)`.

On a side note, not sure about the one pass thing - previously the loop would do multiple passes when there were ghosts in the queues, but maybe I misunderstood your statement

ziemleszcz: One pass over ready queue (i=0, 1, ..., 7).

adamgreloch: Ok, then I agree

proc/threads.h
```diff
@@ -57,11 +58,11 @@ typedef struct _thread_t {
 	struct _thread_t **wait;
 	time_t wakeup;
 
-	unsigned int priorityBase : 4;
-	unsigned int priority : 4;
+	unsigned int priorityBase : 8;
```
ziemleszcz: Use u8 for priorityBase and priority and u8 or u16 for bitfields?

proc/threads.c
```diff
@@ -409,6 +439,7 @@ int _threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
 		selected->state = GHOST;
 		LIST_ADD(&threads_common.ghosts, selected);
 		(void)_proc_threadWakeup(&threads_common.reaper);
+		prio = _readyMinPrio();
```
ziemleszcz: Add blank line above?

proc/threads.c
```diff
@@ -201,6 +205,36 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
...
+			t, proc_getTid(t), prio);
+	LIST_ADD(&threads_common.ready[prio], t);
+	threads_common.readyBitmask |= ((u64)1U << prio);
+}
+
+
+static void _readyRemove(thread_t *t)
+{
+	u8 prio = t->priority;
+	LIB_ASSERT(LIST_BELONGS(&threads_common.ready[prio], t) != 0,
```
ziemleszcz: I'm not sure iterating over the thread list in an assertion is a good idea.

ziemleszcz:
You should probably also update test_proc_threads1/2 to use all priorities.

proc/threads.c
adamgreloch: This assertion was moved _proc_threadSetPriority(). Moved this to `DEBUG_THREADS`, as it can be useful, but likely not in the general DEBUG build.

adamgreloch:
> You should probably also update test_proc_threads1/2 to use all priorities.

Done:

```
test: [proc.threads] Starting indicating thread
test: [proc.threads] P00 o / / - \ | | o  00 90 46 39 80 45 21 04
test: [proc.threads] P08 - | | / | o \ |  91 81 73 66 61 56 52 49
test: [proc.threads] P16 / - | - | - | o  46 43 41 39 37 35 33 32
test: [proc.threads] P24 - / \ - / / | o  31 30 28 27 26 26 25 24
test: [proc.threads] P32 - o o o o o o o  23 00 00 00 00 00 00 00
test: [proc.threads] P40 o o o o o o o o  00 00 00 00 00 00 00 00
test: [proc.threads] P48 o o o o o o o o  00 00 00 00 00 00 00 00
test: [proc.threads] P56 o o o o o o o o  00 00 00 00 00 00 00 00
```

proc/threads.c
```diff
@@ -409,6 +435,8 @@ int _threads_schedule(unsigned int n, cpu_context_t *context, void *arg)
 		selected->state = GHOST;
 		LIST_ADD(&threads_common.ghosts, selected);
 		(void)_proc_threadWakeup(&threads_common.reaper);
+
```
ziemleszcz, line 472: selected = NULL;

proc/threads.h
```diff
@@ -57,11 +58,11 @@ typedef struct _thread_t {
...
 	time_t wakeup;
 
-	unsigned int priorityBase : 4;
-	unsigned int priority : 4;
-	unsigned int state : 2;
-	unsigned int exit : 2;
-	unsigned interruptible : 1;
+	u8 priorityBase;
+	u8 priority;
+	u8 state : 2;
```
ziemleszcz: Unfortunately it seems that MISRA requires bitfields to use `int` or `unsigned int`.

proc/threads.h
```diff
@@ -28,7 +28,8 @@
 #define LIB_ASSERT_THREADS(condition, fmt, ...)
 #endif
 
-#define MAX_PRIO 7U /* Maximum priority value, of the lowest criticality (prio=0 is of the HIGHEST) */
+#define NPRIOS   64U
```
ziemleszcz: Please update scheduler.md in phoenix-rtos-doc.

ziemleszcz:
Please update the commit message (`If needed, the readyNonempty`).

adamgreloch:
> Please update the commit message (`If needed, the readyNonempty`).

Done.

ziemleszcz:
We should probably consider changing the default priority but it's a bigger task.

proc/threads.c
adamgreloch: Unfortunately, I had to use `hal_cpuGetFirstBit` due to [issues on sparc](https://github.com/phoenix-rtos/phoenix-rtos-kernel/actions/runs/33371300580/job/99423617185?pr=833#logs). It seems that the sparc linker has a GOT bug that causes a null deref in the implementation of `__builtin_ctzll`... I'm not deep enough in sparc to tackle the root cause now.

ziemleszcz: What about using `__builtin_ctz` instead? Note that 32-bit targets may not support 64-bits CTZ natively.

proc/threads.c
```diff
@@ -735,35 +773,39 @@ static void _proc_threadSetPriority(thread_t *thread, u8 priority)
...
+int proc_threadPriority(thread_t *t, int val)
 {
 	spinlock_ctx_t sc;
 	int ret, reschedule = 0;
-	u8 priority;
+	priority_t priority;
 
-	if ((signedPriority < -1) || (signedPriority > (int)MAX_PRIO)) {
+	if ((val != PH_GET_PRIO) && ((val < (int)MIN_PRIO) || (val > (int)MAX_PRIO))) {
 		return -EINVAL;
```
ziemleszcz, line 807: This function returns negative error code or negative priority value.

proc/threads.c
ziemleszcz: Negative `err` could be valid negative priority.

ziemleszcz: This function returns negative error code or negative priority value.

include/threads.h
```diff
@@ -34,7 +37,7 @@ struct lockAttr {
 	unsigned char type;
 	unsigned char protocol;
 	unsigned char robust;
-	unsigned char prioceiling;
+	char prioceiling;
```
ziemleszcz: `char` may be unsigned.

proc/threads.c
```diff
@@ -2040,7 +2082,7 @@ int proc_lockPrioCeiling(lock_t *lock, int prioceiling)
 		}
 		hal_spinlockClear(&lock->spinlock, &sc);
 
-		err = (int)__atomic_exchange_n(&lock->attr.prioceiling, (unsigned char)prioceiling, __ATOMIC_RELAXED);
+		err = (int)__atomic_exchange_n((unsigned char *)&lock->attr.prioceiling, (unsigned char)prioceiling, __ATOMIC_RELAXED);
```
ziemleszcz: `unsigned char`?

include/threads.h
adamgreloch: But the prioceiling can be negative

proc/threads.c
adamgreloch: MISRA/cppcheck complains that __atomic_exchange_n -> __atomic_exchange_1 takes unsigned char as an argument, hence the cast

include/threads.h
ziemleszcz: Excatly, so you should use signed type. In C `char`  signedness is implementation defined.

proc/threads.c
```diff
@@ -789,7 +835,9 @@ int proc_threadPriority(thread_t *t, int signedPriority)
 		}
 	}
 
-	ret = (int)t->priorityBase;
+	if (prevPriority != NULL) {
+		*prevPriority = (int)t->priorityBase;
```
ziemleszcz: Do not write to user space address under spinlock. Copy this value to a local variable and write to `prevPriority` after releasing the spinlock (see for example `proc_schedSet()`).

syscalls.c
```diff
@@ -369,13 +369,18 @@ int syscalls_nsleep(u8 *ustack)
...
 {
-	int priority;
+	int val, *prev;
 
-	GETFROMSTACK(ustack, int, priority, 0U);
+	GETFROMSTACK(ustack, int, val, 0U);
+	GETFROMSTACK(ustack, int *, prev, 1U);
 
-	return proc_threadPriority(proc_current(), priority);
+	if (prev != NULL && vm_mapBelongs(proc_current()->process, prev, sizeof(prev)) < 0) {
```
ziemleszcz: sizeof(prev) -> sizeof(*prev)

Please keep MISRA parenthesization.

proc/threads.h
```diff
@@ -103,10 +113,10 @@ thread_t *proc_current(void);
 void threads_canaryInit(thread_t *t, void *ustack);
 
 
-int proc_threadCreate(process_t *process, startFn_t start, int *id, u8 priority, size_t kstacksz, void *stack, size_t stacksz, unsigned int sigmask, void *arg);
+int proc_threadCreate(process_t *process, startFn_t start, int *id, priority_t priority, size_t kstacksz, void *stack, size_t stacksz, unsigned int sigmask, void *arg);
 
 
-int proc_threadPriority(thread_t *t, int signedPriority);
+int proc_threadPriority(thread_t *t, int val, int *prevPriority);
```
ziemleszcz: Unclear semantics - it returns the new priority in `prevPriority` when `val != PH_GET_PRIO`. Maybe split this into two separate functions?

syscalls.c
```diff
@@ -303,13 +303,13 @@ int syscalls_beginthreadex(u8 *ustack)
 {
 	process_t *proc = proc_current()->process;
 	startFn_t start;
-	unsigned int priority, stacksz; /* FIXME: stacksz should probably be size_t */
+	int priority, stacksz; /* FIXME: stacksz should probably be size_t */
```
ziemleszcz: unsigned int stacksz;

perf/trace-events.h
```diff
@@ -205,13 +205,13 @@ static inline void trace_eventThreadMeta(u8 id, const thread_t *t)
 	struct {
 		u16 pid;
 		u16 tid;
-		u8 priority;
+		s8 priority;
```
ziemleszcz, line 208: Also in `_emitThreadsCB()`.

proc/threads.c
```diff
@@ -201,6 +202,45 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
 }
 
 
+static void _readyAdd(thread_t *t)
+{
+	size_t idx = (size_t)t->priority + PRIO_RANGE;
```
ziemleszcz: This probably violates MISRA, prefer `(size_t)((int)t->priority + (int)PRIO_RANGE)`.

proc/threads.h
```diff
@@ -22,13 +22,25 @@
...
 #else
 #define LIB_ASSERT_THREADS(condition, fmt, ...)
 #endif
 
-#define MAX_PRIO 7U /* Maximum priority value, of the lowest criticality (prio=0 is of the HIGHEST) */
+#ifndef NPRIOS
+#define NPRIOS 64U
+#endif
+
+_Static_assert(NPRIOS % 2U == 0U, "NPRIOS should be even");
```
ziemleszcz, line 37: Enforce minimum limit on NPRIOS (so default priority 4 is in range, -7 to 8?).

proc/threads.h
adamgreloch: ah, here it should be a resulting priority. Notion of previous priority only concerns the mutexPrioCeiling.

proc/threads.c
adamgreloch: uh, not only that, it was incorrect. apologies

ziemleszcz:
Note that `priority` in `msg_t` is still `unsigned int`.

test/proc.c
```diff
@@ -110,20 +119,17 @@ static void test_proc_rotthr1(void *arg)
...
-	proc_threadCreate(NULL, test_proc_rotthr1, NULL, 6, stacksz, NULL, 0, 0, (void *)(int *)6);
-	proc_threadCreate(NULL, test_proc_rotthr1, NULL, 7, stacksz, NULL, 0, 0, (void *)(int *)7);
+	proc_threadCreate(NULL, test_proc_indthr, NULL, MIN_PRIO, stacksz, NULL, 0, 0, NULL);
 
-	proc_threadCreate(NULL, test_proc_busythr, NULL, 4, 1024, NULL, 0, 0, NULL);
+	for (i = 1; i < NPRIOS; i++) {
+		proc_threadCreate(NULL, test_proc_rotthr1, NULL, MIN_PRIO + i, stacksz, NULL, 0, 0, (void *)(ptr_t)i);
+	}
+
+	proc_threadCreate(NULL, test_proc_busythr, NULL, (MAX_PRIO - MIN_PRIO) / 2, 1024, NULL, 0, 0, NULL);
```
ziemleszcz, line 132: (MAX_PRIO - MIN_PRIO) / 2 = (31 - (-32)) / 2 = 63 / 2 = 31
Use 0 instead?

adamgreloch: eh, right

proc/threads.h
```diff
@@ -103,10 +114,11 @@ thread_t *proc_current(void);
 void threads_canaryInit(thread_t *t, void *ustack);
 
 
-int proc_threadCreate(process_t *process, startFn_t start, int *id, u8 priority, size_t kstacksz, void *stack, size_t stacksz, unsigned int sigmask, void *arg);
+int proc_threadCreate(process_t *process, startFn_t start, int *id, priority_t priority, size_t kstacksz, void *stack, size_t stacksz, unsigned int sigmask, void *arg);
 
 
-int proc_threadPriority(thread_t *t, int signedPriority);
+/* Sets the priority of t to val or retrieves the current priority of t when val == PH_GET_PRIORITY. Returns the current priority of t in *res if res != NULL. */
```
ziemleszcz: PH_GET_PRIORITY -> PH_GET_PRIO

perf/trace.c
```diff
@@ -170,12 +170,12 @@ static void _emitThreadsCb(void *arg, int i, threadinfo_t *tinfo)
...
 		u16 pid;
 		u16 tid;
-		u8 priority;
+		s8 priority;
 		char name[128];
 	} __attribute__((packed)) ev;
 
 	ev.tid = (u16)tinfo->tid;
-	ev.priority = (u8)tinfo->priority;
+	ev.priority = (s8)tinfo->priority;
```
ziemleszcz, line 178: Nitpick: no cast in similar assignment in `trace_events.h`.

proc/lock.h
```diff
@@ -68,7 +68,7 @@ int proc_lockInit(lock_t *lock, const struct lockAttr *attr, const char *name);
 int proc_lockConsistent(lock_t *lock);
 
 
-int proc_lockPrioCeiling(lock_t *lock, int prioceiling);
+int proc_lockPrioCeiling(lock_t *lock, int prioceiling, int *prev);
```
ziemleszcz: `prev` -> `res`

proc/mutex.c
```diff
@@ -144,7 +144,7 @@ int proc_mutexConsistent(int h)
 }
 
 
-int proc_mutexPrioCeiling(int h, int prioceiling)
+int proc_mutexPrioCeiling(int h, int prioceiling, int *prev)
```
ziemleszcz: `prev` -> `res`

proc/mutex.h
```diff
@@ -39,7 +39,7 @@ int proc_mutexTry(int h);
 int proc_mutexConsistent(int h);
 
 
-int proc_mutexPrioCeiling(int h, int prioceiling);
+int proc_mutexPrioCeiling(int h, int prioceiling, int *prev);
```
ziemleszcz: `prev` -> `res`

proc/threads.h
```diff
@@ -22,13 +22,26 @@
...
 #define LIB_ASSERT_THREADS(condition, fmt, ...)
 #endif
 
-#define MAX_PRIO 7U /* Maximum priority value, of the lowest criticality (prio=0 is of the HIGHEST) */
+#ifndef NPRIOS
+#define NPRIOS 64U
+#endif
+
+_Static_assert(NPRIOS % 2U == 0U, "NPRIOS should be even");
+_Static_assert(NPRIOS >= 16U, "NPRIOS should be greater than 16");
```
ziemleszcz: Add `NPRIOS <= 256` assert.

proc/threads.h
```diff
@@ -22,13 +22,26 @@
...
+
+typedef s8 priority_t;
+
+#define PRIO_RANGE (NPRIOS / 2U)
```
ziemleszcz: Misleading name, maybe PRIO_OFFSET?

syscalls.c
```diff
@@ -762,12 +768,17 @@ int syscalls_mutexConsistent(u8 *ustack)
 int syscalls_mutexPrioCeiling(u8 *ustack)
 {
 	handle_t h;
-	int prioceiling;
+	int prioceiling, *prev;
```
ziemleszcz: `prev` - > `res`

proc/threads.c
```diff
@@ -70,9 +77,7 @@ static struct {
 	time_t prev;
 } threads_common;
 
-
-_Static_assert(MAX_PRIO <= (u8)-1, "MAX_PRIO must fit into priority type");
-
+_Static_assert(NPRIOS <= sizeof(threads_common.readyBitmask) * 8U, "NPRIOS must fit into ready bitmask type");
```
ziemleszcz, line 80: Also limited to <=256 by `priority_t` range. May be useful once `readyBitmask` is changed into table.

proc/threads.c
```diff
@@ -201,6 +206,47 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
 }
 
 
+static void _readyAdd(thread_t *t)
+{
+	int sidx = (int)t->priority + (int)PRIO_RANGE;
```
ziemleszcz: Assert `sidx >= 0`?

proc/threads.c
```diff
@@ -201,6 +206,47 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
...
+	size_t idx = (size_t)sidx;
```
ziemleszcz: `size_t` -> `unsigned int` to match size?

proc/threads.c
```diff
@@ -201,6 +206,47 @@ static void _threads_updateWakeup(time_t now, thread_t *minimum)
...
+	LIB_ASSERT_THREADS(LIST_BELONGS(&threads_common.ready[idx], t) != 0, "thread is not on the ready list");
+	LIST_REMOVE(&threads_common.ready[idx], t);
+	if (threads_common.ready[idx] == NULL) {
+		threads_common.readyBitmask &= ~((u64)1U << idx);
+	}
+}
+
+
+static size_t _readyMinPrioIdx(void)
```
ziemleszcz: I would use `unsigned int` to avoid casting.

proc/threads.c
```diff
@@ -798,9 +850,13 @@ int proc_threadPriority(thread_t *t, int signedPriority)
 		(void)hal_spinlockClear(&threads_common.spinlock, &sc);
 	}
 
+	if (res != NULL) {
+		*res = priorityBase;
+	}
+
 	trace_eventThreadPriority(proc_getTid(t), t->priority);
```
ziemleszcz: Assign `priority = t->priority` under spinlock (next to `priorityBase`) and use it here?

proc/threads.c
```diff
@@ -2013,25 +2069,29 @@ int proc_lockConsistent(lock_t *lock)
 }
 
 
-/* prioceiling == -1 retrieves current priority ceiling for the lock without obtaining it */
-int proc_lockPrioCeiling(lock_t *lock, int prioceiling)
+/* prioceiling == PH_GET_PRIO retrieves current priority ceiling for the lock without obtaining it */
+int proc_lockPrioCeiling(lock_t *lock, int prioceiling, int *prev)
```
ziemleszcz: `prev` -> `res`

proc/threads.c
```diff
@@ -2040,15 +2100,20 @@ int proc_lockPrioCeiling(lock_t *lock, int prioceiling)
...
 
-		err = (int)__atomic_exchange_n(&lock->attr.prioceiling, (unsigned char)prioceiling, __ATOMIC_RELAXED);
+		/* parasoft-suppress-next-line MISRAC2012-RULE_10_3-b "__atomic_exchange_1 uses unsigned char but __atomic_exchange_n requires first two arg types to match." */
+		prevCeiling = (int)__atomic_exchange_n(&lock->attr.prioceiling, (signed char)prioceiling, __ATOMIC_RELAXED);
 
 		(void)proc_lockClear(lock);
 	}
 	else {
-		err = (int)__atomic_load_n(&lock->attr.prioceiling, __ATOMIC_RELAXED);
+		prevCeiling = (int)__atomic_load_n(&lock->attr.prioceiling, __ATOMIC_RELAXED);
```
ziemleszcz: It's not a previous value.

ziemleszcz (approved):

PR #836, opened 2026-08-25
posix/unix: make connect() more POSIX compliant

### LINKED ISSUE #1669: connect() with AF_UNSPEC does not reset the peer address of a connectionless socket
## Description
For a connectionless-mode (`SOCK_DGRAM`) `AF_UNIX` socket that already has a peer address set, calling `connect()` with an
address whose `sa_family` member is `AF_UNSPEC` does not reset the peer address. Instead, `connect()` returns `-1` with
`errno` set to `EISCONN`, and the previously set peer address remains in effect.

NOTE: AF_INET socket's behaviour slightly deviates from what POSIX mandates.

## Expected behavior
When the `sa_family` member of the address passed to `connect()` is `AF_UNSPEC`, the socket's peer address shall be reset.
After the reset, a `send()` with no destination address shall fail because the socket no longer has a peer.

## Actual behavior
`connect()` with `AF_UNSPEC` returns `-1` with `errno` set to `EISCONN` (106) and leaves the peer address unchanged; a
subsequent `send()` still succeeds, proving the peer address was not reset.

Observed output:

```
connect(AF_UNSPEC): ret=-1 errno=106
send() after reset:  ret=1 errno=0
```

## POSIX requirement (if applicable)
POSIX.1-2017 (IEEE Std 1003.1-2017), connect(3p), section DESCRIPTION:

  "If the sa_family member of address is AF_UNSPEC, the socket's peer address shall be reset."

## Minimal reproduction code

```c
#include <stdio.h>
#include <string.h>
#include <errno.h>
#include <unistd.h>
#include <sys/socket.h>
#include <sys/un.h>

int main(void)
{
    int srv, cli, ret;
    struct sockaddr_un addr;
    struct sockaddr unspec;

    unlink("/tmp/repro_dgram");
    srv = socket(AF_UNIX, SOCK_DGRAM, 0);
    memset(&addr, 0, sizeof(addr));
    addr.sun_family = AF_UNIX;
    strcpy(addr.sun_path, "/tmp/repro_dgram");
    bind(srv, (struct sockaddr *)&addr, sizeof(addr));

    cli = socket(AF_UNIX, SOCK_DGRAM, 0);
    connect(cli, (struct sockaddr *)&addr, sizeof(addr));

    memset(&unspec, 0, sizeof(unspec));
    unspec.sa_family = AF_UNSPEC;
    errno = 0;
    ret = connect(cli, &unspec, sizeof(unspec));
    printf("connect(AF_UNSPEC): ret=%d errno=%d (%s)\n", ret, errno, strerror(errno));

    errno = 0;
    ret = (int)send(cli, "x", 1, 0);
    printf("send() after reset: ret=%d errno=%d (%s)\n", ret, errno, strerror(errno));

    close(srv);
    close(cli);
    unlink("/tmp/repro_dgram");
    return 0;
}
```

## Host behavior

```
connect(AF_UNSPEC): ret=0 errno=0 (Success)
send() after reset: ret=-1 errno=107 (Transport endpoint is not connected)
```

## Observed behavior on `AF_INET` socket on ia32-generic-qemu
```
connect(AF_UNSPEC): ret=0 errno=0 (EOK)
send() after reset: ret=-1 errno=113 (EHOSTUNREACH)
```
The return code for "send() after reset" is incorrect. It should be 89 (Destination address required).

## Revision
Discovered on 62b46f93f586ebbadf352b1d35eb3ae580543ac0

## Target
Reproduced on ia32-generic-qemu, not tested on other targets yet.

## Related tests
- `phoenix-rtos-tests/libc/socket/connect.c`

These tests are currently skipped on the affected target. After applying a fix, re-enable them and re-run the suite to
verify the fix.

julianuziemblo:
Fixes: phoenix-rtos/phoenix-rtos-project#1669
YT: RTOS-1437

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/518

## Checklist:

## Special treatment


posix/usocket.c
```diff
@@ -586,6 +585,20 @@ int usocket_connect(usocket_t *s, const struct sockaddr *address, socklen_t addr
...
+			if (s->tx != NULL) {
+				oldTx = s->tx;
+				s->tx = NULL;
+			}
+			s->state = (u8)usocketUnconnected;
+		}
+		(void)proc_lockClear(&s->lock);
+		uchannel_put(oldTx);
+		return EOK;
+	}
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist where does POSIX2018 say that about the error codes for  using AF_UNSPEC on connection-oriented sockets?

We already return error values in multiple points in the function as that's far more readable than doing some tricks to return in one place for no reason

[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist idk where does POSIX require "supporting AF_UNSPEC in connect, but only up to the point of checking the socket state, and then erroring..." - this seems convoluted. from what I've checked, connect to AF_UNSPEC with a AF_UNIX SOCK_STREAM socket on Linux just errors with EINVAL (EAFNOSUPPORT would be more compliant so that's what we do). I'll leave this as-is.

We don't use Rule 15.5. as it's advisory only and creates an environment where the code would be unreadable. I'm leaving this as-is also.

posix/usocket.c
```diff
@@ -572,20 +572,33 @@ static void usocket_connectRollback(usocket_t *s)
...
 	/* TODO: validate `address_len` */
-	if ((address == NULL) || (address_len == 0U)) {
+	if ((address == NULL) || (address_len < sizeof(address->sa_family))) {
 		return -EINVAL;
 	}
 
+	if (address->sa_family == (sa_family_t)AF_UNSPEC && s->type == SOCK_DGRAM) {
+		(void)proc_lockSet(&s->lock);
+		if (s->state == (u8)usocketConnected) {
+			if (s->tx != NULL) {
```
ziemleszcz: You can do this unconditionally since `uchannel_put(NULL)` is OK.

posix/usocket.c
```diff
@@ -572,20 +572,33 @@ static void usocket_connectRollback(usocket_t *s)
...
 	size_t rcvbuf;
 	int err, nonblock;
 
 	/* TODO: validate `address_len` */
-	if ((address == NULL) || (address_len == 0U)) {
+	if ((address == NULL) || (address_len < sizeof(address->sa_family))) {
 		return -EINVAL;
 	}
 
+	if (address->sa_family == (sa_family_t)AF_UNSPEC && s->type == SOCK_DGRAM) {
```
ziemleszcz: Nitpick: `((address->sa_family == (sa_family_t)AF_UNSPEC) && (s->type == SOCK_DGRAM))`

posix/usocket.c
```diff
@@ -572,20 +572,33 @@ static void usocket_connectRollback(usocket_t *s)
...
+		(void)proc_lockSet(&s->lock);
+		if (s->state == (u8)usocketConnected) {
```
ziemleszcz: `usocketConnecting` state is not handled correctly. If another thread is doing `connect()` at the same time this call will return EOK but the socket will be connected. Solving this race may be not simple, at least add FIXME/TODO comment or maybe return some error in this case.

posix/usocket.c
ziemleszcz: Handle concurrent AF_UNSPEC connect():
```
if (s->state == (u8)usocketConnecting) {
    oldTx = s->tx;
    s->tx = ch;
    s->state = (u8)usocketConnected;
}
```

ziemleszcz (approved):

PR #840, opened 2026-09-04
New UNIX sockets implementation

ziemleszcz:
## Description

Rebuilds UNIX sockets on endpoint/channel model.
A socket is an endpoint (usocket_t) and, once connected, two unidirectional
data channels (uchannel_t). One channel is used for reads and the other
for writes. socketpair() and accept() create the channels crosswise. All
state shared between endpoints lives in the channels, so an endpoint never
dereferences its peer after connect() or accept().

## Types of changes

## How Has This Been Tested?

- [x] New test added: (https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/515).

## Checklist:

## Special treatment

- [ ] This PR needs additional PRs to work ().


posix/uchannel.c
```diff
@@ -0,0 +1,440 @@
...
+		}
+		else {
+			/* no complete frame */
+		}
+
+		if (ret > 0) {
+			if ((flags & UCHANNEL_OP_PEEK) == 0U) {
+				if (packs != NULL) {
+					uchannel_takePacks(ch, packs);
+				}
```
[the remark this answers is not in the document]
ziemleszcz: Implementation limitation to be addressed in another PR.

posix/uchannel.c
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+	(void)proc_threadBroadcast(&ch->txwait);
+
+	(void)proc_lockClear(&ch->lock);
+
+	vm_kfree(old.data);
+
+	return 0;
+}
```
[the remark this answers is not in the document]
ziemleszcz: Pre-exisiting issue. Data dropping is marked with FIXME comment.

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+int usocket_bind(usocket_t *s, const struct sockaddr *address, socklen_t address_len)
+{
+	char *path, *name;
+	const char *dir;
+	oid_t odir, dev, node;
+	int err, id;
+
+	if ((address == NULL) || (address->sa_family != (sa_family_t)AF_UNIX)) {
+		return -EINVAL;
+	}
```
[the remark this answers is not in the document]
ziemleszcz: Pre-exisiting issue.

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+{
+	usocket_t *ls;
+	uchannel_t *ch, *rx, *old_tx;
+	oid_t oid;
+	size_t rcvbuf;
+	int err, nonblock;
+
+	if ((address == NULL) || (address->sa_family != (sa_family_t)AF_UNIX)) {
+		return -EINVAL;
+	}
```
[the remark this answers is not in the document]
ziemleszcz: Pre-exisiting issue.

posix/uchannel.c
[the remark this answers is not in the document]
ziemleszcz: Pre-exisiting issue. Data dropping is marked with FIXME comment.

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+
+	uchannel_put(ch);
+
+	return err;
+}
+
+
+int usocket_accept4(usocket_t *ls, struct sockaddr *address, socklen_t *address_len, unsigned int flags, usocket_t **s)
+{
+	usocket_t *c, *ns;
```
julianuziemblo, line 762: nitpick: name it `cs` for "client socket". Maybe also name the input socket in usocket_connect as `cs` to stay consistent.

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+	uchannel_t *c2s, *tx, *old_tx;
```
julianuziemblo, line 763: nitpick: `oldTx`

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+	wantsControl = ((control != NULL) && (controllen != NULL) && (*controllen > 0U)) ? 1 : 0;
+
+	(void)proc_lockSet(&s->lock);
+	op = usocket_opFlags(s, flags);
+	rx = uchannel_ref(s->rx);
+	shutRd = ((s->flags & USOCKET_SHUT_RD) != 0U) ? 1 : 0;
+	(void)proc_lockClear(&s->lock);
+
+	if ((src_addr != NULL) && (src_len != NULL)) {
+		/* the peer of a UNIX socket has no address to report here */
```
julianuziemblo, line 1003: what does this mean? is this a TODO?

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+	oid_t oid;
+	ssize_t ret;
+	unsigned int op;
+	int dgram, err;
+
+	dgram = (s->type == SOCK_DGRAM) ? 1 : 0;
+
+	if ((dgram != 0) && (dest_addr != NULL) && (dest_len != 0U)) {
+		if (dest_addr->sa_family != (sa_family_t)AF_UNIX) {
+			return -EINVAL;
```
julianuziemblo, line 1067: `EAFNOSUPPORT`

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+		}
+
+		if (proc_lookup(dest_addr->sa_data, NULL, &oid) < 0) {
+			return -ECONNREFUSED;
+		}
+
+		if (oid.port != USOCKET_PORT) {
+			return -ECONNREFUSED;
```
julianuziemblo, line 1075: `ENOENT`

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+
+	dgram = (s->type == SOCK_DGRAM) ? 1 : 0;
+
+	if ((dgram != 0) && (dest_addr != NULL) && (dest_len != 0U)) {
+		if (dest_addr->sa_family != (sa_family_t)AF_UNIX) {
+			return -EINVAL;
+		}
+
+		if (proc_lookup(dest_addr->sa_data, NULL, &oid) < 0) {
+			return -ECONNREFUSED;
```
julianuziemblo, line 1071: `ENOENT`

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+	int dgram, err;
+
+	dgram = (s->type == SOCK_DGRAM) ? 1 : 0;
+
+	if ((dgram != 0) && (dest_addr != NULL) && (dest_len != 0U)) {
+		if (dest_addr->sa_family != (sa_family_t)AF_UNIX) {
+			return -EINVAL;
+		}
+
+		if (proc_lookup(dest_addr->sa_data, NULL, &oid) < 0) {
```
julianuziemblo, line 1070: TODO validate `dest_len`

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+	}
+
+	(void)proc_lockSet(&s->lock);
+	op = usocket_opFlags(s, flags);
+	if ((s->flags & USOCKET_SHUT_WR) != 0U) {
+		err = -EPIPE;
+	}
+	else {
+		tx = uchannel_ref(s->tx);
+		err = (tx == NULL) ? -ENOTCONN : EOK;
```
julianuziemblo, line 1126: there is also `EDESTADDRREQ` for "unconnected" SOCK_DGRAM sockets using send. Maybe it'd worth using it here.

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+			return -ECONNREFUSED;
+		}
+
+		if (oid.port != USOCKET_PORT) {
+			return -ECONNREFUSED;
+		}
+
+		d = usocket_get(oid.id);
+		if (d == NULL) {
+			return -ECONNREFUSED;
```
julianuziemblo, line 1080: `EHOSTUNREACH`

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+
+		if (err != EOK) {
+			uchannel_put(tx);
+			return (ssize_t)err;
+		}
+
+		if (tx == NULL) {
+			return -ECONNREFUSED;
+		}
+
```
julianuziemblo, line 1107: why not do this in L1093 (before locking s->lock)?

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+		err = ((s->flags & USOCKET_SHUT_WR) != 0U) ? -EPIPE : EOK;
+		(void)proc_lockClear(&s->lock);
+
+		if (err != EOK) {
+			uchannel_put(tx);
+			return (ssize_t)err;
+		}
+
+		if (tx == NULL) {
+			return -ECONNREFUSED;
```
julianuziemblo, line 1105: `EHOSTUNREACH`

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+		}
+
+		ret = uchannel_write(tx, buf, len, op, fdpack);
+
+		uchannel_put(tx);
+
+		return (ret == -EPIPE) ? -ECONNREFUSED : ret;
```
julianuziemblo, line 1112: `EHOSTUNREACH`

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+
+	nonblock = ((s->flags & USOCKET_NONBLOCK) != 0U) ? 1 : 0;
+
+	while (s->state == (u8)usocketConnecting) {
+		if (nonblock != 0) {
+			(void)proc_lockClear(&s->lock);
+			return -EINPROGRESS;
+		}
+
+		err = proc_lockWait(&s->connq, &s->lock, 0);
```
julianuziemblo, line 720: removed timeout (which granted, was not used anyway, but) - maybe we should add a FIXME or at least explain that "POSIX requires implementation-defined timeout, so we make it infinity, hehe"

posix/posix.c
```diff
@@ -840,12 +847,14 @@ ssize_t posix_write(int fildes, void *buf, size_t nbyte, off_t offset)
...
-		 * not return EPIPE for no longer connected SOCK_DGRAM sockets, because SIGPIPE is not
-		 * required for them.
+		/*
+		 * NOTE: a UNIX socket reports EPIPE in two cases, and both raise SIGPIPE here:
+		 * the socket has been shut down for writing, whatever its type, or its peer
+		 * will never read again. POSIX requires SIGPIPE only for the second case and
+		 * only for SOCK_STREAM and SOCK_SEQPACKET, and permits it for the first one.
+		 * A SOCK_DGRAM socket therefore does get SIGPIPE once shut down for writing,
+		 * but never for a peer that is gone: usocket_sendto() reports that as
+		 * ECONNREFUSED, which POSIX asks for and which EPIPE would turn into a signal.
```
julianuziemblo, line 857: can't find `ECONNREFUSED` in POSIX send/sendto/sendmsg

lib/cbuffer.c
```diff
@@ -105,3 +105,34 @@ size_t _cbuffer_peek(const cbuffer_t *buf, void *data, size_t sz)
...
+	if (sz == 0U || buf->sz == 0U) {
+		return 0U;
+	}
+
+	avail = _cbuffer_avail(buf);
+
+	if (offs >= avail) {
+		return 0U;
+	}
+
```
julianuziemblo, line 126: nitpick: unnecessary double newline

posix/uchannel.c
```diff
@@ -0,0 +1,447 @@
...
+		}
+	}
+	else {
+		/*
+		 * FIXME: the buffered data does not fit the new ring and is dropped
+		 * whole, so that a frame is never left half present. The sender's
+		 * send() has already reported those bytes as accepted and the reader
+		 * has no way to tell they went missing - a byte stream loses a piece
+		 * out of its middle, a record socket loses entire records.
+		 */
```
julianuziemblo, line 425: nitpick: "record socket" -> "framed socket" (so that we use the same terminology)

posix/uchannel.c
```diff
@@ -0,0 +1,447 @@
...
+
+	if (avail <= size) {
+		/* copy the buffered bytes over in order */
+		if (avail > 0U) {
+			first = min(avail, old.sz - old.r);
+			(void)_cbuffer_write(&ch->buffer, (const char *)old.data + old.r, first);
+			if (avail > first) {
+				(void)_cbuffer_write(&ch->buffer, old.data, avail - first);
+			}
+		}
```
julianuziemblo, line 416: maybe make this a generic operation on the cbuffer (`_cbuffer_resize`?)

julianuziemblo (commented):
Initial review - I'll try to make myself even more familiar with the new code and look for more edge cases before accepting. For now, looks like this solution makes a lot of sense and is well thought out.

posix/usocket.c
ziemleszcz: Keeping ECONNREFUSED here (the same error code as in Linux).

posix/usocket.c
ziemleszcz: Keeping ECONNREFUSED here (the same error code as in Linux).

posix/posix.c
```diff
@@ -1814,11 +1826,12 @@ int posix_socket(int domain, int type, int protocol)
...
 	switch (domain) {
 		case AF_UNIX:
-			err = unix_socket(domain, (unsigned int)type, protocol);
-			if (err >= 0) {
+			err = usocket_socket(domain, (unsigned int)type, protocol, &sock);
+			if (err == 0) {
 				p->fds[fd].file->type = ftUnixSocket;
-				p->fds[fd].file->oid.port = US_PORT;
-				p->fds[fd].file->oid.id = (unsigned int)err;
+				p->fds[fd].file->sock = sock;
```
adamgreloch, line 1832: Isn't this racy with `posix_close`?
1. `p->fds[fd].file->type = ftUnixSocket`
2. preemption before `p->fds[fd].file->sock = sock`;
3. another thread does `posix_close`, sees `ftUnixSocket` and does `usocket_close` on `f->sock == p->fds[fd].file->sock == NULL`

posix/uchannel.c
```diff
@@ -0,0 +1,447 @@
...
+{
+	unsigned int events = 0;
+	size_t free;
+
+	(void)proc_lockSet(&ch->lock);
+
+	if ((ch->flags & (UCHANNEL_SHUT_RD | UCHANNEL_SHUT_WR)) == 0U) {
+		free = _cbuffer_free(&ch->buffer);
+		if ((ch->framed == 0U) ? (free > 0U) : (free > sizeof(size_t))) {
+			events |= UCHANNEL_EV_OUT;
```
adamgreloch: why no `events |= UCHANNEL_EV_OUT;` on shutdown? Linux sets it: https://elixir.bootlin.com/linux/v7.2.2/source/net/unix/af_unix.c#L3390

posix/posix.c
ziemleszcz: `return (ret == -EPIPE) ? -ECONNREFUSED : ret;`

posix/uchannel.c
ziemleszcz: This isn't a cbuffer resize. We're copying data from one cbuffer to another. I've added a TODO for now.

posix/posix.c
julianuziemblo: Sorry this is not what I meant: I can't find the error anywhere in POSIX specification for send/sendto/sendmsg's return value

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+
+	(void)proc_lockSet(&s->lock);
+
+	s->state = (u8)usocketClosed;
+	rx = s->rx;
+	tx = s->tx;
+	s->rx = NULL;
+	s->tx = NULL;
+	/* as in usocket_shutdown(), a shared datagram channel is not ours to mark */
+	shutTx = (s->type != SOCK_DGRAM) ? 1 : 0;
```
adamgreloch, line 1375: It makes sense for named datagram sockets. What about unnamed though?
`close(sv[0])` will result in close flags for `ch[1]`/`sv[1]` to never be set

posix/usocket.c
ziemleszcz: To return -EPIPE instead of -ECONNREFUSED.

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+	 */
+	rx = uchannel_alloc(rcvbuf, usocket_isFramed(s));
+	if (rx == NULL) {
+		usocket_put(ls);
+		usocket_connectRollback(s);
+		return -ENOMEM;
+	}
+
+	(void)proc_lockSet(&s->lock);
+	uchannel_put(s->rx);
```
adamgreloch: if s->rx is the last reference, then:

uchannel_put -> fdpass_discard -> posix_fileDeref ->  usocket_close -> proc_lockSet

so more than one lock at a time. Just do 

```
(void)proc_lockSet(&s->lock);
oldRx = s->rx;
s->rx = rx;
(void)proc_lockClear(&s->lock);

uchannel_put(oldRx);
```
instead?

posix/uchannel.c
```diff
@@ -0,0 +1,447 @@
...
+	for (;;) {
+		if (len == 0U) {
+			/*
+			 * Reading nothing must not consume a frame. It still reports
+			 * readability instead of waiting for data it would not take.
+			 */
+			ret = 0;
+			if (_cbuffer_avail(&ch->buffer) > 0U) {
+				break;
+			}
```
adamgreloch: this doesn't break out of the loop, but should per POSIX:

> Before any action described below is taken, and if nbyte is zero, the read() function may detect and return errors as described below. In the absence of errors, or if error detection is not performed, the read() function **shall return zero and have no other results.**

https://pubs.opengroup.org/onlinepubs/9799919799/functions/read.html

there is a guard for that in `posix_read`, but none for recvfrom, recvmsg. Better to do it here and drop the guard in `posix_read`.

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+	(void)proc_lockClear(&s->lock);
+
+	if (state == usocketListening) {
+		if (pending != 0) {
+			revents |= (unsigned int)events & (POLLIN | POLLRDNORM | POLLRDBAND);
+		}
+	}
+	else if (state == usocketAborted) {
+		/* a failed connect() - POLLERR and POLLHUP are reported unconditionally */
+		revents |= POLLERR | POLLHUP;
```
adamgreloch, line 1463: should we report POLLERR forever? On Linux, consuming `SO_ERROR` clears `POLLERR`. Maybe report POLLERR only when `s->err != 0`, since it gets consumed?

posix/usocket.c
```diff
@@ -0,0 +1,1476 @@
...
+
+	(void)proc_lockClear(&s->lock);
+
+	uchannel_put(ch);
+
+	return err;
+}
+
+
+int usocket_accept4(usocket_t *ls, struct sockaddr *address, socklen_t *address_len, unsigned int flags, usocket_t **s)
```
adamgreloch, line 775: validate `flags`?

posix/uchannel.c
```diff
@@ -0,0 +1,447 @@
...
+			ret = (ssize_t)_cbuffer_write(&ch->buffer, buf, len);
+		}
+		else if (len > (ch->buffer.sz - sizeof(len))) {
+			ret = -EMSGSIZE;
+			break;
+		}
+		else if (_cbuffer_free(&ch->buffer) >= (len + sizeof(len))) {
+			(void)_cbuffer_write(&ch->buffer, &len, sizeof(len));
+			(void)_cbuffer_write(&ch->buffer, buf, len);
+			ret = (ssize_t)len;
```
adamgreloch: preexisting, but why do we break out early after any successful write to the cbuf, instead of looping until `len` is fully written? Linux seems to loop.

posix/posix.c
ziemleszcz: This is pre-existing Linux specific behavior.

posix/usocket.c
ziemleszcz: Linux specific behavior, out tests assume ENOTCONN.

posix/posix.c
ziemleszcz: Possibly. It seems to me that the file should be created and populated atomically, but it isn't. Anyway, this is a pre-existing issue and is out of scope for this PR. Feel free to create a new issue for it.

posix/uchannel.c
ziemleszcz: Socket is not writeable - `uchannel_write()` returns -EPIPE when UCHANNEL_SHUT_RD or UCHANNEL_SHUT_WR are set.

posix/usocket.c
ziemleszcz: Please elaborate. I don't understand what you're getting at.

posix/uchannel.c
adamgreloch: Ok, but poll won't return POLLOUT, so the writer won't have a chance to get the -EPIPE from write

posix/usocket.c
adamgreloch: This doesn't shut the tx even if the pair is connected through a private channel. If ch[0] closes, the ch[1] will be essentially dead forever, as it isn't in the shutdown state and also has no way to receive new data (since ch[0] is closed). Seems weird.

But I see that Linux behaves the same, so this is likely expected.

posix/usocket.c
ziemleszcz: According to POSIX and connect() man page: `If connect() fails, the state of the socket is unspecified. Conforming applications should close the file descriptor and create a new socket before attempting to reconnect.`. Clearing POLLERR flag could suggest that socket is usable.

posix/usocket.c
ziemleszcz: Added a TODO for now. Note that this PR is not intended to fix all UNIX socket issues. It reimplements UNIX sockets to fix data races and object ownership problems.

posix/posix.c
adamgreloch: ok, RTOS-1447

posix/usocket.c
ziemleszcz: Good catch.

posix/uchannel.c
ziemleszcz: Good catch. I have no idea how this went unnoticed for so long.

posix/usocket.c
ziemleszcz: Rewrote the comment as a TODO.

posix/usocket.c
ziemleszcz: Keeping ECONNREFUSED here (the same error code as in Linux).

posix/usocket.c
ziemleszcz: Keeping ECONNREFUSED here (the same error code as in Linux).

posix/usocket.c
ziemleszcz: Added a comment.

posix/uchannel.c
ziemleszcz: Changed `poll()` to match Linux behavior for UNIX sockets.

posix/fdpass.c
```diff
@@ -193,17 +193,9 @@ int fdpass_unpack(fdpack_t **packs, void *control, socklen_t *controllen)
 
 int fdpass_discard(fdpack_t **packs)
```
julianuziemblo: nit: this can now just be `void`

posix/uchannel.c
julianuziemblo, line 1: nitpick: commit message: `uchan_t` $\rightarrow$ `uchannel_t`, `UCHAN_SHUT_{RD|WR}` $\rightarrow$ `UCHANNEL_SHUT_{RD|WR}`

posix/uchannel.c
```diff
@@ -0,0 +1,462 @@
...
+			/* not enough room for the whole frame */
+		}
+
+		if ((flags & UCHANNEL_OP_NONBLOCK) != 0U) {
+			ret = (done > 0U) ? (ssize_t)done : -EWOULDBLOCK;
+			break;
+		}
+
+		err = proc_lockWait(&ch->txwait, &ch->lock, 0);
+		if (err == -EINTR) {
```
julianuziemblo, line 165: hm, now unix socket reads and writes (and from what I've seen also connect and accept) are interruptible - it is more POSIX-compliant to be interruptible, but I don't think we want this in Phoenix when we say we have uninterruptible IO.

posix/usocket.c
```diff
@@ -0,0 +1,1520 @@
...
+		return;
+	}
+
+	if (lib_atomicDecrement(&s->refs) != 0) {
+		return;
+	}
+
+	/*
+	 * The last reference is gone so the socket is in no tree, no descriptor and
+	 * no pending list, so nothing can reach it and no lock is needed.
```
julianuziemblo, line 181: I thought `uchannel_put` should not be done under a lock anyway, so why the comment?

posix/uchannel.c
```diff
@@ -0,0 +1,462 @@
...
+				done += chunk;
+				(void)proc_threadBroadcast(&ch->rxwait);
+			}
+
+			if (done == len) {
+				ret = (ssize_t)done;
+				break;
+			}
+		}
+		else if (len > (ch->buffer.sz - sizeof(len))) {
```
julianuziemblo, line 141: nitpick: `len + sizeof(len) > ch->buffer.sz` to be consistent with similar condition in L145

posix/uchannel.c
```diff
@@ -0,0 +1,462 @@
...
+		}
+	}
+
+	(void)proc_lockClear(&ch->lock);
+
+	return ret;
+}
+
+
+static void uchannel_takePacks(uchannel_t *ch, fdpack_t **packs)
```
julianuziemblo: nitpick: `_uchannel_takePacks` as this has to be done under `ch->lock`

posix/uchannel.c
```diff
@@ -0,0 +1,462 @@
...
+		else if (ch->framed == 0U) {
+			if ((flags & UCHANNEL_OP_PEEK) != 0U) {
+				ret = (ssize_t)_cbuffer_peek(&ch->buffer, buf, len);
+			}
+			else {
+				ret = (ssize_t)_cbuffer_read(&ch->buffer, buf, len);
+			}
+		}
+		else if (_cbuffer_avail(&ch->buffer) > sizeof(rlen)) {
+			(void)_cbuffer_peekAt(&ch->buffer, 0U, &rlen, sizeof(rlen));
```
julianuziemblo: peek at offset 0 $\rightarrow$ just `_cbuffer_peek`

posix/usocket.c
```diff
@@ -0,0 +1,1520 @@
...
+			return (ssize_t)err;
+		}
+
+		/* a socket shut down for reading is EOS, not unconnected */
+		return (shutRd != 0) ? 0 : -ENOTCONN;
+	}
+
+	ret = uchannel_read(rx, buf, len, op, (wantsControl != 0) ? &packs : NULL);
+
+	if (packs != NULL) {
```
julianuziemblo, line 1048: shouldn't be `wantsControl != 0`?

posix/usocket.c
```diff
@@ -0,0 +1,1520 @@
...
+		/*
+		 * No lock is held here - fdpass_unpack() reaches into the file
+		 * descriptor table of this process.
+		 */
+		(void)fdpass_unpack(&packs, control, controllen);
+		if (packs != NULL) {
+			uchannel_returnPacks(rx, &packs);
+		}
```
julianuziemblo, line 1056: I'm not sure if it's the expected behavior to return the packs back - shouldn't the extra packs just be discarded? that seems to be the whole reason for `MSG_CTRUNC` flag (to be added separately in my PR)

posix/usocket.c
```diff
@@ -0,0 +1,1520 @@
...
+
+		return (ret == -EPIPE) ? -ECONNREFUSED : ret;
+	}
+
+	if ((dgram == 0) && ((dest_addr != NULL) || (dest_len != 0U))) {
+		return -EISCONN;
+	}
+
+	(void)proc_lockSet(&s->lock);
+	op = usocket_opFlags(s, flags);
```
julianuziemblo: consider aliasing the `op` type as e.g. `uchannel_flags_t`, or renaming the parameter of `uchannel_{read|write}` to e.g. `opFlags`, `uchannelFlags`, `op`, etc., as IMO it could be kinda confusing that we're not just passing `flags` here.

posix/usocket.c
```diff
@@ -0,0 +1,1520 @@
...
+		lib_idtreeRemove(&usocket_common.tree, &s->linkage);
+	}
+
+	(void)proc_lockClear(&usocket_common.lock);
+
+	return named;
+}
+
+
+static unsigned int usocket_opFlags(const usocket_t *s, unsigned int flags)
```
julianuziemblo: this has to be used under a lock, so `_usocket_opFlags`

posix/usocket.c
```diff
@@ -0,0 +1,1520 @@
...
+	}
+
+	if (usocket_nameFree(s) != 0) {
+		usocket_put(s);
+	}
+
+	usocket_put(s);
+
+	return EOK;
+}
```
julianuziemblo, line 1358: consider making it `void` (as you already change its name) or returning meaningful error on L1350

posix/uchannel.c
```diff
@@ -0,0 +1,462 @@
...
+	}
+
+	(void)proc_lockSet(&ch->lock);
+
+	for (;;) {
+		if (len == 0U) {
+			/*
+			 * Reading nothing must not consume a frame. It still reports
+			 * readability instead of waiting for data it would not take.
+			 */
```
julianuziemblo, line 206: when `len == 0; _cbuffer_avail(&ch->buffer) == 0`, then `ret` is not `> 0` (L240), we "wait for data we won't take" - the comment seems misleading to me in that situation. "reports readability" is correct, but "instead of waiting for data it would not take" is misleading as we do wait for the data we won't take

posix/uchannel.c
```diff
@@ -0,0 +1,462 @@
...
+
+		if (ret > 0) {
+			if ((flags & UCHANNEL_OP_PEEK) == 0U) {
+				if (packs != NULL) {
+					uchannel_takePacks(ch, packs);
+				}
+				(void)proc_threadBroadcast(&ch->txwait);
+			}
+			break;
+		}
```
julianuziemblo, line 249: From what I checked, at least Linux allows reading only control information (`recvmsg` with `msg_iovlen == 0`). This is particularly important if the sender sends only the FDs with no data.

posix/usocket.c
```diff
@@ -0,0 +1,1520 @@
...
+	}
+
+	if (address->sa_family != (sa_family_t)AF_UNIX) {
+		return -EAFNOSUPPORT;
+	}
+
+	(void)proc_lockSet(&s->lock);
+	if (s->state != (u8)usocketUnconnected) {
+		(void)proc_lockClear(&s->lock);
+		return -EINVAL;
```
julianuziemblo: why can we only bind an unconnected socket? There's no specification for EINVAL meaning "tried to bind connected socket" in POSIX, and it works in Linux.

posix/uchannel.c
adamgreloch: the uninterruptibility of IO is a limitation, not a feature and concerns only the msg-based IO

posix/uchannel.c
```diff
@@ -0,0 +1,462 @@
...
+
+	(void)proc_lockSet(&ch->lock);
+
+	free = _cbuffer_free(&ch->buffer);
+	if ((ch->framed == 0U) ? (free > 0U) : (free > sizeof(size_t))) {
+		events |= UCHANNEL_EV_OUT;
+	}
+
+	if ((ch->flags & (UCHANNEL_SHUT_WR | UCHANNEL_SHUT_RD)) != 0U) {
+		events |= UCHANNEL_EV_SHUT;
```
adamgreloch, line 397: This is better, one corner case left though: if there's no free space and the channel is shut, `UCHANNEL_EV_OUT` still won't be reported. Just set `events |= (UCHANNEL_EV_SHUT | UCHANNEL_EV_OUT)`?

posix/usocket.c
```diff
@@ -0,0 +1,1520 @@
...
+			ev = uchannel_pollRd(rx);
+			if ((ev & UCHANNEL_EV_IN) != 0U) {
+				revents |= (unsigned int)events & (POLLIN | POLLRDNORM | POLLRDBAND);
+			}
+			if ((ev & UCHANNEL_EV_SHUT) != 0U) {
+				readShut = 1;
+			}
+		}
+
+		if (tx != NULL) {
```
adamgreloch, line 1486: shouldn't poll report POLLOUT for unconnected sockets (`tx == NULL`)?

posix/usocket.c
ziemleszcz: `usocket_common.lock`  and other socket lock is not needed.

posix/usocket.c
ziemleszcz: `wantsControl` must be 1 if `packs != NULL`  (see line above).

julianuziemblo: okay yeah it does make more sense this way as we want to do this only when some packs were provided, and they will be provided only when `wantControl != 0`.

posix/usocket.c
ziemleszcz: Keeping existing unpack semantics. Changing it is out of scope for this PR.

julianuziemblo: ok, I see how it's the same as before. Then this is fine, I'll implement the new behavior in my PR then.

posix/uchannel.c
ziemleszcz: Out of scope for this PR.

julianuziemblo: my bad, didn't see that the current code also checks `err > 0`

posix/usocket.c
ziemleszcz: Linux does, but I would prefer not to make further changes to `poll()` in this PR without fully understanding the Linux implementation, especially since it deviates from POSIX. Maybe raise a separate issue for this and the other poll() comments, and we can address them in a separate PR.

adamgreloch: Could you add TODO/FIXMEs for these two matters instead? This would document the behavior better

posix/uchannel.c
ziemleszcz: No, this check would not catch int overflow.

posix/usocket.c
ziemleszcz: Updated bind() and listen() to allow this case.

posix/uchannel.h
```diff
@@ -0,0 +1,165 @@
...
+
+/* Allocates a channel with one reference. `size` must be a power of two. */
+uchannel_t *uchannel_alloc(size_t size, int framed);
+
+
+uchannel_t *uchannel_ref(uchannel_t *ch);
+
+
+/* Drops one reference; destroys the channel (and discards any descriptors
+ * still queued on it) when the last one is gone. NULL is allowed. */
```
adamgreloch:
```suggestion
/*
 * Drops one reference; destroys the channel (and discards any descriptors
 * still queued on it) when the last one is gone. NULL is allowed.
 */
```

posix/uchannel.c
julianuziemblo: Bump

posix/usocket.c
adamgreloch: Bump

adamgreloch (dismissed):

julianuziemblo (dismissed):
LGTM

adamgreloch (approved):

PR #841, opened 2026-09-08
!posix: implement process groups and sessions

adamgreloch:
Breaking change: adds sys_getsid syscall and renumbers the table.

Assisted-by: Claude-opus-5

TASK: RTOS-1449

## Types of changes

## How Has This Been Tested?

- [X] New test added: lukkrusz/rebased_posix (to be merged)

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/libphoenix/pull/526
https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/696
https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/276
https://github.com/phoenix-rtos/phoenix-rtos-posixsrv/pull/37


posix/posix.c
```diff
@@ -2603,16 +2702,15 @@ int posix_tkill(pid_t pid, int tid, int sig)
 		return -EINVAL;
 	}
 
-	/* TODO: handle pid = 0 */
-	if (pid == 0) {
-		return -ENOSYS;
+	if (pid == -1) {
+		return posix_killAll(sig);
```
ziemleszcz: I don't think killing all processes, including all drivers and servers, is what we want here. It seems like useless functionality, since the system would be left in an unusable state.

POSIX says: `shall be sent to all processes (excluding an unspecified set of system processes)`. Maybe we should kill only processes from the same session?

posix/posix.c
```diff
@@ -318,10 +331,17 @@ int posix_clone(int ppid)
...
 
 	pp = pinfo_find(ppid);
 	if (pp != NULL) {
 		TRACE("clone: got parent");
+
+		(void)proc_lockSet(&posix_common.lock);
+		ppgid = pp->pgid;
+		psid = pp->sid;
+		(void)proc_lockClear(&posix_common.lock);
+
```
ziemleszcz: Setting clone `ppgid` and `psid` is not atomic. Another thread from parent process can change it meanwhile.
Maybe solved by using process locks in posix_setsid/setpgid etc. but maybe you'll find better solution.

lib/atomic.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+#ifndef _PH_LIB_ATOMIC_H_
+#define _PH_LIB_ATOMIC_H_
+
+#include "include/posix-types.h"
+
+/* parasoft-begin-suppress MISRAC2012-RULE_10_3-b "__atomic intrinsics are type-generic and internally are mostly implemented as taking unsigned types, thereby confusing the rule checker" */
+
+static inline void atomic_store_pid(pid_t *ptr, pid_t val, int memmodel)
```
ziemleszcz: Probably should be integrated with existing atomic functions in `lib/utils.h`.

posix/posix_private.h
```diff
@@ -111,11 +112,16 @@ typedef struct _process_info_t {
...
 	pid_t pgid;
+	pid_t sid;
+	unsigned int exec; /* set to 1 once process successfully called exec/spawn */
+
 	lock_t lock;
-	int maxfd;
 	int fdsz;
 	fildes_t *fds;
+
+	int maxfd;
```
ziemleszcz: Move it back where it was, it adds extra padding on 64-bit targets.

posix/posix.c
```diff
@@ -415,6 +443,10 @@ int posix_exec(void)
 		return -1;
 	}
 
+	(void)proc_lockSet(&posix_common.lock);
+	p->exec = 1;
```
ziemleszcz: Use atomic instead of blocking on  global lock?

posix/posix.c
adamgreloch: How? `p` is created in clone and `p->pgid`, `p->sid` are set before exposing it. `posix_common.lock` synchronizes posix_setsid and posix_setpgid

ziemleszcz: `posix_clone()` sets local `ppgid` and `psid`, releases `posix_common.lock`, `posix_setsid()` changes `sid` and `posix_clone()` sets stale values.

posix/posix.c
```diff
@@ -394,13 +415,47 @@ int posix_clone(int ppid)
...
 		p->fds[2].file->status = O_WRONLY;
 
+		/* The first POSIX process is the leader of its own session and group */
 		p->pgid = p->process;
+		p->sid = p->process;
 	}
 
 	(void)proc_lockSet(&posix_common.lock);
+	if (pp != NULL) {
+		(void)proc_lockSet(&pp->lock);
```
ziemleszcz, line 425: ABBA deadlock with posix_died() - it takes these two locks in reverse order. Try using the same order, maybe you can combine it with `if-else` above.

adamgreloch: Corrected. Documented the ordering as well.

posix/posix.c
```diff
@@ -318,16 +331,31 @@ int posix_clone(int ppid)
...
 		p->maxfd = pp->maxfd;
 		p->fdsz = pp->fdsz;
 		p->parent = ppid;
 	}
 	else {
+		/*
+		 * FIXME: a parent reaped before pinfo_find() above is indistinguishable
+		 * from the root process and from a non-POSIX parent, so the child is
+		 * created parentless instead of failing the clone
+		 */
```
ziemleszcz, line 358: Shouldn't `init` adopt parentless process?
```
if (pp == NULL) {
    pp = pinfo_find(1);
    ppid = 1;
}
```

posix/posix.c
```diff
@@ -392,13 +414,25 @@ int posix_clone(int ppid)
...
 		p->fds[2].file->status = O_WRONLY;
 
+		/* The first POSIX process is the leader of its own session and group */
 		p->pgid = p->process;
+		p->sid = p->process;
 	}
 
 	(void)proc_lockSet(&posix_common.lock);
+	if (pp != NULL) {
+		p->pgid = pp->pgid;
```
ziemleszcz, line 432: Read `pgid` and `sid` atomically?

posix/posix.c
```diff
@@ -2680,15 +2872,32 @@ pid_t posix_getpgid(pid_t pid)
...
+}
+
+
+pid_t posix_getsid(pid_t pid)
+{
+	process_info_t *pinfo;
+	pid_t res;
+
+	if (pid < 0) {
+		return -EINVAL;
```
ziemleszcz, line 2890: ESRCH

posix/posix.c
ziemleszcz, line 2947: ESRCH

posix/posix.c
adamgreloch: I don't think that's needed. Writes of both are protected by posix_common.lock. Atomics for pgid are to allow lock-free read.

posix/posix.c
adamgreloch:
```
The getpgid() function may fail if:

[EINVAL]
The value of the pid argument is invalid.
```

ziemleszcz (approved):

